PR #34, opened 2015-11-19
Update formatCommand error messaging

tampajohn:
The information message provided was explicitly for 'goreturns', should reflect the configured formatCommand.  Also not sure if providing go get information is particularly useful here, since extension already allows for the installation of tools.


lukehoban:
Thanks @tampajohn!

Issue #15, opened 2015-11-18
where is console output (stdout)


jeromelaurens:
Hi here, 
not sure it is an issue, probably more a question.
Where is the stdout console ? where we can see all the log.Print...
The logs are not in the terminal that launched VSCode
I'm on MacOSX.

Thanks
Jerome

mattetti:
Same question here, I'm having issues with the debugger and wish I could see the output, I'm sure it's somewhere obvious that I just can't find :(

jchannon:
View -> Toggle Output?

mattetti:
@jchannon when I do that, it only shows the git output

shurcooL:
After seeing [`console.log`](https://github.com/Microsoft/vscode-go/blob/eca07a3a3355f0038460830a6a876111271c141d/src/goPath.ts#L26), I was wondering the same thing.

I found it in the help menu, under "Toggle Developer Tools", on OS X.

![image](https://cloud.githubusercontent.com/assets/1924134/11257166/e14f38d6-8e04-11e5-8e2d-f195cd478f52.png)

mattetti:
thanks @shurcooL but it looks like the debugger isn't logging anything when launching, I was having issues with the debugger because I didn't properly install dlv on OS X: https://github.com/Microsoft/vscode-go/issues/20

I did wish there was a way to see stdout output somewhere but it might be up to the extension you use.

rudygt:
Same here,

the "Debug Console" is always empty after the execution of a simple hello world, if I put a breakpoint in the printf it stops there but the output is not show anywhere.

jeromelaurens:
Thanks @shurcooL but we are talking here about the go program output when debugging. I tried to open the console from developer tools : the output is not showing.
Same behaviour as @rudygt  : the debug console always empty, but breakpoint works with variables.
I have other strange behaviour when debugging, but before posting them, I'd like to see the output.
About Delve, I have installed the latest version with the certificate, everything seems fine (I'm using it with intellig idea go plugin)

lukehoban:
Agreed - this is a bug.  Redirecting the debugees output to the debug console is not yet implemented.

jeromelaurens:
@lukehoban ok , some bugs of course but you've done a great job so far !

rudygt:
[...]

PR #41, opened 2015-11-20
Redirect debug process output to debug window

rudygt:
Hi, 

This is my try on the issue #15, @lukehoban can you please review, let me know if any extra change is required. 

Thanks.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -2,7 +2,7 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


lukehoban:
Thanks for this.  I merged in @yuntan's change that addresses the same issue, so I don't think we'll need this PR as well.

One question - I see the `sanitize` function you added in this PR, but it is not being used.  I'm also not seeing any problems with newline rendering in my own tests.  Is there a case you don't see working correctly with just the changes that have been merged in so far?

rudygt:
Nothing in particular, it was a mistake to leave it there.

Thanks.

PR #43, opened 2015-11-20
debugAdapter: Redirect stdout/err of Delve to console

### LINKED ISSUE #15: where is console output (stdout)

yuntan:
Redirect stdout/err output of Delve to debug console

Fixes #15


lukehoban:
Thanks @yuntan.

Issue #19, opened 2015-11-18
Analysis Tools Missing warning is shown incorrectly when multi-workspace GOPATH is used


lukehoban:
The logic for determining whether all tools are available is not taking into account that GOPATH may contain multiple workspaces.

See discussion in #10.

mishak87:
[There](https://github.com/Microsoft/vscode-go/blob/master/src/goPath.ts#L78-L83) should be loop checking for each path in [`getGOPATHWorkspaces`](https://github.com/Microsoft/vscode-go/blob/master/src/goPath.ts#L30).

ulrichSchreiner:
hi,

i'm currently working on a docker-container ([vsc](https://github.com/ulrichSchreiner/vsc), somthing like [atocker](https://github.com/ulrichSchreiner/atocker)) which bundles all binaries in the container and starts vscode. The `GOPATH` is the current working directory of the host and the `docker run ...` mounts this directory as a `GOPATH` to the container. 

It would be great to fix this bug to support multiple `GOPATH` entries in your `setupGoPathAndOfferToInstallTools` function so i can create a `GOPATH` in the container with two parts (one for the source, the second for the tools). So all the tools would be hidden from "outside" the container.

Imho it would be more better to have all the build-tools in the `PATH`, not in the `GOPATH`. It would be much easier to separate the build-tools from the workspaces. Would this be an option?

mishak87:
@ulrichSchreiner When this is fix you can always install them globally manually.

ulrichSchreiner:
@mishak87 I don't think so. All the tools are referenced via GOPATH entries in the `bin` subdirectory, never in the `PATH`. So "install them globally manually" won't help if they won't be found :-)

mishak87:
@ulrichSchreiner Gotcha! I will try to make pull request over weekend to get this fixed.

lukehoban:
Thanks to @pcmind for PR #46 which fixed this issue.

lukehoban:
@ulrichSchreiner and @mishak87 Good scenario.  Would love a PR to support this.  Feel free to open another issue as well if you'd like to continue to track this.

@ulrichSchreiner To your first point, we preferred GOPATH/bin over PATH just because we saw a number of developer setups where PATH was not configured to include GOPATH/bin.  That said, perhaps it makes sense to search PATH as well when looking for the tools?  Probably as a fallback after looking in the workspaces GOPATH/bin.

[...]

PR #46, opened 2015-11-20
Fix gopath bin location search and error in error line parsing in windows

pcmind:
I request to merge to correction tested for windows plugin usage. 
First bin path extraction from gopath should be consistent, and in windows it is required to test file with it's extension
Build tool will emit errors like: myfile.go:4:6: expected 'STRING', found newline
where 4 is the line number and not 6

### DIFF
--- a/src/goPath.ts
+++ b/src/goPath.ts
@@ -14,6 +14,7 @@ import { showGoStatus, hideGoStatus } from './goStatus'
 var binPathCache : { [bin: string]: string;} = {}
 
 export function getBinPath(binname) {
+	binname = correctBinname(binname)
 	if(binPathCache[binname]) return binPathCache[binname];
 	var workspaces = getGOPATHWorkspaces();
 	var binpath: string;
@@ -25,6 +26,13 @@ export function getBinPath(binname) {
 	}
 	console.log("Couldn't find a binary in any GOPATH workspaces: ", binname, " ", workspaces)
 	return path.join(process.env["GOPATH"], "bin", binname);
+}
+
+function correctBinname(binname) {
+	if (process.platform === 'win32')
+		return binname + ".exe";
+	else
+		return binname
 }
 
 function getGOPATHWorkspaces() {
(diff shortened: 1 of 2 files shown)


lukehoban:
Looks good - thanks for the fixes.

lukehoban:
Fixes #19.

src/goPath.ts
```diff
@@ -27,6 +28,13 @@ export function getBinPath(binname) {
 	return path.join(process.env["GOPATH"], "bin", binname);
```
mishak87: `workspaces` will always be of `.length > 0` and if no bin is found returning `ws1:ws2/bin/name` makes no sense.

lukehoban: What would you suggest instead?  Perhaps just return `binname` and rely on finding it in `PATH`?

src/goPath.ts
```diff
@@ -14,6 +14,7 @@ import { showGoStatus, hideGoStatus } from './goStatus'
 var binPathCache : { [bin: string]: string;} = {}
 
 export function getBinPath(binname) {
+	binname = correctBinname(binname)
```
mishak87: Correcting should be done after checking cache.

src/goPath.ts
```diff
@@ -14,6 +14,7 @@ import { showGoStatus, hideGoStatus } from './goStatus'
 var binPathCache : { [bin: string]: string;} = {}
```
mishak87: `binPathCache` is not used?

src/goPath.ts
mishak87: Leaving it on `PATH` would make sense.

src/goPath.ts
lukehoban: Indeed - looks like we've never been writing to the cache.  We should either do that or remove the cache code for now pending figuring out how much of a perf hit this codepath is.

src/goPath.ts
lukehoban: Agreed.

PR #59, opened 2015-11-22
[WIP] Format on save

### LINKED ISSUE #14: Add format-on-save support (instead of requiring use of explicit Format Document command)
The gopath seems to be correctly set because typing errors ('Undefined: ...') en vet messages ('exported function should have comment') are reported correctly.
However, the formatting is not corrected ? Not on disk and not in the editor. I tried to override the default goreturns with goimports and then with gofmt, but that does not seems to matter.
Any ideas ?

lukehoban:
Work on format on save.

Fixed #14.


lukehoban:
Closing this in favor of #115.

Issue #31, opened 2015-11-19
Add support for "on the fly compilation" even when Auto Save is off


newhook:
My gomode plugin for go-sublime has support for "on the fly compilation". Namely, the gocode is compiled as I type which is super useful.

https://github.com/newhook/gomode

It would be really cool if something like this could work with this plugin.

lukehoban:
The suggestion in the README is to turn on `Auto Save` in Visual Studio Code, which saves your code as you type.  It's a generally nice feature once you get used to it, but especially nice with the vscode-go extension as you then get on-the-fly error reporting.

I hadn't see `goflymake` before - but that looks like it might make it possible to do this without relying on build-on-save + auto-save.  If that works well, I'd be open to moving error checking away from the 'on-save' hook and to an on-th-fly model that works even with auto-save turned off.

newhook:
Basically what my sublime plugin does is run flymake when the file changes, and then not for a minimum of 5s (to balance responsiveness of the editor with getting error messages when typing). It works quite well. if you have questions or need help with the impl let me know.

lukehoban:
Thanks @newhook.  That sounds great.  

BTW - Definitely also open to PRs for this if you are interested.

newhook:
I've been working on this and have it basically working :) However, one thing i ran into with the latest master is that if have:

foo_test.go

and then type

package foo_test

this ends up creating a bunch of files
f.test
fo.test
foo.test

and so forth. Is this a known issue?

newhook:
One other thought: With support for flymake it isn't necessary or even desirable to use auto-save any longer. In fact, I would recommend against it. This means that its easy then to support gofmt (or goimport or whatever) on save (which is an open issue if I'm not mistaken).

lukehoban:
@newhook I haven't seen the first issue you mention.  Could you open another issue to track this with details on how to repro it.  From a quick test of what you mention I didn't see those files being generated.

On the second point I completely agree.  I think doing both of these features together could make sense.

Feel free to open up a WIP PR when you have something working here.

lukehoban:
Per gitter discussion, we'll go ahead and close this for now, since we're not ready to move over to a model where Go tools that need global understanding can operate on in-memory state yet. 

We'll likely want to revisit this in the future.

PR #62, opened 2015-11-22
Support for flymake.

newhook:
This is a preliminary PR. This requires https://github.com/newhook/goflymake from my repo, as the original does not support multiple flymake files.

Please check for basic structure and whether you think this is acceptable in general before I fix the  below.

Things I know which require work:
- auto build and flymake are incompatible (but not lint/vet), but I didn't implement that (yet).
- didn't add a flymake config option.
- The path to goflymake is hard coded (the repo doesn't yet work correctly with multi-path GOPATH).
- Only tested on OSX.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -37,7 +37,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	ctx.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(showHideStatus));
 	setupGoPathAndOfferToInstallTools();
 	ctx.subscriptions.push(startBuildOnSaveWatcher());
-	
+	ctx.subscriptions.push(flymakeWatcher());
 	ctx.subscriptions.push(vscode.commands.registerCommand("go.gopath", () => {
 		var gopath = process.env["GOPATH"];
 		vscode.window.showInformationMessage("Current GOPATH:" + gopath);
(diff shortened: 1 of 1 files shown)


newhook:
I remember an issue on this, but the linting doesn't show errors from other sources files which I find less than desirable. This is easily fixed:

Change:
cp.execFile(golint, [filename], {cwd: cwd}, (err, stdout, stderr) => {

to:

cp.execFile(golint, [], {cwd: cwd}, (err, stdout, stderr) => {

lukehoban:
Thanks!

Generally this looks like a good start.  I'm open to moving the error checking over to this approach if we can get this as complete as the current solution.

Few things:

I'm seeing the `flymake_*` files being emitted next to the real source files.

When I make edits in `_test.go` files, I stop seeing any errors either in the `_test.go` file or in any of the other files (even ones that were displaying errors previously).

When I tried this on the `github.com/derekparker/delve` codebase (modifying the `proc/prog.go` file), I don't see any errors being reported but see this on stdout of the `goflymake` command:

```
      __cgo_d5684f8859a6_Cfunc_raise_exception in proc_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_raise_exception)
  "_read_memory", referenced from:
      __cgo_d5684f8859a6_Cfunc_read_memory in threads_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_read_memory)
  "_resume_thread", referenced from:
      __cgo_d5684f8859a6_Cfunc_resume_thread in threads_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_resume_thread)
  "_set_pc", referenced from:
      __cgo_d5684f8859a6_Cfunc_set_pc in registers_darwin_amd64.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_set_pc)
  "_set_registers", referenced from:
      __cgo_d5684f8859a6_Cfunc_set_registers in registers_darwin_amd64.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_set_registers)
  "_single_step", referenced from:
      __cgo_d5684f8859a6_Cfunc_single_step in threads_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_single_step)
  "_thread_blocked", referenced from:
      __cgo_d5684f8859a6_Cfunc_thread_blocked in threads_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_thread_blocked)
  "_thread_count", referenced from:
      __cgo_d5684f8859a6_Cfunc_thread_count in proc_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_thread_count)
  "_write_memory", referenced from:
      __cgo_d5684f8859a6_Cfunc_write_memory in threads_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_write_memory)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Are you aware of any other limitations or differences in behaviour in terms of when real errors will/wont be displayed after this change?

newhook:
Cool, thanks for looking at that!

- For the flymake_ files I'm not sure what can reasonably be done here. New files have to be created in the same directory as the package otherwise go doesn't like that and we cannot overwrite the existing files. The only option I can see here is if the editor itself can supress showing these files?

- I cannot duplicate the issue that you pointed out with _test.go files. I only tested this with go 1.4, however.

- I duplicated the issue with delve and cgo. I haven't looked into this in great detail yet, however, this shows the same issue.

```
➜  proc git:(master) ✗ go build .
```

vs:

```
➜  proc git:(master) ✗ go build arch.go breakpoints.go doc.go eval.go go_version.go proc.go proc_darwin.go ptrace_darwin.go registers.go registers_darwin_amd64.go threads.go threads_darwin.go variables.go
# command-line-arguments
Undefined symbols for architecture x86_64:
  "_acquire_mach_task", referenced from:
      __cgo_d5684f8859a6_Cfunc_acquire_mach_task in proc_darwin.cgo2.o
     (maybe you meant: __cgo_d5684f8859a6_Cfunc_acquire_mach_task)
```

lukehoban:
Per https://github.com/Microsoft/vscode-go/issues/31#issuecomment-159697233 I'm going to close out the PR for now.  Feel free to reopen if you want to keep working on this.

Issue #71, opened 2015-11-23
gorename: wrong path logic


newhook:
https://github.com/Microsoft/vscode-go/blob/master/src/goRename.ts#L25

lukehoban:
Fixed with #72.

PR #72, opened 2015-11-23
gorename: fix incorrect path logic.

newhook:
Also don't return a null thenable if gorename isn't found.

https://github.com/Microsoft/vscode-go/issues/71

### DIFF
--- a/src/goRename.ts
+++ b/src/goRename.ts
@@ -22,17 +22,17 @@ export class GoRenameProvider implements vscode.RenameProvider {
 			var filename = this.canonicalizeForWindows(document.fileName);
 			var offset = document.offsetAt(position);
 
-			var gorename = path.join(process.env["GOPATH"], "bin", "gorename");
+			var gorename = getBinPath("gorename");
 
 			cp.execFile(gorename, ["-offset", filename + ":#" + offset, "-to", newName], {}, (err, stdout, stderr) => {
 				try {
 					if (err && (<any>err).code == "ENOENT") {
 						vscode.window.showInformationMessage("The 'gorename' command is not available.  Use 'go get golang.org/x/tools/cmd/gorename' to install.");
-						return resolve(null);
+						return Promise.resolve<vscode.WorkspaceEdit>(null);
 					}
 					if (err) return reject("Cannot rename due to errors: " + err);
 					// TODO: 'gorename' makes the edits in the files out of proc.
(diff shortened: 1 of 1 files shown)


src/goRename.ts
```diff
@@ -22,17 +22,17 @@ export class GoRenameProvider implements vscode.RenameProvider {
...
 
-			var gorename = path.join(process.env["GOPATH"], "bin", "gorename");
+			var gorename = getBinPath("gorename");
 
 			cp.execFile(gorename, ["-offset", filename + ":#" + offset, "-to", newName], {}, (err, stdout, stderr) => {
 				try {
 					if (err && (<any>err).code == "ENOENT") {
 						vscode.window.showInformationMessage("The 'gorename' command is not available.  Use 'go get golang.org/x/tools/cmd/gorename' to install.");
-						return resolve(null);
+						return Promise.resolve<vscode.WorkspaceEdit>(null);
```
lukehoban: Actually both of these should be `resolve(null)`.  The `return` itself doesn't matter really (just short circuits the exit from the callback).  The call to the `resolve` from the `new Promise()` above is what ensures a value is being returned from the promise.  (aside - moving to async/await will dramatically improve a lot of this code - hoping to do that soon now that TS supports this with the same Node version as VS Code uses)

newhook: If I'm not mistaken this causes an error in the editor (you get a big red error saying then isn't found on undefined).

lukehoban: I think the bug in the current code is in the second one of these on L36.  I haven't seen this error - do you have steps to reproduce this?

newhook: The original code caused that for me (since gorename isn't in my GOPATH/bin. It would would go into the block at line 29 and then return resolve(null).

lukehoban: Huh - yeah, I see what you are saying.  Calling `resolve` here is definitely the right thing to be doing, but VS Code doesn't seem to be handling the returned promise correctly.  @egamma Any idea what's going on here?

newhook: Perhaps in the meantime the patch could be merged as it fixes an issue :) I could add a comment that this should work but does not if you like.

lukehoban:
Agreed - I merged this as it net makes things better.  It seems the result should be that the "rename" operation is technically never completing (in both error and success cases).

PR #75, opened 2015-11-23
initial oracle support using codeLens #8

tecbot:
Make it possible to show CodeLens using the oracle command.
Currently, only the implements mode is implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.lintOnSave": true,
 	"go.vetOnSave": true,
 	"go.formatTool": "goreturns",
+	"go.activateCodeLens": false,
 	"go.gopath": "/Users/lukeh/go"
 }
 ```
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This extension adds rich language support for the Go language to VS Code, includ
 - Build-on-save (using `go build` and `go test`)
 - Format (using `goreturns` or `goimports` or `gofmt`)
 - [_partially implemented_] Debugging (using `delve`)
+- [_partially implemented_] CodeLens (using `oracle`)
 
 ### IDE Features
 ![IDE](http://i.giphy.com/xTiTndDHV3GeIy6aNa.gif)
@@ -109,6 +110,7 @@ The extension uses the following tools, installed in the current GOPATH.  If any
 - go-outline: `go get -u -v github.com/lukehoban/go-outline`
 - goreturns: `go get -u -v sourcegraph.com/sqs/goreturns`
 - gorename: `go get -u -v golang.org/x/tools/cmd/gorename`
+- oracle: `go get -u -v golang.org/x/tools/cmd/oracle`
 
 To install them just paste and run:
 ```bash
@@ -119,6 +121,7 @@ go get -u -v github.com/lukehoban/go-find-references
 go get -u -v github.com/lukehoban/go-outline
 go get -u -v sourcegraph.com/sqs/goreturns
 go get -u -v golang.org/x/tools/cmd/gorename
+go get -u -v golang.org/x/tools/cmd/oracle
 ```
 
 And for debugging:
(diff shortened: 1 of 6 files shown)


lukehoban:
Great! Thanks for looking at this.  Really excited about this.

A few  initial thoughts from trying out the current code:

1. Could you add a new configuration setting in `package.json` and put this feature behind a feature flag?  That will make it easier to merge early without risking the default editing experience.
2. When I use this and then edit my source file I see it get out-of-date and then get stuck showing incorrect information.
3. It will be interesting to discuss what lenses make sense for Go, and which to display where.  For example, I can imagine that `implements` I can imagine it may make sense to show all the implementors of an interface on the interface itself, and the interfaces implemented by a struct on the struct, but may not make sense to put all of the interface method information at the method level.  But it will be interesting to see what info is available and then make some decisions on this.
4. Do you have a list in mind of the Oracle features you'd like to show in CodeLens?

newhook:
Can resolve the state information in the same way I did with flymake? Write a temporary file with the current editor content and run oracle on that?

newhook:
The right hands side list seems to be off by one. It should show the type names, not the line _after_ the type names?

oracle

newhook:
I've asked for clarification from Alan Donovan on how we can get more complete information from oracle implements.

One side effect of this change is all sorts of additional empty lines in the editor before type defintions. Is there anything that can be done about that?

lukehoban:
> One side effect of this change is all sorts of additional empty lines in the editor before type defintions. Is there anything that can be done about that?

I don't think we'll want to turn this feature on until there is useful information in the CodeLens display for every top level declaration.  At that point, there shouldn't be any empty lines, but there will be decorations on each declaration.

Here's what it looks like for C# currently in Code.  I think the "references" provider is one we'd ideally want to add early as it applies meaningfully to every declaration (and we have the find-all-reference feature that can be triggered when clicking it).  I am not sure though whether finding references is fast enough to use in this context.

codelens

lukehoban:
@tecbot While I sort of agree about this being a failure of Code itself, I actually think that it makes sense to focus on providing information that will make sense to display on all declarations.  If we are going to show CodeLens at all, it makes sense for it to be consisentnly applied on all declarations and to show generally useful information.

I'm honestly still not sure what set of things from Oracle will make sense to display in this sort of UX vs. in a more on-demand UX like Pythia.

newhook:
This is why implements is a bit brain dead.

https://github.com/golang/go/issues/13457

newhook:
Apparently the Oracle bug is fixed.

newhook:
So with the oracle bug fix implements now can take quite some time.

oracle -pos=$GOROOT/src/io/io.go:\#9401 implements  27.02s user 2.11s system 425% cpu 6.845 total

lukehoban:
@newhook Yeah - that's unfortunate.  Sounds like that may be too slow to use in this particular UX.  We could still think about some other UX for exposing this command?  

I'm also still generally curious on what other things make sense to show in the CodeLens UX.

newhook:
I haven't tried it in my large code base since the fix. I'll give it a try tomorrow.

I'd really like support for callers. Where do you think that makes sense? A right-click menu?

lukehoban:
To be clear, I'd love to try it out and see what it feels like to put these things in CodeLens.  If they can work there, the UX is a really nice way to provide this info in context.  

If we decide they don't work there, we'd have to go with a command accessible through the palette and keybindings (and possibly context menu if that's possible yet?).

ironcladlou:
This is probably worth revisiting with the new `guru` tool (see https://github.com/Microsoft/vscode-go/issues/284).

mattetti:
I believe this feature was implemented using the contextual menu > Find all references, maybe we should close this PR (which is quite outdated).

ironcladlou:
>I believe this feature was implemented using the contextual menu > Find all references, maybe we should close this PR (which is quite outdated).

Agree that this should be closed as inactive.

ramya-rao-a:
Closing as this is outdated

PR #76, opened 2015-11-23
Fix for unicode issue with gocode

zsxawerdu:
Fixes this issue:

1. go get gopl.io/ch1/fetchall
2. Fails to autocomplete with text 'fmt.'
3. Simple off by one.

I tried this on a few other projects after this change, and auto complete is working as expected.


lukehoban:
Aha!  I don't think this is a correct fix, but I think it points at a root cause of a problem several others have seen.

The file in question has a unicode character early on:

```go
// Copyright © 2016 Alan A. A. Donovan & Brian W. Kernighan.
// License: https://creativecommons.org/licenses/by-nc-sa/4.0/
```

I think the problem is that we are not passing the `c` prefix to gocode per this: https://github.com/nsf/gocode/issues/52#issuecomment-1992781.

lukehoban:
Could you try changing this:

```
var p = cp.execFile(gocode, ["-f=json", "autocomplete", filename, "" + offset], {}, (err, stdout, stderr) => {
```

to this:
```
var p = cp.execFile(gocode, ["-f=json", "autocomplete", filename, "c" + offset], {}, (err, stdout, stderr) => {
```

instead?  If that works, can you update the PR to use that change instead?

zsxawerdu:
Tried on 4 projects, all work.  Good find luke.

PR #99, opened 2015-11-25
[WIP] Experimental work on SignatureHelper

zsxawerdu:


src/goSignature.ts
```diff
@@ -0,0 +1,100 @@
...
+		// TODO: make this clear
+		var ret = [];
+		for(var results = wordPattern.exec(searchString); results != null; results = wordPattern.exec(searchString)) {
+			ret.push(results[0])
+		} 
+
+		return ret;
+	}
+	
+	public lastParentheses(document: TextDocument, position: Position): Position {
```
lukehoban: Will this handle previous arguments that contain parens?  I would expect some balanced paren matching walking backward through the document here. 

Something like `foo(bar(), |)`.

src/goSignature.ts
```diff
@@ -0,0 +1,100 @@
...
+		// TODO: handle double '(('
+		var currentLine = document.lineAt(position.line).text.substring(0,position.character);
+		var lastIndex = currentLine.lastIndexOf("(");
```
lukehoban: I think you'll need to not require things to all be on the same line.

lukehoban:
@zsxawerdu Checking in on whether you still expect to do more work on this?  If not, I'll likely look at this myself in the next couple days.

zsxawerdu:
Feel free to, I am out of town for a few days.

lukehoban:
I took another look at this, and ended up reworking the changes in https://github.com/Microsoft/vscode-go/tree/feature/sighelp.  

However, in looking at that, I am more firmly of the belief that we'll need to build a tool in Go to really parse the AST surrounding the call to make sure we accurately determine which function is being called and where we are in the parameter list.  It just won't be robust enough to try to approximate the lexing and parsing of Go in the TypeScript side of the extension.

So I think this requires a more complex solution.  I'm going to close out this PR since I don't think we'll be able to pursue this solution.  But I'm also going to hold off on opening a PR from the branch until we have (or create) a Go tool for doing the signature help analysis.

PR #111, opened 2015-11-27
Gofmt patch

newhook:
Use diff-match-patch to apply gofmt as a series of edits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
 		"vscode.go"
 	],
 	"dependencies": {
-		"json-rpc2": "^1.0.2"
+		"json-rpc2": "^1.0.2",
+		"diff-match-patch": "~1.0.0"
 	},
 	"devDependencies": {
 		"vscode": "0.10.x",
--- a/src/goFormat.ts
+++ b/src/goFormat.ts
@@ -7,6 +7,7 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
+import dmp = require('diff-match-patch');
 import { getBinPath } from './goPath'
 
 export class GoDocumentFormattingEditProvider implements vscode.DocumentFormattingEditProvider {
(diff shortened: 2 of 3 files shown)


lukehoban:
Two thoughts while looking at this - I don't think either need changes in the PR, but things to think about:

1. For performance reasons, it seems likely that if we could rely on `gofmt -d` to produce the diff, then parse and process that instead of recomputing the diff from full source code, we might be avoiding duplicating work that is already happening inside `gofmt`.  It looks like perhaps `diff_match_patch` can do this for us?  I'm not 100% sure that rendering+parsing the patch file is  faster than recomputing the diff though, so not sure this will be a win or not - and haven't seen perf-sensitive cases with formatting yet.
2. I had hoped that providing the individual TextEdits would have let Code preserve the cursor position correctly.  It looks like it doesn't.  I've opened https://github.com/Microsoft/vscode/issues/755 to see if that's intentional or not.  Especially for when we have format-on-save, it will be critical to leave the cursor in the correct location.

newhook:
Yeah, what gofmt -d does is generate the full text for both revisions and then run "diff -u" on them, so I'm not sure its that much different.

I also wrote the change initially to take the gofmt -d output and then process the patch... there might be a simpler way to do this but using diff_match_patch means you need to create the patches from the text, then process the patch on the document text, then get a set of diffs and convert the diffs to edits which is more work that this implementation :) I, however, does show how a gorename impl might work.

I've attached the implementation here.

[mkpatch.txt](https://github.com/Microsoft/vscode-go/files/45982/mkpatch.txt)

lukehoban:
Got it - that makes sense then to do the processing in the extension.

lukehoban:
I did I noticed at least one regression where files with spaces instead of tabs at the beginning of any line cause no formatting to be applied at all.

I'm pretty sure this is another bug in Code's processing of TextEdit's. But we may need to figure out how to workaround it before doing a release with these updates.

newhook:
Thats annoying, I can take a look at that. I know I was testing some edits like:
```
// line, character
insert(0, 0, "hi")
insert(0, 0, "hi")
```
And that appears to be illegal, whereas:
```
insert(0, 0, "hi")
insert(0, 1, "hi")
```
Worked. When you try an illegal edit it just doesn't work. No feedback, no nothing.

lukehoban:
Yeah - it looks like one thing that might work is when there is a `delete` followed by an `insert` where the end range of the delete is the same as the start range of the insert, replace the original delete with a `replace` instead of adding the `insert` following the `delete`.

A quick test showed that working.

newhook:
Re: the cursor until (or if) code fixes this we could compute the new offset?

lukehoban:
Yes - it's easy enough to compute. The slight challenge may be in correctly moving the cursor.  The formatting API is passed a Document not a TextEditor. Could probably just cheat a little and use vscode.window.activeEditor at least until/unless this is fixed in Code.

newhook:
Hopefully fixed the formatting regression with https://github.com/Microsoft/vscode-go/pull/112

lukehoban:
@newhook I hit one more regression from this, which I fixed with 44601a58b3075d5bd22a14354d8cffb75baf573e.  Can you check that to make sure the change makes sense?

Here's the repro case:

```go
package main

import (
    "net/http"  
)

func main() {
	http.ListenAndServe(":8080", http.FileServer(http.Dir("/usr/share/doc")))
}
```

The `"new/http"` line has four space in front and two trailing spaces.  Previously it would be changed to `\t"net/hp"` because the delete at the end of the line was not tracking it's location correctly after the replace earlier on the same line.

newhook:
Nice catch, the lesson here is that you cannot look at gofmt -d output to see the diff_match_patch diff patterns :)

I tested this and haven't been able to break it yet.

PR #112, opened 2015-11-28
Fix regression caused by vscode bug

newhook:
https://github.com/Microsoft/vscode/issues/761


newhook:
The basic idea here is to combine deletes into deletes. delete-insert into replace. And inserts into inserts.

lukehoban:
Great - thanks!

PR #114, opened 2015-11-28
Introduce command to run test at cursor

### LINKED ISSUE #88: Support running unit test at cursor
Add support for for running a unit test at the cursor.

As I'm still learning the VSCode and vscode-go internals, I'm a bit fuzzy on whether this would be implemented as a command or task.

Split off from #21.

ironcladlou:
Introduce a 'go.test.cursor' command which executes the unit test at
the primary cursor location. Output is placed in the 'Go' channel.

Refactor references to the Go runtime path.

Closes #88

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -9,17 +9,9 @@ import cp = require('child_process');
 import path = require('path');
 import os = require('os');
 import fs = require('fs');
-import { getBinPath } from './goPath'
+import { getBinPath, getGoRuntimePath } from './goPath'
 
-//TODO: Less hacky?
-var go: string;
-if (process.env.GOROOT) {
-	go = path.join(process.env["GOROOT"], "bin", "go");
-} else if (process.env.PATH) {
-	var pathparts = (<string>process.env.PATH).split((<any>path).delimiter);
-	go = pathparts.map(dir => path.join(dir, 'go' + (os.platform() == "win32" ? ".exe" : ""))).filter(candidate => fs.existsSync(candidate))[0];
-}
-if (!go) {
+if (!getGoRuntimePath()) {
 	vscode.window.showInformationMessage("No 'go' binary could be found on PATH or in GOROOT.");
 }
 
@@ -38,7 +30,7 @@ export function check(filename: string, buildOnSave = true, lintOnSave = true, v
 		if (filename.match(/_test.go$/i)) {
 			args = ['test', '-copybinary', '-o', tmppath, '-c', '.']
 		}
-		cp.execFile(go, args, { cwd: cwd }, (err, stdout, stderr) => {
+		cp.execFile(getGoRuntimePath(), args, { cwd: cwd }, (err, stdout, stderr) => {
 			try {
 				if (err && (<any>err).code == "ENOENT") {
 					vscode.window.showInformationMessage("The 'go' compiler is not available.  Install Go from http://golang.org/dl/.");
@@ -91,7 +83,7 @@ export function check(filename: string, buildOnSave = true, lintOnSave = true, v
 
 	var govet = !vetOnSave ? Promise.resolve([]) : new Promise((resolve, reject) => {
 		var cwd = path.dirname(filename)
-		cp.execFile(go, ["tool", "vet", filename], { cwd: cwd }, (err, stdout, stderr) => {
+		cp.execFile(getGoRuntimePath(), ["tool", "vet", filename], { cwd: cwd }, (err, stdout, stderr) => {
 			try {
 				if (err && (<any>err).code == "ENOENT") {
 					vscode.window.showInformationMessage("The 'go tool vet' compiler is not available.  Install Go from http://golang.org/dl/.");
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -1,0 +1,82 @@
+'use strict';
+
+import cp = require('child_process');
+import path = require('path');
+import vscode = require('vscode');
+import util = require('util');
+import { getGoRuntimePath } from './goPath'
+
+/**
+* Executes the unit test at the primary cursor using `go test`. Output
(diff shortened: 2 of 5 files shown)


ironcladlou:
This is my first time using VS Code or TypeScript or the vscode-go plugin, so feedback on anything and everything is greatly appreciated as I learn the ropes.

src/goTest.ts
```diff
@@ -0,0 +1,78 @@
...
+				return;
+			}
+		});
+		if (!testFunction) {
+			vscode.window.setStatusBarMessage('No test function found at cursor.', 5000);
+			return;
+		}
+		// Run the test and present the output in a channel.
+		// TODO: The timeout should probably be configurable somehow. An input
+		// popup seemed intrusive to ask every time.
```
lukehoban: You could add a configuration setting in `package.json` to let users configure this.

src/goCheck.ts
```diff
@@ -10,16 +10,9 @@ import path = require('path');
 import os = require('os');
 import fs = require('fs');
 import { getBinPath } from './goPath'
+import { getGoRuntimePath } from './goPath'
```
lukehoban: Can combine this into the previous line as an additional import from `goPath`

src/goTest.ts
```diff
@@ -0,0 +1,78 @@
...
+* so output doesn't produce navigable line references.
+*/
+export function testAtCursor() {
+	var editor = vscode.window.activeTextEditor;
+	// TODO: This command should be returning a SymbolInformation[] but is instead
+	// returning an object with slightly different fields which must be adapted.
+	vscode.commands.executeCommand<any[]>('vscode.executeDocumentSymbolProvider', editor.document.uri).then(res => {
+		var testFunction: string;
+		// Find any test function containing the cursor.
+		res.forEach(obj => {
```
lukehoban: Perhaps a regular for loop so you can break out early once you find the right function?

src/goTest.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+* Executes the unit test at the primary cursor using `go test`. Output
+* is sent to the 'Go' channel.
+* 
+* TODO: go test returns filenames with no path information for failures,
+* so output doesn't produce navigable line references.
+*/
+export function testAtCursor() {
+	var editor = vscode.window.activeTextEditor;
```
lukehoban: I think this can be null/undefined (if there is no file open). Should test for that and render warning.

src/goTest.ts
ironcladlou: I do break early using the `return` on L32, but I can switch to a normal for loop if that's more clear.

src/goTest.ts
ironcladlou: Done

src/goCheck.ts
ironcladlou: Done

src/goTest.ts
ironcladlou: Done

ironcladlou:
@lukehoban Feedback addressed, PTAL. If we get to a point where you want to merge, please let me know first so I can squash commits.

I'm going to take a close look at the WallabyJS plugin per https://github.com/Microsoft/vscode-go/issues/88#issuecomment-160276790 for presentation ideas. How advanced would you like the UX before merging?

src/goTest.ts
lukehoban: That return won't break early since it just returns from the callback for that particular item in the collection.  There is no way to break early from a `.forEach` in JavaScript unfortunately.  I believe you can use `for..of` in TS instead here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

src/goTest.ts
```diff
@@ -0,0 +1,85 @@
...
+* TODO: go test returns filenames with no path information for failures,
+* so output doesn't produce navigable line references.
+*/
+export function testAtCursor(timeout: string) {
+	var editor = vscode.window.activeTextEditor;
+	if (!editor) {
+		vscode.window.showInformationMessage("No editor is active.");
+		return;
+	}
+	if (!timeout) timeout = '30s';
```
lukehoban: Curious - is this necessary?  I would have expected the default set in package.json would be provided in the configuration value you get back.

ironcladlou: I'm probably being overly defensive/paranoid. The config value isn't being strongly validated.

src/goTest.ts
ironcladlou: Ohhh, I see. Thanks!

ironcladlou:
Rebased and addressed latest feedback. Thanks for the review!

lukehoban:
Thanks @ironcladlou!  Just did a final review and pulled these changes in.

I made a couple of updates on top of your changes in 36dbc1d92770dca55c2efc88ea13ef2359cd827a.  Let me know if there's anything there that you don't like form a workflow perspective.  (In particular, opening test results to the right by default?  Code tends to do that in many other cases, but it's not for everyone).

Do you plan to add a similar `Go: Run tests in current package` and `Go: Run tests in workspace` that run all tests in the package where the active document lives and in all packages underneath the current workspace respectively?

ironcladlou:
@lukehoban Thanks for the merge!

>I made a couple of updates on top of your changes in 36dbc1d. Let me know if there's anything there that you don't like form a workflow perspective. (In particular, opening test results to the right by default? Code tends to do that in many other cases, but it's not for everyone).

Love it.

>Do you plan to add a similar Go: Run tests in current package and Go: Run tests in workspace that run all tests in the package where the active document lives and in all packages underneath the current workspace respectively? 

Yeah, I was thinking of doing "test in current package" next, then "all tests in package/workspace". "All tests in file" would also be useful but I'd like to figure out if my only option is to build up a regex for `-run`.

PR #122, opened 2015-11-29
Document go.formatOnSave option in readme

kylebrandt:


lukehoban:
Thanks @kylebrandt.  For now I've mentioned this flag in the README with `false` as the default since that's still what's in the current releases.  

I'll add more text on the two modes of use instead of the current suggestion about `AutoSave` soon to make it even more clear how to set up for format-on-save.

PR #123, opened 2015-11-29
Introduce more unit test commands

### LINKED ISSUE #89: Support running all tests in file
Add support for for running all tests defined in the current file.

As I'm still learning the VSCode and vscode-go internals, I'm a bit fuzzy on whether this would be implemented as a command or task.

Split off from #21. Would build upon #88.

ironcladlou:
Introduce 'go.test.package' and 'go.test.file' commands.

Closes #89.


lukehoban:
Thanks - this looked good - just added a little additional promise error propagation code.

PR #69, opened 2015-11-23
[WIP] Adding "Go: Build", "Go: Test" and "Go: Install" commands

lukehoban:
Fixes #21.

Issue #21, opened 2015-11-18
Add commands for executing `go build`, `go test` and `go install`


ppone:
On OS X 11.11.1, most of the Go commands seem to be working, however `Go Build` on save does not work.  

Go Path seems to be correct.

ppone:
If there is an error in the code I see red squiggles, so yes if the code has errors I see them in the IDE.

I was under the impression it would generate a binary with Go Build, but your saying that it is working correctly as Build on Save lets you know if there are errors.  

It would be great if "Go for Visual Studio Code" could include those tasks like Go Build, etc like GoSublime on SublimeText.

Off topic, VS Code is fast.  It is up there IMO with SublimeText.

Thanks,

Parag

lukehoban:
@ppone Makes sense.  Let's keep this issue to track adding more integration with `go build`, `go install`, `go run` and `go test`.

There are likely two ways to include these:

1. Add new commands to the command palette for `Go: Build`, `Go: Test`, etc. which operate either in the context of the opened workspace or the current file?
2. Provide a command which automatically generates VS Code tasks for these in a `.vscode/tasks.json` in your current workspace.

The second is more native to VS Code, but likely the first is a convenient shortcut to this.

Thoughts?

ppone:
The first option I think will the best for users.

The beauty of a ideal IDE is that CMD+Shift+P would all of the Go tooling, such that you never need to leave the IDE.

ahmetalpbalkan:
+1 for first option

Mistermatt007:
Why not providing the two options ? Some users coming from Sublime Text will find the first option more intuitive, and the second will be more useful for those who master VSCode like pros !

lukehoban:
@Mistermatt007 Agreed - I think both can be supported.

pootow:
How to vote up this feature?

lukehoban:
Question for folks looking for (1).  

Should `Go: Build` invoke the build in the root of the workspace that is opened in Code, or in the folder containing the file currently open in the active text buffer?  Both feel potentially wrong in some set of cases.

Mistermatt007:
Maybe when you invoke the command, a second menu should ask you what to build when it's ambigous

lukehoban:
Well it will frequently be ambiguous and I expect a second menu is more friction than is desired in this workflow. Option (2) is also available for full configuration of build/test tasks.

My inclination is to go with the active document's folder (even if it is not a .go file).

[...]

PR #134, opened 2015-12-01
[WIP] Add a go install command

ironcladlou:
### :construction: WIP :construction: 
Introduce a 'go.install' command which executes 'go install' for a
configurable list of packages.

Related to #21 and #69.

Outstanding design questions:

* Can file references in channel output be made clickable? I didn't see any way to achieve it in the API.
  * Note that the file references emitted by `go install` are relative to the workspace root
* Should this integrate with the diagnostics collection?

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,11 @@
 				"command": "go.test.file",
 				"title": "Go: Run tests in file",
 				"description": "Runs all unit tests in the current file."
+			},
+			{
+				"command": "go.install",
+				"title": "Go: Install packages",
+				"description": "Installs all configured packages for the workspace."
 			}
 		],
 		"debuggers": [
@@ -213,6 +218,14 @@
 					"type": "string",
 					"default": "30s",
 					"description": "Specifies the timeout for go test in ParseDuration format."
+				},
+				"go.installPackages": {
+					"type": "array",
+					"items": {
+						"type": "string"
+					},
+					"default": ["./.."],
+					"description": "Specifies the packages to install when using the install command."
 				}
 			}
 		}
--- a/src/goInstall.ts
+++ b/src/goInstall.ts
@@ -1,0 +1,29 @@
+'use strict';
+
+import cp = require('child_process');
+import path = require('path');
+import vscode = require('vscode');
+import util = require('util');
+import { getGoRuntimePath } from './goPath'
+
+export function goInstall(packages: string[]) {
+	let editor = vscode.window.activeTextEditor;
+	if (!editor) {
+		vscode.window.showInformationMessage("No editor is active.");
+		return;
+	}
+	let channel = vscode.window.createOutputChannel('Go');
+	channel.clear();
+	channel.show(2);
+	let args = ['install', '-v', packages.join(' ')];
+	let proc = cp.spawn(getGoRuntimePath(), args, { env: process.env, cwd: vscode.workspace.rootPath });
+	proc.stdout.on('data', chunk => channel.append(chunk.toString()));
+	proc.stderr.on('data', chunk => channel.append(chunk.toString()));
+	proc.on('close', code => {
+		if (code) {
+			channel.append("Error: Install failed.");
+		} else {
+			channel.append("Success: Install completed.");
+		}
+	});
+}
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -20,6 +20,7 @@ import { setupGoPathAndOfferToInstallTools } from './goInstallTools'
 import { GO_MODE } from './goMode'
 import { showHideStatus } from './goStatus'
 import { testAtCursor, testCurrentPackage, testCurrentFile } from './goTest'
+import { goInstall } from './goInstall'
 
 let diagnosticCollection: vscode.DiagnosticCollection;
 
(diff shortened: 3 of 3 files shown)


ironcladlou:
@lukehoban I'd like to get back into contributing to the plugin, and closing the loop on this PR seems like a decent place to start. Do you have any thoughts on how to unify the `install` and `build` experience in terms of the diagnostics engine? My first pass at this was to basically make `go install` totally disjoint from diagnostics/error reporting: just run it and report the output. If you think that's useful enough for a first draft I'll go ahead and rebase and push this through with some followup issues for further discussion on an improved experience.

lukehoban:
@ironcladlou Sorry for a long delay getting back to you.

I think it does make sense to go ahead with adding a `Go: Install` command to the extension whcih behaves similarly to the existing `Go: Run tests` commands.  

I think it's okay for this to be fully separate from the diagnostic reporting and build-on-save.

The one place they should overlap is in reusing the `go.buildFlags` settings.

The big question is how to scope this.  I can imagine either:

1. Run `go install ./...`  in ${workspaceRoot}
2. Run `go install` in the directory of the currently opened file

For advanced uses, I'm okay with saying that folks should use VS Code Tasks.  So I think this should try to cover the most common 80% for folks who don't want/need to set up the task.

I think ultimately we want:

* `Go: Build`
* `Go: Install`
* `Go: Run`
* `Go: Generate`

Which will all follow a similar model to `Go: Run tests *` (I wonder if we should retitle those as `Go: Test (current file)`, etc. to align with the naming pattern?).

ironcladlou:
>For advanced uses, I'm okay with saying that folks should use VS Code Tasks. So I think this should try to cover the most common 80% for folks who don't want/need to set up the task.

I want to reply more fully to your comment, but I only have a moment and this caught my attention because I was playing with takes recently for this very purpose. The problem I encountered is that the `process.env` set by vscode-go wasn't being inherited by the `go install` process. I don't have the launch JSON at hand right now- just wanted to bring this up in case you know of anything offhand that I could have been doing wrong. I couldn't find any environment fields in the tasks configuration schema.

lukehoban:
@ironcladlou I left a comment here (https://github.com/Microsoft/vscode-go/issues/21#issuecomment-200582995) about the env issue in tasks - it should be solvable.

ironcladlou:
@lukehoban 
>@ironcladlou I left a comment here (#21 (comment)) about the env issue in tasks - it should be solvable.

Thanks- I wrote up a reply and THEN saw the other message... :blush:

ironcladlou:
@lukehoban just a quick followup: I had a chance to play with your suggestion in https://github.com/Microsoft/vscode-go/issues/21#issuecomment-200582995. It works really well. One usability issue with this approach is the user needs to define a problem matcher for every task. For example, here's one I whipped up which seems to work pretty well and ensures that compilation errors during install appear in the vscode problems list:

```json
"tasks": [
    {
      "taskName": "install",
      "args": [
        "-v",
        "github.com/openshift/origin/cmd/openshift"
      ],
      "isBuildCommand": true,
      "problemMatcher": {
        "owner": "go",
        "fileLocation": [
          "relative",
          "${workspaceRoot}"
        ],
        "pattern": {
          "regexp": "^([^:]+):(\\d+):\\ (.*)$",
          "file": 1,
          "line": 2,
          "message": 3
        }
      }
    }
]
```

I notice that vscode ships with several [out of the box matchers](https://code.visualstudio.com/docs/editor/tasks#_processing-task-output-with-problem-matchers). Is there some way vscode-go can contribute a matcher which could then be referenced like `$go-install`? With such a reusable matcher, I think tasks work great for this (and probably many other) purposes.

lukehoban:
> I notice that vscode ships with several out of the box matchers. Is there some way vscode-go can contribute a matcher which could then be referenced like  $go-install ? With such a reusable matcher, I think tasks work great for this (and probably many other) purposes.

@egamma - Is this possible currently?  I recalling looking for a way to do this some time ago and not seeing anything.  If not, is it on the roadmap or something we should open a new tracking issue for on the vscode repo?

egamma:
@lukehoban we currently have no contribution point for the problem matchers. However, if it is a generally useful matcher, then we would take a pull request for a default problem matcher. This is the location:
https://github.com/Microsoft/vscode/blob/master/src/vs/platform/markers/common/problemMatcher.ts#L395

lukehoban:
@ironcladlou Interested in opening a PR for adding this problem matcher to VS Code per @egamma's suggestion?  I believe the error format is close to or the same as GCC/Clang - I wonder if it can be generalized to be useful for those cases as well?  

FWIW - the regex currently being used inside `vscode-go` to extract error messages from all of `go build` `golint` and `go tool vet` is this:

```javascript
let match = /^([^:]*: )?((.:)?[^:]*):(\d+)(:(\d+))?: (.*)$/.exec(lines[i]);
if (!match) continue;
let [_, __, file, ___, lineStr, ____, charStr, msg] = match;
```

ironcladlou:
@lukehoban 
>FWIW - the regex currently being used inside vscode-go to extract error messages from all of go build golint and go tool vet is this:

I just got a chance to test the existing regex for build/lint and it also works fine for install. I haven't had time yet to hack on VScode itself, although that may change soon since I can't wait for an official build to use a certain feature that just got committed. I might try and get to it this weekend.

ironcladlou:
https://github.com/Microsoft/vscode/pull/5115 will allow for reusable `go` toolchain tasks.

ironcladlou:
@lukehoban 
>(I wonder if we should retitle those as Go: Test (current file), etc. to align with the naming pattern?)

Continuing to chip away at things discussed here: https://github.com/Microsoft/vscode-go/pull/286

ironcladlou:
Closing this PR as tasks are sufficient for the use case it meant to solve. Opened https://github.com/Microsoft/vscode-go/issues/287 to track the actual feature that emerged from discussions.

Issue #80, opened 2015-11-24
Add support for coverage markup


newhook:
I've been using gocov and gocov-html (https://github.com/matm/gocov-html) to generate coverage diagrams for my tests. It would be awesome if it was possible to get these integrated into vscode-go.

tecbot:
Did you tried out https://github.com/joefitzgerald/go-plus for atom? I really liked the integrated highlighting of code coverage inside the editor, so you don't need to switch to something else to check your coverage :).
But didn't now if it's possible to change the background color of some code positions in vscode.

newhook:
I think it could use the diagnostics to do this. For example, DiagnosticSeverity.Information, or Hint.

egamma:
@tecbot look into the support for decorators in the VS Code API. Here is an example https://github.com/Microsoft/vscode-extension-samples/tree/master/decorator-sample

newhook:
That's a 404.

lukehoban:
@egamma Look's like the `vscode-extension-samples` is still private - not sure if that is intentional or not?

@tecbot See [`setDecorations`](https://github.com/Microsoft/vscode-extension-vscode/blob/master/vscode.d.ts#L744) for part of the API for working with decorations.  There's also an extension doing some code coverage rendering here: https://github.com/bmeck/vscode-code-cover/blob/master/lib/ui.js

egamma:
This is an oversight, will fix, sorry.

Am 24.11.2015 um 18:50 schrieb Luke Hoban <notifications@github.com<mailto:notifications@github.com>>:


@egamma<https://github.com/egamma> Look's like the vscode-extension-samples is still private - not sure if that is intentional or not?

@tecbot<https://github.com/tecbot> See setDecorations<https://github.com/Microsoft/vscode-extension-vscode/blob/master/vscode.d.ts#L744> for part of the API for working with decorations. There's also an extension doing some code coverage rendering here: https://github.com/bmeck/vscode-code-cover/blob/master/lib/ui.js

egamma:
Fixed the vscode-extension-samples is now public.

tecbot:
Thanks for your hints! Really helpful :)

tampajohn:
@tecbot I've taken a stab at adding coverage highlighting in #143, trying to keep it close to what we're used to in Atom (only highlights current package files for example).  Also added a "get test coverage in this package" command...

lukehoban:
[...]

PR #139, opened 2015-12-01
Add support for coverage markup

tampajohn:
First attempt at addressing #80, need to add timeout for go test.

Broke out goCover.ts from goCheck just for maintainability... file seems to be getting large.


tampajohn:
@lukehoban  a couple of things I'd like to address are:

1. In go-plus in Atom, it saves off the coverage results of the last tested package, so if you open a previously unopened package file, you see the coverage... I like this feature, thoughts?

PR #142, opened 2015-12-03
Issue 57

newhook:
Provisional PR for issue 57. Requires github.com/newhook/go-symbols.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -41,7 +41,8 @@ export function setupGoPathAndOfferToInstallTools() {
 		godef: "github.com/rogpeppe/godef",
 		golint: "github.com/golang/lint/golint",
 		"go-find-references": "github.com/lukehoban/go-find-references",
-		"go-outline": "github.com/lukehoban/go-outline"
+		"go-outline": "github.com/lukehoban/go-outline",
+		"go-symbols": "github.com/newhook/go-symbols"
 	}
 	var keys = Object.keys(tools)
 	Promise.all(keys.map(tool => new Promise<string>((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


newhook:
I'm continuing to work on go-symbols with some help from Alan Donovan. I have an implementation that is more than twice as fast as the one in my current repo, but am still thinking :)

It would be interesting to have some input from the vscode authors on what symbols exactly can be shown. It isn't documented, and code seems to only show func and type.

egamma:
Adding @jrieken he can provide more information.

newhook:
Thanks for that! IMO, code should show the symbols presented to the API. In this way it is up to the extension authors to decide what types of symbols are shown.

jrieken:
@bpasero will know

bpasero:
@newhook isn't the list documented here:

```typescript
	export enum SymbolKind {
		File,
		Module,
		Namespace,
		Package,
		Class,
		Method,
		Property,
		Field,
		Constructor,
		Enum,
		Interface,
		Function,
		Variable,
		Constant,
		String,
		Number,
		Boolean,
		Array
	}
```

If something does not show up nicely in vscode after you use those types it is most likely a bug on our end how these things are converted.

bpasero:
Actually we do some filtering on the handler side and the list we allow is: 'class', 'interface', 'enum', 'function', 'method'

@aeschli can also chime in if this list should be configurable maybe.

newhook:
Pushed change to maxBuffer. Also pushed changes to go-symbols to improve the speed, and fix the null issue if there are no symbols.

newhook:
This PR has been languishing for some time...

egamma:
Ping @lukehoban.

lukehoban:
Sorry for the long delay on merging this.  Just re-reviewed and pulled this in.

PR #161, opened 2015-12-16
Setup Travis CI

lukehoban:


bpasero:
@lukehoban I published vscode@0.10.7 with the ci-test command so you can go back from "next" to "0.10.x". Also, since 0.10.5 is released you do not need to configure a custom download URL!

PR #164, opened 2015-12-17
Fix Add Import inserting "undefined" if user didn't select a package …

### LINKED ISSUE #163: Add Import inserts "undefined" if package wasn't selected
If press `Escape` when the popup with package list is shown, `"undefined"` string would be inserted into the import section.

bredov:
…to import.

Fixes #163


lukehoban:
Thanks!

PR #174, opened 2016-01-09
Allow user refresh currently GOPATH.

andy-zhangtao:
If user needs modify currently GOPATH, user could use this command (go:
refresh gopath) to refresh it rather than restart vscode.

Signed-off-by: Andy Zhang andy.zhangtao@hotmail.com

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "onCommand:go.gopath",
     "onCommand:go.test.cursor",
     "onCommand:go.test.package",
-    "onCommand:go.test.file"
+    "onCommand:go.test.file",
+	"onCommand:go.gopath.refresh"
   ],
   "main": "./out/src/goMain",
   "contributes": {
@@ -101,7 +102,12 @@
         "command": "go.import.add",
         "title": "Go: Add Import",
         "description": "Add an import declaration"
-      }
+      },
+	  {
+		"command": "go.gopath.refresh",
+        "title": "Go: Refresh GOPATH",
+        "description": "Load currently GOPATH"
+	  }
     ],
     "debuggers": [
       {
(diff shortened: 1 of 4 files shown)


lukehoban:
@andy-zhangtao It looks like this PR now includes two features:

1. A command to refresh GOPATH
2. A command to run a Go Build

For the first, could you explain the scenario in which you would use this?  How would the new GOPATH be set and discovered during the refresh?

For the second, the existing way to accomplish this is with VSCode "tasks".  See https://github.com/Microsoft/vscode-go/wiki/Usage-FAQ#how-do-i-use-vs-codes-tasks-with-go.  That said, I think it will make sense to add a `Go: Build` and `Go: Install` commands.  There was some work on these here: https://github.com/Microsoft/vscode-go/pull/134.  I think there are still some design questions that should be answered on this.

I'm going to close this particular PR out for now, but if you'd like to submit as a PR for one or the other of these, feel free to reopen or create anew.

andy-zhangtao:
Hi Luke,

For the first feature, there maybe multiple folder to store packages. 

Assume there are three folder : 

1.$HOME/golang/go/(golang SDK package)
2.$HOME/golang/thirdparty/(third party package)
3.$HOME/golang/myproject/(myself package)

In vscode, user can put the three folders configuration to settings.json. But if user needs a new package, which store in an other folder($HOME/golang/novalidation/).

In current mode, user needs modify the settings.json and restart vscode. 

If vscode has the first feature(refresh GOPATH), user only need modify the settings.json, don’t need restart vscode. Meanwhile, if vscode has this feature, every project could has independent GOPATH setting, without interfering with each other.

User could set up a settings.json in each project.  The content of json file is as same as the global settings.json file. 

When user execute the refresh via command (GO:Refresh GOPATH), vscode will scan this settings.json file and reset the current process’s env variable(env[GOPATH]).



If you have any questions, please let me know. 

Thanks
Andy

Issue #128, opened 2015-11-30
Code completion for method calls can fill out parameters


ahmetalpbalkan:
![image](https://cloud.githubusercontent.com/assets/159209/11481176/a5be36fa-9750-11e5-9004-944f90187b70.png)

hitting `return` here in the code completion context gives me:

    azure.NewServicePrincipalToken

it could give:

    azure.NewServicePrincipalToken(id, secret, tenantID, resource)

GoSublime does it and it is very successful.

lukehoban:
The current way we plan on addressing this is https://github.com/Microsoft/vscode-go/issues/9.  

This is a different UX, but I think satisfies the same need (and arguably with more context.

I'm inclined to focus on getting that added before considering also doing something like the UX proposed here.

ahmetalpbalkan:
Agreed, can be better. If you try  GoSublime, it actually just prints the method signature and then allows you to circle on the parameters with Tab key.

Ideally, I'd expect this completion to complete the left side of the call as well by adding `,err` if missing (and just add an `if err != nil` check to the next line if that's the case).

PR #181, opened 2016-01-27
func param completion

zippoxer:
Func parameter completion or hinting is a bit old but significant issue mentioned at #9, #128 and #135.

This change solves this issue by inserting a complete snippet of the function or method selected from the suggestions.

An example:
Example func snippet

At last argument, tab would set the cursor after the closing bracket.

This takes the same approach as Atom's go-plus and GoSublime.

Personally, I like this approach better than a pop-up that hides my code as I'm filling parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.buildFlags": [],
 	"go.lintFlags": [],
 	"go.vetFlags": [],
+	"go.useCodeSnippetsOnFunctionSuggest": false,
 	"go.formatOnSave": false,
 	"go.formatTool": "goreturns",
 	"go.goroot": "/usr/local/go",
--- a/package.json
+++ b/package.json
@@ -225,6 +225,11 @@
           "default": "goreturns",
           "description": "Pick 'gofmt', 'goimports' or 'goreturns' to run on format."
         },
+		"go.useCodeSnippetsOnFunctionSuggest": {
+			"type": "string",
+			"default": false,
+			"description": "Complete functions with their parameter signature"
+		},
         "go.gopath": {
           "type": "string",
           "default": null,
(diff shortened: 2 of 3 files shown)


lukehoban:
Thanks for doing this!

For consistency with the rest of VS Code, I do still think that it would be good to have Signature Help instead here.  That's the standard experience for Code, and is somewhat less invasive on the default typing experience (though as you say, has the negative that it covers some of the text with the pop-up).

But definitely there are different perspectives on this.  

If you'd like to add a config toggle, and enable this feature under that toggle, with the feature off-by-default for now, I'd be happy to merge this in.  Perhaps call the flag "parameterSnippetCompletion" or something similar

In parallel, I'll take another look at the Signature Help implementation.

zippoxer:
@lukehoban 

Yeah, my first try was fixing the Signature Help from issue #9. When I saw how it's going to look like, I just preferred the snippets.

I'll add the config toggle then ;)

lukehoban:
I've opened a PR on true SignatureHelp support: #184.  It appears to be pretty much complete from my testing so far.

BTW - you could consider naming the configuration option `useCodeSnippetsOnMethodSuggest` or `useCodeSnippetsOnFunctionSuggest`.  The former appears to be the named used by a very similar opt-in feature for TypeScript in VS Code: https://github.com/Microsoft/vscode/issues/297

zippoxer:
@lukehoban 

Great job there ;-)

I'm really excited to see vscode-go moving forward.

I added the same opt-in as TypeScript's, for the sake of consistency.

lukehoban:
Merged this in with a couple of additions in https://github.com/Microsoft/vscode-go/commit/c2ee598a2d681777e9e621e48b81d3a56b8c581c, including a fix for an edge case of parameter lists with nested commas.  

Thanks!

PR #254, opened 2016-03-20
Added support for remote debugging.

giacoman:
I have added support for remote debugging.  I did this by adding remotePath, port, and host properties to configuration.  Also, mode property must be set to debug_remote.  Could not find graceful way to shut down remote dlv process, so I just issue a halt command and the developer must Ctrl-C on the remote dlv process.  Hopefully, you can merge this request in.  Nice to be able to debug in vscode!  Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,9 @@
             "type": "go",
             "request": "launch",
             "mode": "debug",
+			"remotePath": "",
+			"port": 2345,
+			"host": "127.0.0.1",
             "program": "${workspaceRoot}",
             "env": {},
             "args": []
@@ -140,7 +143,7 @@
               },
               "mode": {
                 "type": "string",
-                "description": "One of 'debug', 'test', 'exec'.",
+                "description": "One of 'debug', 'debug_remote', 'test', 'exec'.",
                 "default": "debug"
               },
               "stopOnEntry": {
@@ -175,6 +178,21 @@
                 "type": "string",
                 "description": "Init file, executed by the terminal client.",
                 "default": ""
+              },
+              "remotePath": {
+                "type": "string",
+                "description": "If remote debugging, the path to the source code on the remote machine, if different from the local machine.",
+                "default": ""
+              },
+              "port": {
+                "type": "number",
+                "description": "The port that the delve debugger will be listening on.",
+                "default": 2345
+              },
+              "host": {
+                "type": "string",
+                "description": "The host name of the machine the delve debugger will be listening on.",
+                "default": "127.0.0.1"
               }
             }
           }
--- a/package.json
+++ b/package.json
@@ -124,9 +124,9 @@
             "type": "go",
             "request": "launch",
             "mode": "debug",
-			"remotePath": "",
-			"port": 2345,
-			"host": "127.0.0.1",
+            "remotePath": "",
+            "port": 2345,
+            "host": "127.0.0.1",
             "program": "${workspaceRoot}",
             "env": {},
             "args": []
@@ -143,7 +143,7 @@
               },
               "mode": {
                 "type": "string",
-                "description": "One of 'debug', 'debug_remote', 'test', 'exec'.",
+                "description": "One of 'debug', 'remote', 'test', 'exec'.",
                 "default": "debug"
               },
               "stopOnEntry": {
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -144,69 +144,69 @@ function log(msg?: any, ...args) {
 }
 
 class Delve {
-    program: string;
-    remotePath: string;
+	program: string;
+	remotePath: string;
 	debugProcess: ChildProcess;
 	connection: Promise<RPCConnection>;
 	onstdout: (str: string) => void;
 	onstderr: (str: string) => void;
 
 	constructor(mode: string, remotePath: string, port: number, host:string, program: string, args: string[], cwd: string, env: { [key: string]: string }, buildFlags: string, init: string) {
-        this.program = program;
-        this.remotePath = remotePath;
-        this.connection = new Promise((resolve, reject) => {
-            let serverRunning = false;
-            if (mode === 'debug_remote') {
-                this.debugProcess = null;
-                serverRunning = true;  //assume server is running when in debug_remote mode
-                connectClient(port, host);
-            }
-            else 
-            {
-                let dlv = getBinPath('dlv');
-                log('Using dlv at: ', dlv);
-                if (!existsSync(dlv)) {
-                    return reject('Cannot find Delve debugger. Ensure it is in your `GOPATH/bin` or `PATH`.');
-                }
-                let dlvEnv: Object = null;
-                if (env) {
-                    dlvEnv = {};
-                    for (let k in process.env) {
-                        dlvEnv[k] = process.env[k];
-                    }
-                    for (let k in env) {
-                        dlvEnv[k] = env[k];
-                    }
-                }
-                let dlvArgs = [mode || 'debug'];
-                if (mode === 'exec') {
-                    dlvArgs = dlvArgs.concat([program]);
-                }
-                dlvArgs = dlvArgs.concat(['--headless=true', '--listen=' + host + ':' + port.toString(), '--log']);
-                if (buildFlags) {
-                    dlvArgs = dlvArgs.concat(['--build-flags=' + buildFlags]);
-                }
-                if (init) {
-                    dlvArgs = dlvArgs.concat(['--init=' + init]);
-                }
-                if (args) {
-                    dlvArgs = dlvArgs.concat(['--', ...args]);
-                }
+		this.program = program;
+		this.remotePath = remotePath;
+		this.connection = new Promise((resolve, reject) => {
+			let serverRunning = false;
+			if (mode === 'remote') {
+				this.debugProcess = null;
+				serverRunning = true;  //assume server is running when in remote mode
+				connectClient(port, host);
+			}
+			else
+			{
+				let dlv = getBinPath('dlv');
+				log('Using dlv at: ', dlv);
+				if (!existsSync(dlv)) {
+					return reject('Cannot find Delve debugger. Ensure it is in your `GOPATH/bin` or `PATH`.');
+				}
+				let dlvEnv: Object = null;
+				if (env) {
+					dlvEnv = {};
+					for (let k in process.env) {
+						dlvEnv[k] = process.env[k];
+					}
+					for (let k in env) {
+						dlvEnv[k] = env[k];
+					}
+				}
+				let dlvArgs = [mode || 'debug'];
+				if (mode === 'exec') {
+					dlvArgs = dlvArgs.concat([program]);
+				}
+				dlvArgs = dlvArgs.concat(['--headless=true', '--listen=' + host + ':' + port.toString(), '--log']);
+				if (buildFlags) {
+					dlvArgs = dlvArgs.concat(['--build-flags=' + buildFlags]);
+				}
+				if (init) {
+					dlvArgs = dlvArgs.concat(['--init=' + init]);
+				}
+				if (args) {
+					dlvArgs = dlvArgs.concat(['--', ...args]);
+				}
 
-                let dlvCwd = dirname(program);
-                try {
-                    if (lstatSync(program).isDirectory()) {
-                        dlvCwd = program;
-                    }
-                } catch (e) { }
-                this.debugProcess = spawn(dlv, dlvArgs, {
-                    cwd: dlvCwd,
-                    env: dlvEnv,
-                });
-            }
+				let dlvCwd = dirname(program);
+				try {
+					if (lstatSync(program).isDirectory()) {
+						dlvCwd = program;
+					}
+				} catch (e) { }
+				this.debugProcess = spawn(dlv, dlvArgs, {
+					cwd: dlvCwd,
+					env: dlvEnv,
+				});
+			}
 
-            function connectClient(port: number, host: string) {
-                 let client = Client.$create(port, host);
+			function connectClient(port: number, host: string) {
+				 let client = Client.$create(port, host);
 				client.connectSocket((err, conn) => {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -124,6 +124,9 @@
             "type": "go",
             "request": "launch",
             "mode": "debug",
+			"remotePath": "",
```
lukehoban: Here and elsewhere whitespace appears to be off.  The project uses tabs.  Could you make sure whitespace and formatting match?

package.json
```diff
@@ -140,7 +143,7 @@
               },
               "mode": {
                 "type": "string",
-                "description": "One of 'debug', 'test', 'exec'.",
+                "description": "One of 'debug', 'debug_remote', 'test', 'exec'.",
```
lukehoban: I'd lean toward using just "remote" here.  The previous names in this list were all aligned with arguments to `dlv`, but this one won't be since it won't launch `dlv` itself.  In principle, it could be used to do the remote equivalent of any of the other modes, depending on what was used on the server.

src/debugAdapter/goDebug.ts
```diff
@@ -141,57 +144,69 @@ function log(msg?: any, ...args) {
...
+        this.program = program;
+        this.remotePath = remotePath;
+        this.connection = new Promise((resolve, reject) => {
+            let serverRunning = false;
+            if (mode === 'debug_remote') {
+                this.debugProcess = null;
+                serverRunning = true;  //assume server is running when in debug_remote mode
+                connectClient(port, host);
+            }
+            else
```
lukehoban: To avoid the additional nesting here - can the remote path complete with a return and then keep the local path at top level of the Promise?  The `if (this.debugProcess != null)` further down can then be removed as well I believe.

src/debugAdapter/goDebug.ts
```diff
@@ -246,7 +263,22 @@ class Delve {
...
+                // {
+                //     this.call<number>('Detach', [true], (err, ret) => {
+                //         if (err) {
+                //             console.error('Failed to detach.');
+                //         }
+            	// 	});                                                                                           
+                // }
+    		});                                               
+        }
+		else this.debugProcess.kill();
```
lukehoban: Curly braces around this and similar in line with the rest of the formatting here.

Also - run `npm run lint` to see lint failures generally - they will fail the travis build as well.

src/debugAdapter/goDebug.ts
```diff
@@ -246,7 +263,22 @@ class Delve {
...
 	close() {
-		this.debugProcess.kill();
+        if (this.debugProcess === null) {                                                    
+            this.call<DebuggerState>('Command', [{ name: 'halt' }], (err, state) => {
+                if (err) {
+                    console.error('Failed to halt.');
+                }
+                // else TODO:  Can't find a way to kill remote delve.
+                // {
+                //     this.call<number>('Detach', [true], (err, ret) => {
```
lukehoban: I think it probably doesn't make sense to detach from the debuggee here - since the Code debugger didn't do the attach itself, it assumed the debug server was already running and attached to a process.

src/debugAdapter/goDebug.ts
```diff
@@ -280,7 +312,10 @@ class GoDebugSession extends DebugSession {
 
 	protected launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
 		// Launch the Delve debugger on the program
-		this.delve = new Delve(args.mode, args.program, args.args, args.cwd, args.env, args.buildFlags, args.init);
+		let remotePath = typeof args.remotePath === 'undefined' ? "":args.remotePath;
```
lukehoban: Just for consistency, the codebase typically just uses `args.remotePath || ""` style.

src/debugAdapter/goDebug.ts
```diff
@@ -321,6 +356,18 @@ class GoDebugSession extends DebugSession {
 		this.sendResponse(response);
 		log('ExceptionBreakPointsResponse');
 	}
+    
+    protected toDebuggerPath(path): string {
+        if (this.delve.remotePath.length === 0)
+            return this.convertClientPathToDebugger(path);
+        return path.replace(this.delve.program, this.delve.remotePath);
```
lukehoban: Can you use Node's `path.relative` or similar - to make sure missing trailing '/' and/or Windows/Unix differences are handled cleanly?

src/debugAdapter/goDebug.ts
```diff
@@ -328,6 +375,7 @@ class GoDebugSession extends DebugSession {
 			this.breakpoints.set(args.source.path, []);
 		}
 		let file = args.source.path;
+        let remotePath = this.toDebuggerPath(file)
```
lukehoban: `remoteFile` perhaps?

lukehoban:
Thanks @giacoman - this looks great.  A handful of code review comments.  

Could you rebase on current master?

giacoman:
I will go through all the recommendations and update the code.  Maybe the weekend before I can get to it.

src/debugAdapter/goDebug.ts
giacoman: I removed the commented out code.  I think the other code should remain though.  I shut down remote dlv process by hitting Ctrl-C.  If I don't issue halt command, the remote dlv process will not shut down gracefully.  It will hang trying to delete the debug binary and you will have to hit Ctrl-X and process will be left open.  With halt command, hitting Ctrl-C will shut down process smoothly.

package.json
giacoman: Fixed.

package.json
giacoman: Fixed.

src/debugAdapter/goDebug.ts
giacoman: Fixed.

src/debugAdapter/goDebug.ts
giacoman: Fixed.

src/debugAdapter/goDebug.ts
giacoman: Fixed.

src/debugAdapter/goDebug.ts
giacoman: I updated code so that if remotePath has trailing '/' or '\', it will remove it.  Assumption is that program will not have any trailing slashes.  Not sure if that is what you want.

src/debugAdapter/goDebug.ts
giacoman: Changed to remoteFile.

src/debugAdapter/goDebug.ts
giacoman: Updated to handle case where local and remote paths have different path separators.

lukehoban:
Looks great.  Really good to add this support - opens up a number of new scenarios.  

Only one remaining question:

> I removed the commented out code. I think the other code should remain though. I shut down remote dlv process by hitting Ctrl-C. If I don't issue halt command, the remote dlv process will not shut down gracefully. It will hang trying to delete the debug binary and you will have to hit Ctrl-X and process will be left open. With halt command, hitting Ctrl-C will shut down process smoothly.

At least in the Docker-based setup I've been testing this with (see https://github.com/lukehoban/webapp-go/tree/debugging), what has made the most sense is to send a `restart` command to the remote `dlv` server when you stop debugging in VS Code.  That puts the dlv server back in the state closest to what it was presumably in before attaching the VS Code debugger.  And it allows restarting the debug session in VS Code.  If you send a `halt`, it seems you then have to recycle the debug server on the remote machine every time you want to start a new debug session.

Regarding the `ctrl-c` issue, that sounds like a bug in Delve.  I can't think of a reason why the headless mode `dlv` server should not shutdown gracefully on ctrl-c.

Thoughts?

giacoman:
How did you issue restart command?  Did you test running remote dlv and have it shutdown gracefully after hitting Ctrl-C while doing restart instead of halt?  What I am seeing is that if I don't issue halt command, calling Ctrl-C will hang trying to delete 'Debug'.  I played around with issuing restart command, but may have not been doing it right.

lukehoban:
I tested `restart` just by changing from `halt` to `restart` in the `Delve#close` implementation in your code.

I didn't have any problem with restarting the docker container - but I'm not sure what signal that sends to the process - maybe different than `ctrl-c`.  I didn't have need in the Docker approach to manually kill/restart the debug server.

That said, I think it would be good to raise the issue about being about to `ctrl-c` the debug server process as a Delve bug.  It seems like something that should be fixable there.  

FWIW - Here's the patch I have currently on top of your branch:

```patch
diff --git a/src/debugAdapter/goDebug.ts b/src/debugAdapter/goDebug.ts
index e73e63a..ceea72e 100644
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -234,7 +234,7 @@ class Delve {
 			});
 			this.debugProcess.on('close', function(code) {
 				// TODO: Report `dlv` crash to user.
-				console.error('Process exiting with code: ' + code);
+				logError('Process exiting with code: ' + code);
 			});
 			this.debugProcess.on('error', function(err) {
 				reject(err);
@@ -264,14 +264,15 @@ class Delve {
 	}

 	close() {
-		if (this.debugProcess === null) {
-			this.call<DebuggerState>('Command', [{ name: 'halt' }], (err, state) => {
+		if (!this.debugProcess) {
+			this.call<DebuggerState>('Command', [{ name: 'restart' }], (err, state) => {
 				if (err) {
-					console.error('Failed to halt.');
+					logError('Failed to restart.');
 				}
 			});
+		} else {
+			this.debugProcess.kill();
 		}
-		else this.debugProcess.kill();
 	}
 }

@@ -359,15 +360,17 @@ class GoDebugSession extends DebugSession {
 		log('ConfigurationDoneRequest');
 	}

-	protected toDebuggerPath(path): string {
-		if (this.delve.remotePath.length === 0)
+	protected toDebuggerPath(path: string): string {
+		if (this.delve.remotePath.length === 0) {
 			return this.convertClientPathToDebugger(path);
+		}
 		return path.replace(this.delve.program, this.delve.remotePath).split(this.localPathSeparator).join(this.remotePathSeparator);
 	}

-	protected toLocalPath(path): string {
-		if (this.delve.remotePath.length === 0)
+	protected toLocalPath(path: string): string {
+		if (this.delve.remotePath.length === 0) {
 			return this.convertDebuggerPathToClient(path);
+		}
 		return path.replace(this.delve.remotePath, this.delve.program).split(this.remotePathSeparator).join(this.localPathSeparator);
 	}

@@ -385,9 +388,11 @@ class GoDebugSession extends DebugSession {
 		})).then(() => {
 			log('All cleared');
 			return Promise.all(args.lines.map(line => {
-				if (this.delve.remotePath.length === 0)
+				if (this.delve.remotePath.length === 0) {
 					log('Creating on: ' + file + ':' + line);
-				else log('Creating on: ' + file + ' (' + remoteFile + ') :' + line);
+				} else {
+					log('Creating on: ' + file + ' (' + remoteFile + ') :' + line);
+				}
 				return this.delve.callPromise<DebugBreakpoint>('CreateBreakpoint', [{ file: remoteFile, line }]).catch(err => {
 					log('Error on CreateBreakpoint');
 					return null;
@@ -450,7 +455,6 @@ class GoDebugSession extends DebugSession {
 					new Source(
 						basename(location.file),
 						this.toLocalPath(location.file)
-						// this.convertDebuggerPathToClient(location.file)
 					),
 					location.line,
 					0
```

giacoman:
Hmmm.  Looking at the delve code, I don't see a 'Restart' option for 'Command'.  Is it possible you are sending non-existent command and nothing was executed?  If so, I would suggest still going with the halt option.  For me, when I am done debugging, I would want the process to stop on the remote server so I could rebuild and run normally.

I was calling Restart differently.  Something like:
this.call<number>('Restart', [null, null], (err, ret) => {
    if (err) {
	console.error('Failed to restart.');
    }
});

If I pass command that is not there, Delve will complain.  It didn't complain for me, but it also didn't seem to restart or work for me.  

I also tried what you had done and it didn't work either.

lukehoban:
You are right - I was not correctly restarting in my tests earlier.

I just tried a few things, and I am now seeing the behaviour I expect if I use:

```typescript
this.call<DebuggerState>('Command', [{ name: 'halt' }], (err, state) => {
    if (err) return logError('Failed to restart.');
    this.call<DebuggerState>('Restart', [], (err, state) => {
        if (err) return logError('Failed to restart.');
    });
});
```

Does that work in your usecase as well?  I believe it should leave the debug server in a state where you can still `ctrl-c`, but will ensure also ensure that `Play`/`Stop`/`Play`/`Stop`/etc. in VS Code will work.

giacoman:
I tried the code where you issue halt and follow it with restart.  I didn't see anything different to just issuing halt.  I am not able to restart debugger without relaunching remote delve process.  One thing I want to make sure on.  In the debugger controls in vscode, there is restart and stop button.  I have been hitting the stop button.  I did test using restart button and it does not successfully restart.

lukehoban:
I just tried again, and without the call to Restart, I'm not able to reliably restart the vscode debugger.  With  the call to Restart in there, I can restart the vscode debugger reliably.

By "restart the vscode debugger", I mean any of:

* While the process is running, hit Stop, then hit Play again.
* While the process is paused in the debugger, hit Stop, then hit Play again.
* While the process is running, hit Restart.
* While the process is paused in the debugger, hit Restart.

This did require sending a `halt` followed by a `Restart` to Delve.  

My particular test case at least is https://github.com/lukehoban/webapp-go/tree/debugging, with a breakpoint in `loadPage`, and a container built and launched per https://github.com/lukehoban/webapp-go/blob/debugging/.vscode/launch.json#L15.

giacoman:
Wonder if we could be using different version of Delve.  On my remote vm, I am using:
Delve Debugger
Version: 0.11.0-alpha

Calling restart doesn't affect me, so I am good with leaving it in.

lukehoban:
Great - this is merged into master now.  I'll push an update to the extension in the next day or so with these changes.

Thanks @giacoman for the work on this!

Issue #255, opened 2016-03-20
HoverProvider raises an exception any time godef fails to return a match at the cursor


ironcladlou:
This exception is being thrown by the HoverProvider every time my mouse hovers over the editor and godef doesn't return a match.

```
Cannot read property 'lines' of null: TypeError: Cannot read property 'lines' of null
    at /Users/dan/.vscode/extensions/vscode-go/out/src/goExtraInfo.js:13:39
    at process._tickCallback (node.js:369:9)
```

This results in a lot of error noise in the console.

ironcladlou:
A related exception occurs when hovering with the `⌘` key:

```
Cannot read property 'file' of null: TypeError: Cannot read property 'file' of null
    at /Users/dan/.vscode/extensions/vscode-go/out/src/goDeclaration.js:76:68
    at process._tickCallback (node.js:369:9)
```

PR #256, opened 2016-03-20
Make HoverProvider tolerate missing locations

ironcladlou:
GoHoverProvider was throwing a TypeError trying to read a property
from null definition locations. It may be more appropriate to try
and prevent the hover provider from entering its handler function if
definitionLocation finds no match, but rejecting the Promise didn't
see to do the trick. This patch makes the HoverProvider call a no-op
when no definition is found.

Fixes https://github.com/Microsoft/vscode-go/issues/255


ironcladlou:
The Travis failure is a mystery...

ironcladlou:
I bundled another commit into this PR to address a related exception in `GoDefinitionProvider` which manifests both when using "Go to definition" directly and also via the HoverProvider (when hover occurs with the `⌘` key pressed).

lukehoban:
Great - thanks @ironcladlou.

PR #258, opened 2016-03-20
Migrate mis-use of setExceptionBreakPointsRequest to configurationDon…

weinand:
Since VS Code 0.10.11 (February) VS Code no longer calls `setExceptionBreakPointsRequest` if no breakpoints filters are returned from the `initializeRequest`. Since you are using `setExceptionBreakPointsRequest` to set the initial breakpoints, 'go-debug' no longer works with the latest version (alpha) of VS Code (and it will not work the GA version of VS Code).
Since January we are warning about this and suggested to migrate this special use of `setExceptionBreakPointsRequest` to the `configurationDoneRequest` and return a corresponding flag from the `initializeRequest`. This PR fixes this issue.
As a nice side effect, you will no longer see the 'All Exception' and 'Uncaught Exceptions' checkboxes that were no-ops for go-debug anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   "devDependencies": {
     "fs-extra": "^0.26.0",
     "tslint": "^3.3.0",
-    "typescript": "^1.6.2",
+    "typescript": "1.8.7",
     "vscode": "^0.10.7"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


lukehoban:
Thanks @weinand.

> As a nice side effect, you will no longer see the 'All Exception' and 'Uncaught Exceptions' checkboxes that were no-ops for go-debug anyway.

FWIW - I am still seeing these two checkboxes on VS Code 0.10.11 after applying your change.  Is that expected?

weinand:
@lukehoban the checkboxes are persisted so they will stay for old workspaces. For newly created workspaces they should be gone.

@isidorn any insights under what circumstances the checkboxes will disappear?

isidorn:
@weinand for old workspaces the checkboxes should disapear as soon as a debug session starts and an adapter sends the new configuration - no exception breakpoints.
For new workspaces they should just not be there any more

PR #270, opened 2016-03-27
Add .editorconfig and add final newline

sijad:
final newline was missing from some of files, also `.editorconfig` added too
`.editorconfig` copied from [vscode](https://github.com/Microsoft/vscode/blob/master/.editorconfig) main repo


sijad:
I was look at vscode files and non of them has final new line, but some of source files of this repo has and some, should all of them removed? let me know please

Issue #252, opened 2016-03-18
vs.code Go forkbombs Mac with very large numbers of Go compilations


smarterclayton:
On El Capitan with version 0.10.11 and the Go tools installed, "build on save" has recently started attempting to build each individual package in my (very large) project in parallel, resulting in hundreds of Go compilations running at any one time and eventually wedging the machine due to memory / CPU contention.  If I `pgrep 6g | kill` the processes I can recover, but it happens every time something kicks off a change to the go file.  The number of processes is much more than a go build would use normally, so it looks like it's trying to build every package in my go path at once.

I don't recall seeing this until the most recent vscode update, however we have made other changes to our build structure and it's possible this is something we introduced.

Scenario:

1. Golang 1.4.2 from home-brew
2. Using github.com/openshift/origin (very large package and dependency tree)
3. Base directory `~/projects/origin/src/github.com/openshift/origin`, 
4. GOPATH set via workspace settings to `~/projects/origin/src/github.com/openshift/origin/Godeps/_workspace:~/projects/origin`
5. "Build on save" turned on

Is there a way to get traces on which code is starting the compilation for the packages to get addition debug info?

ironcladlou:
Interesting, I haven't hit this. Although, I'm now using 1.5.3 (I've found it's actually fine for origin now that compiler cache works right for `install`).

Debugging the plugin easy. [These instructions](https://github.com/Microsoft/vscode-go#building-and-debugging-the-extension) are good. It's fun to have an interactive debugger again!

ironcladlou:
Build on save/linting is [implemented here](https://github.com/Microsoft/vscode-go/blob/master/src/goCheck.ts).

lukehoban:
There's three things this could be:

1. Build on save
2. Completion lists
3. Hover


I would be a bit surprised  if (1) was the problem, since it just does a pretty vanilla `go build` and only once per save.

There was a change not too long ago to set `autobuild` to true on `gocode`, which means that (2) will try to build dependencies that aren't built yet when you ask for completion lists.  It's possible that something is going wrong that is causing that to not be able to reuse the build cache.

As mentioned in https://github.com/Microsoft/vscode-go/issues/257#issuecomment-199512489, (3) has changed recently to get more CPU intensive and to make a call to `godoc`.  That could in principle be the issue.

Anyway you can isolate which of these three features is causing the slowdown and the process count explosion in your case?

ironcladlou:
[...]

PR #283, opened 2016-04-09
Introduce gocode autobuild setting

ironcladlou:
Introduce a setting to allow disabling gocode's experimental autobuild
support. Enabling it can cause performance issues in large codebases,
and users should have the capability to assume full control over when
go build and install is invoked.


ironcladlou:
Probably fixes https://github.com/Microsoft/vscode-go/issues/252 for most people.

lukehoban:
Thanks - great to make this something that is configurable.

PR #285, opened 2016-04-09
Replace oracle with guru

### LINKED ISSUE #284: Switch from oracle to guru
Alan Donovan has [renamed oracle to guru](https://groups.google.com/forum/#!msg/golang-dev/trpStscxjKQ/ujuxvypIAgAJ) where (incompatible) progress is finally being made on the tool. Progress is being tracked [here](https://docs.google.com/document/d/1UErU12vR7jTedYvKHVNRzGPmXqdMASZ6PfE7B-p6sIg/edit#heading=h.t5rbcuy7di7j) and the new tool page is [here](https://godoc.org/golang.org/x/tools/cmd/guru).

I think we should get ahead of this by immediately switching to use guru and abandoning oracle since:

1. It's very slow for large projects in common use cases
2. It's now deprecated

@lukehoban, I noticed you just [switched to oracle](https://github.com/Microsoft/vscode-go/commit/509d6abbe360319596d44eeebfb7917832d89dbb) for the find references implementation, so this is probably very fresh on your mind. Would you be willing to change to `guru`? I haven't yet had a chance to investigate any changes in the CLI or API which might require work in vscode beyond executing the new binary.

ironcladlou:
The oracle tool has been forked and replaced with the guru tool.
Switch vscode-go to use guru instead of oracle to take advantage
of the revived progress on the tool.

Fixes #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This extension adds rich language support for the Go language to VS Code, includ
 - Snippets
 - Quick Info (using `godef`)
 - Goto Definition (using `godef`)
-- Find References (using `oracle`)
+- Find References (using `guru`)
 - File outline (using `go-outline`)
 - Workspace symbol search (using `go-symbols`)
 - Rename (using `gorename`)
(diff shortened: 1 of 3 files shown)


tamird:
`guru defintion` is also a thing now - it should be used instead of `godef.`

freeformz:
guru was oracle: https://github.com/golang/tools/commit/734737930440fc305a816e577cab457fbbc807c1

ottob:
In my fairly large project guru definition lookups are much slower than godef:
godef time: 0.03 seconds
guru time: 2.13 seconds

So I would prefer if the godef option can be kept.

tamird:
Sounds worthy of an upstream bug report. For instance, vim-go has already
replaced godef with guru.

ottob:
Synchronicity! With this patch https://go-review.googlesource.com/#/c/22526/ problem seems to be solved. Consider my objection withdrawn.

godef time: 0.0670468807220459 seconds
guru time: 0.09794092178344727 seconds

Issue #168, opened 2015-12-22
Method Suggestion


klaidliadon:
When you write func in a package the suggestion list show also a method constructor for each type in the package that produces the following output:

``` go
func (p *PkgType) Method(){

}
```

Pressing tab should focus in order:
1. the `*` to remove the pointer sign
2. the `Method` to change the name
3. In the parenthesis to change the argument
4. After the parenthesis to change the return type
5. In the method body

lukehoban:
I like the idea of this.  I noticed that GoSublime has a similar feature.  

It should be possible to add this on top of the same "smart snippets" infrastructure that was added in f8508b40f80a3ce812201ba5eadc268225684b61.

ramya-rao-a:
This feature is now out in the latest update of the Go extension (0.6.70)

PR #306, opened 2016-04-20
add typed function declaration snippet

azr:
Hello there !

vscode-go made me switch to vscode from sublime text ! So nice one and thanks !

And I event recommend it ! Could you imagine ? a microsoft product !

Anyways, I'm proposing this snippet because I used it a lot in sublime text.

In fact the GoSublime snippet is a little bit better because it will propose a list with $1 prepopulated with known local types, which is awesome.

cheers !

### DIFF
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -33,7 +33,7 @@
 			"body": "func $1($2) $3 {\n\t$0\n}"
 		},
 		"typed function declaration": {
-			"prefix": "ft",
+			"prefix": "funct",
 			"body": "func ($1) $2($3) $4 {\n\t$0\n}"
 		},
 		"variable declaration": {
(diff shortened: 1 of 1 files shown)


lukehoban:
Thanks for this.

The feature you really want here is being tracked in #168.  I'd love to get that added.

I'm not as sure about the addition in this PR:

1) I'm not sure `funct` is the right name.  What was your reasoning for the name?  I suspect discoverability would be low and thus this wouldn't get used much.
2) More importantly, feels like this shouldn't need a name.  If #168 was implemented, the completion list could just include generated snippets.

azr:
1/ to create a method you'd type `func` so the autocompletion would trigger.

2/ if someone's on #168 I think we could just close this pr :)

Issue #325, opened 2016-05-12
3rd party tools aware GOPATH


saml:
Although latest go standard tools are aware of vendor/ directory, some 3rd party tools aren't aware of vendor/. Possibly add vendor/ (and other directories that have src/) to GOPATH before executing 3rd party tools.

One example 3rd party tool that is not aware of vendor/: [godef](https://github.com/rogpeppe/godef/issues/13).

One solution for such 3rd party tools: [vim-go](https://github.com/fatih/vim-go/pull/429).

lukehoban:
Unfortunately, I don't think adding `vendor/` to GOPATH is a generally safe solution to this, and I expect it would start introducing subtle issues for some user's environments.

The "right" thing here is for 3rd party Go tools to add support for vendoring.  We won't be able to get reliable results without that.  For `godef` in particular, sounds like this is just waiting on @rogpeppe to accept the pending PR - or, if that will not happen, then for someone to fork and start maintaining an up to date version of this tool which `vscode-go` could use.

Note that `vscode-go` does support mult-gopath environments already, so users should be able to individually use this approach of augmenting their GOPATH to include `vendor/` if they want to workaround the limitation in `godef` in the near term.

saml:
I agree that altering GOPATH in an editor's extension is not a generally safe solution.

When I put dependencies in vendor/ (go15vendorexperiment), vscode-go cannot find dependencies. That is because godef isn't aware of go15vendorexperiment like you said. 

When I put dependencies in vendor/src (multi-gopath), it's tricky to get vscode-go to find dependencies because vscode still executes login shell, erasing my project's multi-gopath setup https://github.com/Microsoft/vscode/issues/4779 .
I am aware that I can set go.gopath to `${workspaceRoot}/vendor:/home/saml/go` in .vscode/settings.json  to override how vscode deals with environment variables. But I cannot share my .vscode/settings.json with other developers cause they might not have /home/saml/go as their GOPATH.

As things currently stand, I can either:

1. Wait for godef to support go15vendorexperiment and put my project's dependencies in vendor/.
2. Or, wait for vscode to respect GOPATH environment variable that I set in my session and put my dependencies to vendor/src (multi-gopath).
3. Or, wait for vscode-go to provide a way to override GOPATH environment variable in .vscode/settings.json in a way that can be shared with other developers (maybe `${workspaceRoot}/vendor:${env.GOPATH}`).
[...]

PR #326, opened 2016-05-12
Adds vendor/ to GOPATH envvar at startup.

saml:
Fixes Microsoft/vscode-go#325


lukehoban:
Per discussion in #325, I think this is not the best path to addressing this right now.

PR #348, opened 2016-06-03
Replace godef with guru

ironcladlou:
Replace the `godef` command with `guru`. The `guru` tool offers
improved definition lookups which are compatible with Go 1.5+
vendoring. Performance seems to be comparable.

This eliminates `godef` as a vscode-go dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ install:
   - npm install
   - npm run vscode:prepublish
   - go get -u -v github.com/nsf/gocode
-  - go get -u -v github.com/rogpeppe/godef
   - go get -u -v github.com/golang/lint/golint
   - go get -u -v github.com/lukehoban/go-outline
   - go get -u -v sourcegraph.com/sqs/goreturns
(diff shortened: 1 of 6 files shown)


ironcladlou:
@lukehoban I was in a hurry to get this working. If you like the idea, please give the factoring some review, I'm sure it could be improved.

lukehoban:
I love the idea of centralizing more of these features onto `guru`.  However, currently, the results seem to be meaningfully less helpful.

For example, hovering a variable shows

ironcladlou:
I've also noticed that kind of stuff over the past few days but haven't investigated further. In the short term I prefer better go-to-definition support and so don't mind the hover regressions, but I agree we shouldn't switch without a solution. I'll see if I can come up with something.

hickeng:
I honestly think I'd chose to have `Find All References` working again with Go 1.6 over the richer types, but maybe that's only because I've not experienced it. Do you know if there's a workaround to support that function in the interim?

jboelter:
I"m encountering a bug with godef that also led me down this path.  If guru is not as information rich vs godef we can ask Alan for some enhancements.  Guru has some other features that look interesting also (whicherrors, implements, peers)

https://github.com/rogpeppe/godef/issues/19 with godef.

ironcladlou:
I'd like to close this PR. `godef` now supports vendoring and is *significantly* faster.

PR #351, opened 2016-06-06
Allow Go tools to be installed in a separate GOPATH

samherrmann:
This is a prototype for issue #5.

The approach used in this prototype allows users to install the Go tools in a separate GOPATH by setting a VSCODE_GOTOOLS env variable to the directory of where they want to install the tools. 
1. What do you think of the overall approach?
2. Is VSCODE_GOTOOLS a good name for the env variable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,9 @@ The extension uses the following tools, installed in the current GOPATH.  If any
 - guru: `go get -u -v golang.org/x/tools/cmd/guru`
 - gotests: `go get -u -v github.com/cweill/gotests/...`
 
-To install them just paste and run:
+If you wish to have the extension use a separate GOPATH for its tools, set the VSCODE_GOTOOLS environment variable to the desired path.
+
+To install the tools manually in the current GOPATH, just paste and run:
 ```bash
 go get -u -v github.com/nsf/gocode
 go get -u -v github.com/rogpeppe/godef
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -126,6 +126,14 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 			HTTPS_PROXY: httpProxy,
 		});
 	}
+
+	// If the VSCODE_GOTOOLS environment variable is set, use 
+ 	// its value as the GOPATH for the "go get" child precess.
+	let toolsGoPath = process.env['VSCODE_GOTOOLS'];
+	if (toolsGoPath) {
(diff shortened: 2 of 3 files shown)


samherrmann:
hmm right after I created the pull request I went back to the issue thread and realized that I should have probably used the bin lookup instead of appending the tools path to the GOPATH.

lukehoban:
This is great - I'm actually pleasantly surprised that it's this simple to add support for a separate Go tools installation path.

I think eventually it would be good to expand upon this and choose a default location for the tools and automatically download them for the user on startup, as well as tracking/managing versions, to avoid all of the current manual acquisition steps.

But this feels like a reasonable step toward that which enables users to easily setup their own global installation path, and allows us to go further in the future, but also to offer configurability of the installation location in the future even if the default changes.

On the env var name - I'm happy with `VSCODE_GOTOOLS`.

/cc @ironcladlou in case he has any thoughts on this.

ramya-rao-a:
@samherrmann There has been quite a few changes to `goPath.ts` and `goInstallTools.ts` files since this PR. Can you resend this PR, after getting the latest from master?

samherrmann:
Yup, I will try to do this tonight.

ramya-rao-a:
Thanks @samherrmann !

ahmetalpbalkan:
@samherrmann @ramya-rao-a  Is there a reason why this is implemented as an environment variable and not a configuration key in user's vscode profile? The way I see it, people run vscode as `code .` and you can't expect them to have this ENV key set at all times. (I certainly do not wish to be in that business.)
 
I think a config key would be a "less ephemeral" solution and would preserve between launches. For instance, I manually install all go tools to my `~/gotools` folder, I wish I could just set it to a vscode preference key and _forget about it forever_ and vscode would handle things for me (such as perhaps periodically updating these tools).

ironcladlou:
Totally agree. Great idea.

samherrmann:
Yes I think I agree as well. I think I used an env var because it seemed to make sense when looking at the existing code, but from a user's point of view a config key makes probably more sense.

Here is my first attempt at implementing the change:
https://github.com/samherrmann/vscode-go/commit/18829b582766f13fd49599fdf3277e591974e8fe

Please provide feedback if anything already jumps out at you.

ramya-rao-a:
One catch with setting this as a user/workspace setting is that the debug adapter will not be able to read it, and so this will work for all tools except `dlv`

PR #352, opened 2016-06-06
Disable the delve --log option by default.

bawr:
This is a small patch that makes it possible to toggle delve's debugging output.

I've found it to introduce a lot of pointless noise, distracting from *meaningful* stderr output.


lukehoban:
Looks good - thanks!

PR #362, opened 2016-06-12
append cwd to gopath when debug program

muidea:
adjust current workspace path for debug path
set current workspace path to go path, like eclipse.
add cwd to debugger initialConfigurations

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -207,10 +207,6 @@ class Delve {
 					dlvCwd = program;
 				}
 			} catch (e) { }
-			log('cwd:', cwd);
-			log('program:', program);
-			log('dlvArgs:', dlvArgs);
-			log('dlvCwd:', dlvCwd);
 			
 			this.debugProcess = spawn(dlv, dlvArgs, {
 				cwd: cwd,
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -9,7 +9,7 @@ import { readFileSync, existsSync, lstatSync } from 'fs';
 import { basename, dirname } from 'path';
 import { spawn, ChildProcess } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
-import { getBinPath } from '../goPath';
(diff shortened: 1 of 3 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -183,7 +187,7 @@ class Delve {
 				}
 			}
 			let dlvArgs = [mode || 'debug'];
-			if (mode === 'exec') {
+			if (mode === 'exec' || mode === 'debug') {
```
lukehoban: Why is the program name being passed to `dlv debug`?  I don't believe it takes a program name as an argument.

src/debugAdapter/goDebug.ts
```diff
@@ -196,15 +200,16 @@ class Delve {
...
+						
+			let dlvCwd = dirname(program);			
 			try {
 				if (lstatSync(program).isDirectory()) {
 					dlvCwd = program;
 				}
 			} catch (e) { }
+			
 			this.debugProcess = spawn(dlv, dlvArgs, {
-				cwd: dlvCwd,
```
lukehoban: Is `dlvCwd` not being used anymore?

lukehoban:
It's not clear to me that it's safe to assume that all users want their GOPATH to be modified in this way.

What is the specific issue that this is fixing?

epsniff:
what's the problem this is trying to solve?   If it's running tests in subpackages, that can be solve with the `${file}` template directive in your debug config. 
```
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Testcase Debug",
			"type": "go",
			"request": "launch",
			"mode": "test",
			"program": "${file}",
			"env": {"TMPDIR": "${env.HOME}/tmp"},
			"args": [""]
		}
	]
}
```

mattetti:
I agree with @epsniff, the issue OP is trying to address can be fixed in a way that wouldn't surprised programmers who don't expect their go path to be modified.

ramya-rao-a:
@muidea Can you respond to the questions posed by @lukehoban @epsniff and @mattetti ?

ramya-rao-a:
Closing this PR as we haven't heard back any answers from the OP to the questions posed.

Issue #275, opened 2016-04-01
Go specific contextual menu


mattetti:
Does VSC extension API support customization of the contextual menu. There are a few things I'd love ot see for instance:

* right click on a function/method call that isn't implemented and generate a snippet named after the selected symbol.
* right click on a go file and choose create a test file which would generate a new test file named after the selected file.
* extract/refactor feature - select a chunk of code, right click > extract as function, pick a name, the selected code is now used as the body of a new function which is called from the location where the code used to be.

Is that kind of stuff doable, part of the scope and something you're interested in seeing, or is that something that should live as its own extension?

lukehoban:
Not currently - but see https://github.com/Microsoft/vscode/issues/4762.

As for the `Go` features mentioned here - all make sense to me.  All of these could probably be added as commands that can be invoked through the command palette already - and then added into a context menu in the future if/when that extension point is opened up.  

Happy to accept PRs for features like these.

vastbinderj:
My only comment on testing, is that in my experience it is better to have tests in a separate package from the package being tested.  In this manner you are testing the package in the same way it will be used.

I really like the other suggestions and in addition would like `go build`, `go run` and `go test` shortcuts which I did create after fashion by wading other issues in this repo.

mattetti:
@lukehoban is that feature now in vscode? It would be really great if we could add a context menu to show interfaces a value/type is implementing. That can be done using guru implements query (we already use guru): https://docs.google.com/document/d/1SLk36YRjjMgKqe490mSRzOPYEDe0Y_WQNRv-EiFYUyw/view#heading=h.15us5y9rxrbh

egamma:
@mattetti VS Code now supports that an extension can contribute to the menus. Please see https://code.visualstudio.com/Updates#_extension-authoring.

ramya-rao-a:
@mattetti Closing the loop on this. As you know contextual menu is now possible in VS Code.

> right click on a function/method call that isn't implemented and generate a snippet named after the selected symbol.

#277 is tracking this request

> right click on a go file and choose create a test file which would generate a new test file named after the selected file.

We now have commands to generate unit tests 


![image](https://cloud.githubusercontent.com/assets/16890566/19872566/ae097bdc-9f77-11e6-81fa-90bc7055cb3b.png)


> extract/refactor feature - select a chunk of code, right click > extract as function, pick a name, the selected code is now used as the body of a new function which is called from the location where the code used to be.

#588 will track this request

PR #370, opened 2016-06-30
Add command to extract method out of selected code

ramya-rao-a:
Related to a feature ask in #275 
This commit adds a command to the command palette that can be used to extract a method out of selected code and replace the selected code with a call to the extracted method using godoctor. Read more about godoctor at http://gorefactor.org/

### DIFF
--- a/src/goExtractMethod.ts
+++ b/src/goExtractMethod.ts
@@ -68,13 +68,6 @@ export function extractMethodUsingGoDoctor(methodName: string, selection: Select
 	let position = `${selection.start.line + 1},${selection.start.character + 1}:${selection.end.line + 1},${selection.end.character + 1}`;
 	
 	return new Promise((resolve, reject) => {
-
-		methodName = methodName.trim();
-		// TODO: Add other checks for methodName
-		if (methodName.indexOf(' ') > -1){
-			return resolve('Invalid name for method');		
-		}
-
 		var process = cp.execFile(godoctor, ['-pos', position, 'extract', methodName], {},(err, stdout, stderr) => {
 			if (err){
 				var errorMessageIndex = stderr.indexOf('Error:');				
@@ -93,7 +86,7 @@ export function extractMethodUsingGoDoctor(methodName: string, selection: Select
 			}
 					
 			applypatches(patches, editor).then(validEdit => {
-				return validEdit? resolve(null): resolve('Edits could not be applied to the document');				
+				return resolve (validEdit? null: 'Edits could not be applied to the document');				
 			});;
 				
 		});
--- a/src/goExtractMethod.ts
+++ b/src/goExtractMethod.ts
@@ -28,30 +28,11 @@ export function extractMethod() {
 		
 	let showInputBoxPromise = window.showInputBox({placeHolder: 'Please enter the name for the extracted method'});
 	showInputBoxPromise.then((methodName: string) => {		
-
-		// Ensure there is always an eol char at the end of the document	
-		// Else "No End of Line" char gets added by godoctor which will fail diff_match_patch.patch_fromtext()
-		let editPromise: Thenable<boolean>;
-		if(!editor.document.getText().endsWith("\n")){			
-			editPromise = editor.edit((editBuilder) => {
-				editBuilder.insert(new Position(editor.document.lineCount, 0), "\n");
-			});
-		} else {
-			editPromise = Promise.resolve(true);
-		}
-	
-		editPromise.then((validEdit) => {
-			if (!validEdit) {
-				window.showErrorMessage('Edits could not be applied to the document');
-				return;
+		extractMethodUsingGoDoctor(methodName, editor.selection, editor).then(errorMessage => {
+			if (errorMessage){
+				window.showErrorMessage(errorMessage);
 			}
-			extractMethodUsingGoDoctor(methodName, editor.selection, editor).then(errorMessage => {
-				if (errorMessage){
-					window.showErrorMessage(errorMessage);
-				}
-			});;
-		});
-				
+		});				
 	});		
 }
 
@@ -121,7 +102,18 @@ function applypatches(patches: dmp.Patch[], editor: TextEditor): Thenable<boolea
 			}	
 		})			
 	})
-	return editPromise;	
+
+	// While inserting the extracted method, the end of line char (\n) at the end of the document might get replaced.
+	// If this happens, the next time godoctor is called, it adds "No End of Line" char which will fail diff_match_patch.patch_fromtext()
+	// Therefore, add the end of line char at the end of the document if it does not exists.
+	return editPromise.then(done => {
+		if (done && !editor.document.getText().endsWith("\n")){
+			return editor.edit((editBuilder) => {
+				editBuilder.insert(new Position(editor.document.lineCount, 0), "\n");
+			});
+		}
+		return Promise.resolve(done);
+	});	
 }
 
 /**
--- a/src/goExtractMethod.ts
+++ b/src/goExtractMethod.ts
@@ -15,138 +15,138 @@ import dmp = require('diff-match-patch');
  * Extracts method out of current selection and replaces the current selection with a call to the extracted method.
  */
 export function extractMethod() {
-	
+
 	let editor = window.activeTextEditor;
 	if (!editor) {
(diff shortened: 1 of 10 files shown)


mattetti:
This is a cool feature/tool @ramya-rao-a!  Would you mind checking why the build is failing? It might also be nice to add this feature to the context menu now that it's available to extensions.

mattetti:
Looks like something is wrong with the cert used by `9fans.net` used by github.com/rogpeppe/godef:

```
import "9fans.net/go/acme": https fetch: Get https://9fans.net/go/acme?go-get=1: x509: certificate has expired or is not yet valid
```

ramya-rao-a:
Thanks @mattetti . Once I get this PR merged, I'll get to the context menu.
Initially build failed as go get godoctor failed with Go v.1.5
setting GO15VENDOREXPERIMENT=1 fixes that.   
Not able to repro the cert error though.. Will try with a clean env

src/goExtractMethod.ts
```diff
@@ -0,0 +1,172 @@
...
+		});
+	});
+
+	// While inserting the extracted method, the end of line char (\n) at the end of the document might get replaced.
+	// If this happens, the next time godoctor is called, it adds "No End of Line" char which will fail diff_match_patch.patch_fromtext()
+	// Therefore, add the end of line char at the end of the document if it does not exists.
+	return editPromise.then(done => {
+		if (done && !editor.document.getText().endsWith('\n')) {
+			return editor.edit((editBuilder) => {
+				editBuilder.insert(new Position(editor.document.lineCount, 0), '\n');
```
Tyriar: Can you add the `\n` if necessary on line 74? Also this should probably insert the same line endings as are used in the document (`\r\n` or `\n`).

```ts
process.stdin.end(editor.document.getText() + '\n');
```

src/goExtractMethod.ts
```diff
@@ -0,0 +1,172 @@
...
+				edit.end = new Position(line, patch.diffs[i][1].length);
+				line++;
+				break;
+			case dmp.DIFF_INSERT:
+				if (edit == null) {
+					edit = new Edit(EditTypes.EDIT_INSERT, new Position(line, 0));
+				} else if (edit.action === EditTypes.EDIT_DELETE) {
+					edit.action = EditTypes.EDIT_REPLACE;
+				} else {
+					edit.text += '\n';
```
Tyriar: Again will the editor handle the line ending conversions when required?

src/util.ts
```diff
@@ -84,6 +84,8 @@ export function parameters(signature: string): string[] {
 }
 
 export function canonicalizeGOPATHPrefix(filename: string): string {
+
```
Tyriar: Remove these extra lines

src/goExtractMethod.ts
ramya-rao-a: @Tyriar Extracted methods get added at the end of the document. If `\n` gets added to the input, then the resulting diffs/patches will assume this `\n` exists and not output one. This will result in the extracted method getting appended to the last line instead of starting from a new line _after_ the last line.

src/util.ts
ramya-rao-a: Done

mattetti:
I'm doing a remote meetup during GopherCon this week and will demo VS Code, I'd love to show off this feature :)

egamma:
I've looked at the code and it LGTM.

@mattetti we are having some struggle with the publishing of this extension. For the meetup can you try to [side-load](https://code.visualstudio.com/docs/extensions/install-extension#_sharing-privately-with-others-sideloading) the extension?  Basically:
- git clone/pull
- npm install
- vsce package 
- code Go-xxx.vsix   <- the vsix created by the step before

mattetti:
@egamma it turned out the internet connection at the conference wasn't good enough for the live streaming so we postponed the event. I however showed off VS Code to many people, especially because there was a talk on Guru (used by VS Code) and Delve (the debugger used by Code). Lots of people were very impressed by how fast and feature full VS Code for Go is.

I'll side-load master and show the refactoring function tomorrow during the hack day.

Thanks a lot

mattetti:
Note to others finding this issue, you need to have vsce installed globally to side load the extension:

`npm install -g vsce`

egamma:
@mattetti thanks for the update and for sharing vscode-go with the Go developers. 

>Note to others finding this issue, you need to have vsce installed globally to side load the extension:
 npm install -g vsce 

Thanks, I've updated the instructions [above](https://github.com/Microsoft/vscode-go/pull/370#issuecomment-231684759).

mattetti:
One small detail: make sure to restart Code (I forgot the first time around ;))

A few things about this new feature, while it's the promise of this refactoring tool is really exciting, the practical use case seems to be pretty limited. I couldn't extract code with return statements, I saw weird errors of undefined types (no idea what was going on) and finally this interesting error: `Error: Completing the transformation will introduce the following error: <stdin>:201:2: no new variables on left side of :=`

The error is interesting because it's meaningless for the user, I have not idea what code is on line 201 of the stdin buffer, and no idea why it ended being an error. My guess is that my extracted code was added as a function but the new code wasn't valid (it needed some massaging?), so the underlying tool failed with this cryptic error.

@ramya-rao-a did you have good luck using this refactoring tool? It's highly possible I got very unlucky and the code I was trying to refactor wasn't a good example.

lukehoban:
I just had a chance to try out this feature.  In the 4 projects I tried it in, and half dozen locations trying to extract method in each, I didn't actually find a single case where it worked :-).

Are we sure that GoDoctor is robust enough for this integration, and if so, that the integration is working sufficiently often that this feature is a net win for the extension?

Here are a few of the errors I saw during an Extract Method:

* `Failed to parse the patches from godoctor: Illegal escape in patch_fromText: default:`
* `Error: Completing the transformation will introduce the following error: <stdin>:59:2: no new variables on left side of :=`
* `Error: could not import golang.org/x/net/http2/hpack (cannot find package "golang.org/x/net/http2/hpack" in any of:`
* `Error: Code containing return statements may change behavior if it is extracted.`

I also agree with @mattetti that the last of these makes this hard to use in practice in Go, since return statements are so common in error checking interspersed through Go function bodies.  

Also, in most of the projects I tried, these changes appear to have broken formatting with `Cannot read property 'apply' of undefined`.  And format-on-save, which is probably the most used feature of the extension, doesn't appear to get triggered at all anymore (perhaps the error above is causing that?).

The issues with `format` will need to get fixed before we can push an update for the extension, and I suspect it may be best to revert this feature from master until it is more polished such that it works a reasonable percent of the time that users try to use it.

mattetti:
I talked with my friend Fatih who develops and maintains go-vim. He told me
he chose to not integrate this tool on purpose, apparently because it
doesn't seem to work quite right and he would have to  deal with support.

I am excited about the opportunity of having a great refactoring/function
extractor but I would suggest to wait until we have a more solid solution.

ramya-rao-a:
godoctor does have limitations in the [extract method](http://gorefactor.org/doc.html#refactoring-extract) feature. 

> Code containing return statements, defer statements, or anonymous functions cannot be extracted

@mattetti My examples did not contain any of the above and the refactoring did work.

@lukehoban The code that applies the diffs has been extracted out of goFormat into utils. There was a missing check on null which broke the format feature which can be easily fixed.

Regardless, since it seems like godoctor doesn't extract the method well in many common scenarios like you both have experienced, I will revert the changes for now.

lukehoban:
Thanks @ramya-rao-a, and sorry for not reviewing this sooner! 

It would be great to reopen this PR with the fixes for formatting so that folks can try out the Extract Method feature and provide feedback that may help inform whether GoDoctor could be further improved to be a useful extension.

ramya-rao-a:
@lukehoban I have reopened the PR after fixing the issue with formatting. #404  
I am curious about few of the errors you encountered.

Can you provide me sample go code where you found the below errors? It would help me determine if the problem is with godoctor or my integration.

* `Failed to parse the patches from godoctor: Illegal escape in patch_fromText: default:`
* `Error: Completing the transformation will introduce the following error: <stdin>:59:2: no new variables on left side of :=`
* `Error: could not import golang.org/x/net/http2/hpack (cannot find package "golang.org/x/net/http2/hpack" in any of:`

Thanks!

PR #388, opened 2016-07-12
Preserve focus when running tests

oliverkofoed:
Use the preserveFocus[1] argument on OutputChannel.Show() to ensure that the editor does not lose focus when running tests.

This is, for instance, useful when you have a setup such as this keybinding that runs tests in the current package whenever cmd+t is pressed:

[
   {
       "key":"cmd+t",
       "command":"go.test.package",
       "when":"editorTextFocus && editorLangId=='go'"
   }
]

[1] https://github.com/Microsoft/vscode/commit/1da94d28ee74302bc30747403a314c49129519be

### DIFF
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -119,7 +119,7 @@ export function testCurrentFile(timeout: string) {
 function goTest(config: TestConfig): Thenable<boolean> {
 	return new Promise<boolean>((resolve, reject) => {
 		outputChannel.clear();
-		outputChannel.show(2);
+		outputChannel.show(2, true);
 		let buildFlags: string[] = vscode.workspace.getConfiguration('go')['buildFlags'];
 		let buildTags: string = vscode.workspace.getConfiguration('go')['buildTags'];
 		let args = ['test', '-v', '-timeout', config.timeout, '-tags', buildTags, ...buildFlags];


lukehoban:
The test appears to be failing because this overload is not in the vscode.d.ts.  I tried updating vscode 0.11.14 which appears to be the latest, but I still don't see this being present.  I do see the overload present at https://raw.githubusercontent.com/Microsoft/vscode/master/src/vs/vscode.d.ts.  I'm not sure why there is a discrepancy here, but would be good to make sure this change builds correctly before merging.

/cc @jrieken

jrieken:
Latest is here: https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L2803. It contains a deprecated signature that takes a `column` and the `preserve` flag and a newer which only has the `preserve` flag. 

@lukehoban Does you build script maybe pull in an older version of `vscode.d.ts`?

ironcladlou:
@lukehoban did you figure that out? I was pulling this locally (because I like the change) and am running into the same issue with vscode-go master (+ some of my PRs) and vscode 1.5.2.

ramya-rao-a:
Closing this PR as #577 has done the same

PR #396, opened 2016-07-18
Revert "Add command to extract method out of selected code"

ramya-rao-a:
Reverts Microsoft/vscode-go#370
Godoctor is not robust enough for common scenarios. Therefore reverting this change


lukehoban:
@donatj See discussion at https://github.com/Microsoft/vscode-go/pull/370.

PR #404, opened 2016-07-25
Add command to extract method out of selected code

ramya-rao-a:
Related to a feature ask in #275 
 This commit adds a command to the command palette that can be used to extract a method out of selected code and replace the selected code with a call to the extracted method using godoctor. Read more about godoctor at http://gorefactor.org/

godoctor does have limitations in the [extract method](http://gorefactor.org/doc.html#refactoring-extract) feature. 

> Code containing return statements, defer statements, or anonymous functions cannot be extracted

Do provide feedback if godoctor seems like a wise choice for the extract method feature


ramya-rao-a:
Closing this PR in favor of #2139

PR #406, opened 2016-07-25
Move away from deprecated language configuration settings

aeschli:
In the July release of VSCode we added new properties to the language configuration files so that the deprecated __characterPairSupport and __electricCharacterSupport APIs in vscode.d.ts are no longer necessary.
See Microsoft/vscode#9281.

I've already added the comment, brackets and bracket pairs to the go language definition in the built-in go extension. See Microsoft/vscode@8020da609fefa8a96e8b538ecba0e6da1e2de2e3 

You can keep your current state a while if you want to be backward compatible to pre-July versions.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -86,35 +86,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 			// ^.*\{[^}'']*$
 			increaseIndentPattern: /^.*\{[^}'']*$/
 		},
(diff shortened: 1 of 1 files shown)


aeschli:
Oh, that looks like a bug. `wordPattern` has to stay.

ramya-rao-a:
Closing this in favor of #587

Issue #410, opened 2016-07-27
delve debugger non-functional in VSCode 1.3.1


twexler:
Running on OS X 10.11.5, VSCode 1.3.1, vscode-go 0.6.42, delve 0.11.0-alpha.

Just upgraded to 1.3.1, `launch.json` has been unchanged since delve worked on a previous version. When I try to run *any* of my launch configurations the debug controls come up, the progress indicator on the debug side panel spins and nothing happens. 

I can't find anything relevant in the developer console, in OS X's logs or any other relevant location. `dlv` works fine on the command line.

manucorporat:
Same here 😢

ryanvz:
In 0.6.38, `--log` was always set in the call to Delve.  A boolean to show logs was added to `src/debugAdapter/goDebug.ts` in 0.6.40.  The plugin does not seem to work with logging disabled.
 
**Workaround**
Add `"showLog": true` to `launch.json` in your go project:
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Launch",
      "type": "go",
      "request": "launch",
      "mode": "debug",
      "remotePath": "",
      "port": 2345,
      "host": "127.0.0.1",
      "program": "${workspaceRoot}",
      "env": {},
      "args": [],
      "showLog": true
    }
  ]
}
```

twexler:
Looks like adding `"showLog": true` to my launch configs fixed it, thanks @ryanvz!

PR #412, opened 2016-07-28
Add showLog to initialConfigurations

ryanvz:
Related to [delve debugger non-functional in VSCode 1.3.1](https://github.com/Microsoft/vscode-go/issues/410)

`showLog` was added as a new option in 0.6.40.  When set to false (the default), the debugger appears not to work.  Adding `"showLog": true` to the initialConfigurations allows the debugger to work with the auto generated launch.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,8 @@
             "host": "127.0.0.1",
             "program": "${workspaceRoot}",
             "env": {},
-            "args": []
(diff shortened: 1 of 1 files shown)


lukehoban:
I think we ultimately want to turn this off by default, but given that its causing problems, pulling this change in for now.

PR #413, opened 2016-07-30
update dependencies, add goimports, add channel select snippets

leaxoy:
I have update dependencies (lodash 3->4), add goimports dep, and add **`channel select`** snippet.
```
"channel select": {
	"prefix": "sl",
	"body": "select {\ncase ${1:condation}: \n\t${2:break}\ndefault:\n\t${0}\n}"
}
```

Now I am consider to add some command to `command pattern`, like `go run`, `go get`, `go get all`, `go build` , etc.


snippets/go.json
```diff
@@ -155,6 +155,10 @@
 		"test function": {
 			"prefix": "tf",
 			"body": "func Test$1(t *testing.T) {\n\t$0\n}"
+		},
+		"channel select": {
+			"prefix": "sl",
+			"body": "select {\ncase ${1:condation}: \n\t${2:break}\ndefault:\n\t${0}\n}"
```
lukehoban: typo?

src/goInstallTools.ts
```diff
@@ -22,6 +22,7 @@ let tools: { [key: string]: string } = {
 	'golint': 'github.com/golang/lint/golint',
 	'go-outline': 'github.com/lukehoban/go-outline',
 	'go-symbols': 'github.com/newhook/go-symbols',
+	'goimports': 'golang.org/x/tools/cmd/goimports',
```
lukehoban: Why is this needed?  Goreturns is the default tool, which should be a superset of `goimports`.  Is there an issue tracking a scenario for which this is needed (and for which we should have all vscode-go users download this additional binary?)

.vscode/launch.json
```diff
@@ -1,29 +0,0 @@
-{
```
lukehoban: These files shouldn't be removed.  They are needed to build and debug the extension from within VS Code.

lukehoban:
Took care of the dependency upgrade in https://github.com/Microsoft/vscode-go/commit/61a886c956193ecbca9e61472cbc04401966ea7f.  For the rest, if there are remaining changes you think should be made - want to reopen as separate PRs?

PR #424, opened 2016-08-11
Extra info now checks doc

### LINKED ISSUE #87: Add godoc to hover provider
I noticed in the javascript editor hovering over a function shows a hover popout with the function documentation.  It would be nice to add support for this.

abarisain:
This patch enables godoc provided documentation on function hover.

capture

I have no idea why this was implemented but not enabled. Was there any reason?
This may not be desirable since it runs two programs on hover, and can be improved in a followup PR to use gogetdoc. That said, these processes are called everytime the developer writes a function call, which happens (at least for me) way less often than hovering a method.

It's also better than the documentation on parenthesis open, since it is not truncated.
Note that this does not enable the godoc in resolveCompletionItem, that may come in another PR :)

Fixes #87

### DIFF
--- a/src/goExtraInfo.ts
+++ b/src/goExtraInfo.ts
@@ -10,7 +10,7 @@ import { definitionLocation } from './goDeclaration';
 
 export class GoHoverProvider implements HoverProvider {
 	public provideHover(document: TextDocument, position: Position, token: CancellationToken): Thenable<Hover> {
-		return definitionLocation(document, position, false).then(definitionInfo => {
+		return definitionLocation(document, position, true).then(definitionInfo => {
 			if (definitionInfo == null) return null;
 			let lines = definitionInfo.lines;
 			lines = lines.map(line => {
@@ -28,7 +28,12 @@ export class GoHoverProvider implements HoverProvider {
 			} else {
 				text = lines[0];
 			}
-			let hover = new Hover({ language: 'go', value: text });
+			let hoverTexts = [];
+			if (definitionInfo.doc != null) {
+				hoverTexts.push({ language: null, value: definitionInfo.doc});
+			}
+			hoverTexts.push({ language: 'go', value: text});
+			let hover = new Hover(hoverTexts);
 			return hover;
 		});
 	}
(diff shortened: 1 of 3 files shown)


abarisain:
Whoops, the tests are broken. Sorry about that, I'll issue a fix commit soon

test/go.test.ts
```diff
@@ -32,19 +32,29 @@ suite('Go Extension Tests', () => {
 
 	test('Test Hover Provider', (done) => {
 		let provider = new GoHoverProvider();
-		let testCases: [vscode.Position, string][] = [
+		let printlnDoc = `Println formats using the default formats for its operands and writes to
+standard output. Spaces are always added between operands and a newline
+is appended. It returns the number of bytes written and any write error
+encountered.
+`
```
abarisain: Not sure I really like how templates strings break indentation, but it does not destroy the file's column count that way

abarisain:
The only failing tests are gometalinter's

mattetti:
+1 I tested it locally and it works great. I believe that's the expected behavior and the way it works when editing TS.

src/goExtraInfo.ts
```diff
@@ -28,7 +28,12 @@ export class GoHoverProvider implements HoverProvider {
 			} else {
 				text = lines[0];
 			}
-			let hover = new Hover({ language: 'go', value: text });
+			let hoverTexts = [];
+			if (definitionInfo.doc != null) {
+				hoverTexts.push({ language: null, value: definitionInfo.doc});
```
lukehoban: I think that passing this as a string will format in a slightly different way which matches what other extensions do: https://github.com/Microsoft/vscode/search?utf8=%E2%9C%93&q=%22new+Hover%22

lukehoban:
Thanks for this - big improvement.

abarisain:
Thanks for merging :)

src/goExtraInfo.ts
abarisain: Sorry about that, didn't notice that there was a simple string constructor for this

Issue #343, opened 2016-05-28
Support variable substitution for go.path setting


osechet:
It would be nice if environment variable substitution was possible with the `go.path` setting. I use a separate workspace for each of my go projects and I append the current project path to the `GOPATH` variable. That way I keep the shared tools (gofmt, golint...) together and I don't have to download them for each project.
With the `go.path` setting I can configure vscode to work this way, however I have to "hard code" my path: 
`"go.gopath": "C:/devel/go;${workspaceRoot}"`
where `C:/devel/go` is my `GOPATH`.

I would like to be able to set it as follow:
`"go.gopath": "${env.GOPATH};${workspaceRoot}"`

lukehoban:
> I would like to be able to set it as follow: `"go.gopath": "${env.GOPATH};${workspaceRoot}"`

This is something that VS Code would need to support - I believe https://github.com/Microsoft/vscode/issues/2809 is tracking this, and https://github.com/Microsoft/vscode/issues/3759 was another similar report related to `vscode-go`.

> That way I keep the shared tools (gofmt, golint...) together and I don't have to download them for each project.

FWIW - Addressing that particular issue without forcing you to use a dual GOPATH is what https://github.com/Microsoft/vscode-go/issues/5 has been tracking.

osechet:
> This is something that VS Code would need to support
Yes I agree, it would be better if it was supported by vscode natively. However, this extension is already doing it for $workspaceRoot [(here)](https://github.com/Microsoft/vscode-go/blob/master/src/goInstallTools.ts#L50) so why not for environment variables? I saw all the issues you are refering to, but all of them have been closed or postponed.

About #5, I don't think managing the go tools in a specific way to the IDE is a good idea. What if I want to use another editor/IDE? Should the go tools be duplicated for each IDE I use? I think it's best to work as suggested by golang and with the solutions provided by golang. Multiple GOPATH is such a solution: one path for all the shared tools and dependencies, another one for the current project, then the vendor directory for the specific dependencies.

osechet:
I already implemented this in my own repo. I can create a pull request if you are interested.

KevinSjoberg:
I actually am asking for the same thing over at #928. Personally, I think it makes sense to support the following variables: https://code.visualstudio.com/docs/editor/tasks#_variable-substitution.

ramya-rao-a:
This feature is now available in the latest update to the Go extension (0.7.0)
Thanks to @sphawk!

PR #448, opened 2016-08-22
Substitute environment variables in gopath setting

osechet:
Related to issues #343 and Microsoft/vscode#3759.
I would like to propose this PR. I understand this solution is not as good as a real support from vscode. However it does the job without changing the normal behavior.


ramya-rao-a:
Closing this PR in favor of https://github.com/Microsoft/vscode-go/pull/1743

PR #460, opened 2016-08-30
Fixes #13. Use diffs from the gorename tool instead of letting gorename update the file on disk. 

### LINKED ISSUE #13: Cannot undo rename
If you use the rename functionality then press CMD+Z nothing happens

ramya-rao-a:


src/util.ts
```diff
@@ -94,4 +95,206 @@ export function canonicalizeGOPATHPrefix(filename: string): string {
...
+
+/**
+ * Gets Edits from given diff array
+ *
+ * @param diffs The array of diffs which are translated to edits
+ * @param line The line number from where the edits are to be applied
+ * @param isEachDiffSeparateLine boolean that denotes if each diff corresponds to a complete separate line
+ * @returns Array of Edits that can be applied to the document
+ */
+export function GetEditsFromDiffs(diffs: dmp.Diff[], line: number, isEachDiffSeparateLine: boolean = false): Edit[] {
```
lukehoban: camelCase

src/util.ts
```diff
@@ -94,4 +95,206 @@ export function canonicalizeGOPATHPrefix(filename: string): string {
...
+	return edits;
+}
+
+/**
+ * Applies patch_fromText on given patch text to get patches and then returns edits for each of the patches
+ *
+ * @param patchText: Text starting with @@ following the Unified format for patches.
+ * Read more http://www.gnu.org/software/diffutils/manual/diffutils.html#Unified-Format
+ */
+function GetEditsFromPatchText(patchText: string): Edit[] {
```
lukehoban: camelCase

src/util.ts
```diff
@@ -94,4 +95,206 @@ export function canonicalizeGOPATHPrefix(filename: string): string {
 		}
 	}
 	return filename;
+}
```
lukehoban: Perhaps pull all this Edit processing code into a separate file?

lukehoban:
Nice!  Great to be able to undo renames finally :-).

It looks like the test failure is a real problem with the newly added tests.

I noticed that the cursor moves sometimes after applying the rename.  I believe this was not the case previously - though it may have been an existing problem?

There's a lot of new custom logic related to parsing the diff.  Any chance there's an existing Node package that handles this?  It would be great not to have something custom for this extension if possible.

ramya-rao-a:
@lukehoban half the custom logic was pulling things out of `goFormat.ts`. Rest is to parse the diff output and fetch diffs which was needed if we were to continue using the `diff_match_patch` library. I'll see if there is an existing Node package that handles the diff output parsing.

Regarding the cursor, I didn't notice the difference between before and after. Will take a look

Thanks!

PR #461, opened 2016-08-31
Add go.formatFlags configuration parameter to pass parameters to form…

### LINKED ISSUE #444: Need go.formatFlags to support gofmt -s simplify mode
Our build server enforces the use of gofmt with the simplify mode. Right now I don't see a way to pass the -s flag to the formatting tool in vscode. Can we add a go.formatFlags setting?

Example code for testing:

```
var final = strings.Split("a b c", " ")
var first = strings.Split("d e f", " ")
final = append(final, first[2:len(first)]...)
//gofmt -s main.go would change the previous line to
final = append(final, first[2:]...)
```

bredov:
…at tool. Fixes #444

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,6 +267,14 @@
           "default": "goreturns",
           "description": "Pick 'gofmt', 'goimports' or 'goreturns' to run on format."
         },
+        "go.formatFlags": {
+          "type": "array",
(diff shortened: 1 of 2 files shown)


lukehoban:
Looks great.

PR #463, opened 2016-08-31
Fix the executable name so ".exe" doesn't end up in the install warning

abarisain:
On windows, when the format tool doesn't exist, vscode-go will show ".exe" at the end of the command, breaking the installation.

capture

This patch fixes it.


ramya-rao-a:
The fix is now available in the latest version of the Go extension : 0.6.44

PR #471, opened 2016-09-08
Adding run main and kill command

Centny:
adding kill running test or main
adding run main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,9 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.formatFlags": [],
 	"go.goroot": "/usr/local/go",
 	"go.gopath": "/Users/lukeh/go",
-	"go.gocodeAutoBuild": false
+	"go.gocodeAutoBuild": false,
+	"go.runDir": "${file_dir}",
+	"go.runArgs": []
 }
 ```
 
@@ -95,9 +97,11 @@ In addition to integrated editing features, the extension also provides several 
 
 * `Go: Add Import` to add an import from the list of packages in your Go context
 * `Go: Current GOPATH` to see your currently configured GOPATH
+* `Go: Run main or test at curso` to run main or test in the current active document
 * `Go: Run test at cursor` to run a test at the current cursor position in the active document
 * `Go: Run tests in current package` to run all tests in the package containing the active document
 * `Go: Run tests in current file` to run all tests in the current active document
+* `Go: Kill running test or main` to kill running test or main
 
 ### _Optional_: Debugging
 
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -24,7 +24,7 @@ import { updateGoPathGoRootFromConfig, setupGoPathAndOfferToInstallTools } from 
 import { GO_MODE } from './goMode';
 import { showHideStatus } from './goStatus';
 import { coverageCurrentPackage, getCodeCoverage, removeCodeCoverage } from './goCover';
-import { testAtCursor, testCurrentPackage, testCurrentFile } from './goTest';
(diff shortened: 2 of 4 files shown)


src/goTest.ts
```diff
@@ -112,6 +140,22 @@ export function testCurrentFile(timeout: string) {
 }
 
 /**
+ * Kill all running go task.
+ *
+ */
+export function killRunning() {
```
ironcladlou: Kind of scary to do a blanket kill with a pattern match. Would be far safer to track PIDs if anything.

@lukehoban, process management seems like something VScode or Electron should provide some facilities for. Any clue?

src/goTest.ts
```diff
@@ -71,6 +73,32 @@ export function testAtCursor(timeout: string) {
...
 /**
+* Executes the unit test or main at the primary cursor using `go run/test`. Output
+* is sent to the 'Go' channel.
+* 
+* @param timeout a ParseDuration formatted timeout string for the tests.
+*
+* TODO: go test returns filenames with no path information for failures,
+* so output doesn't produce navigable line references.
+*/
+export function runAtCursor(timeout: string) {
```
ironcladlou: IMO it would be clearer if each function had a single conceptual responsibility. A function to run a program/main, and a function to run a test. I'm not quite seeing the benefit of grouping these together.

src/goMain.ts
```diff
@@ -57,6 +57,15 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		vscode.window.showInformationMessage('Current GOPATH:' + gopath);
 	}));
 
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.kill.running', () => {
+		killRunning();
+	}));
+
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.test.cursor', () => {
```
ironcladlou: The `go.test.cursor` command is already registered [here](https://github.com/Centny/vscode-go/blob/fd9d200269c632b68987463c2ead869eca0e69f6/src/goMain.ts#L64).

README.md
```diff
@@ -95,9 +97,11 @@ In addition to integrated editing features, the extension also provides several
 
 * `Go: Add Import` to add an import from the list of packages in your Go context
 * `Go: Current GOPATH` to see your currently configured GOPATH
+* `Go: Run main or test at curso` to run main or test in the current active document
```
ironcladlou: s/curso/cursor

README.md
```diff
@@ -56,7 +56,9 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.formatFlags": [],
 	"go.goroot": "/usr/local/go",
 	"go.gopath": "/Users/lukeh/go",
-	"go.gocodeAutoBuild": false
+	"go.gocodeAutoBuild": false,
+	"go.runDir": "${file_dir}",
+	"go.runArgs": []
```
ironcladlou: I think this brings the feature a little close to overlapping with the task system. It seems highly unlikely that a single project-scoped argument array for a main will be applicable to every main you might try to run with an ad-hoc "run main at cursor" command. If you need to statically define arguments and stuff, there's a robust task runner system already in place.

src/goTest.ts
Centny: @ironcladlou 
agreed,i am using the task system instead.

src/goTest.ts
Centny: @ironcladlou 
It is only track PIDs which started by Go:Test, indeed not safed.
but ,to track running test which is started by vscode-go plugin is hard for VScode, i think. 

finally, this features to kill running process is very useful when coding by TDD, it will pull request again.

PR #478, opened 2016-09-15
Add a command to execute the previous test

ironcladlou:
Add a `go.test.previous` command which runs the last test which was
executed.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -24,7 +24,7 @@ import { updateGoPathGoRootFromConfig, setupGoPathAndOfferToInstallTools } from 
 import { GO_MODE } from './goMode';
 import { showHideStatus } from './goStatus';
 import { coverageCurrentPackage, getCodeCoverage, removeCodeCoverage } from './goCover';
-import { testAtCursor, testCurrentPackage, testCurrentFile } from './goTest';
+import { testAtCursor, testCurrentPackage, testCurrentFile, testPrevious } from './goTest';
 import { addImport } from './goImport';
 import { installAllTools } from './goInstallTools';
 
@@ -70,6 +70,10 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	ctx.subscriptions.push(vscode.commands.registerCommand('go.test.file', () => {
 		let goConfig = vscode.workspace.getConfiguration('go');
 		testCurrentFile(goConfig['testTimeout']);
+	}));
+
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.test.previous', () => {
+		testPrevious();
 	}));
 
 	ctx.subscriptions.push(vscode.commands.registerCommand('go.test.coverage', () => {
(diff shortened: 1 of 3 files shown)


ironcladlou:
@lukehoban do you have any idea what's up with those Travis failures?

ramya-rao-a:
@ironcladlou The test for gometalinter tries to set the user settings for `linttool`. 
```
let config = vscode.workspace.getConfiguration('go');
config['lintTool'] = 'gometalinter';
```
which fails with the latest version of VSCode as you can no longer update config that way.

There have been changes to the configuration service https://github.com/Microsoft/vscode/issues/1396
I'll look into updating the test case

src/goMain.ts
```diff
@@ -72,6 +72,11 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		testCurrentFile(goConfig['testTimeout']);
 	}));
 
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.test.previous', () => {
+		let goConfig = vscode.workspace.getConfiguration('go');
+		testPrevious();
```
lukehoban: Do you intend to pass `testTimeout` here as in other test commands?  Or to not use `goConfig`?

src/goTest.ts
```diff
@@ -112,12 +116,32 @@ export function testCurrentFile(timeout: string) {
 }
 
 /**
+ * Runs the previously executed test.
+ */
+export function testPrevious() {
+	let editor = vscode.window.activeTextEditor;
+	if (!editor) {
```
lukehoban: Why does an editor need to be active in this case?  It sounds like the semantics for this command are not tied to what is in the current editor (if there is one).  It just executes whatever the last Go test was that was invoked.

src/goMain.ts
lukehoban: Actually - looks like you are intentionally not passing `testTimeout` - so should drop the `goConfig` line.

src/goTest.ts
ironcladlou: You're right, I'll remove that check.

src/goMain.ts
ironcladlou: Removed the config lookup for now.

src/goTest.ts
ironcladlou: Removed the editor check.

ironcladlou:
@ramya-rao-a, thanks for tracking down that Travis issue!

ironcladlou:
Added a small commit to rename the title of the command to something more auto-complete friendly (as a complement to https://github.com/Microsoft/vscode-go/pull/495).

ironcladlou:
Squashed commits.

ironcladlou:
Will this be merged?

ramya-rao-a:
Sure, I tested your changes and they look good.

Could you trigger another travis runs so that the tests are all green? The last time it failed due to an upstream issue with vscode-debugadapter

ironcladlou:
@ramya-rao-a Thanks! How do I trigger a new build without making some dummy commit?

PR #480, opened 2016-09-15
Fixes #479 Do not update workspace config to test gometalinter

### LINKED ISSUE #479: Test for Gometalinter failing as configuration is read-only
The test for gometalinter tries to set the user settings for `linttool`. 
```
let config = vscode.workspace.getConfiguration('go');
config['lintTool'] = 'gometalinter';
```

This fails in the latest version of VSCode (1.5 and higher), as you can no longer update config that way.

Interestingly the comment for the above config in vscode.d.ts says it is a readonly dictionary, though it has been writable till the last release

There have been changes to the configuration service in VS Code in August and more to come in September 

I'd say we comment the test case for now until vscode comes up with the right way to set the configs

ramya-rao-a:


lukehoban:
Merged #481 as a fix for this.

PR #482, opened 2016-09-15
Enable testflags via settings and keyboard bindings

### LINKED ISSUE #401: Support running tests without the -v flags
Sometimes, specially when tests are failing, it's useful to omit the -v flag to prevent noise. It would be nice to make the flag optional or let users include `-v` in extra flags.

ramya-rao-a:
(Updated on 21/1/2017)

Fixes #401 and #683

Added a new setting `go.testFlags`
    - If null, then all the test commands will continue to use the `go.buildFlags` and '-v' while running `go test` as always.
    - Else, the provided flags will be used by the test commands  instead of `go.buildFlags`  while running `go test`

The usage of `go.testTimeout` and `go.buildTags` will continue to remain as is.

Users can also customize the flags and have keyboard shortcuts for the same.

```json
{
        "key": "ctrl+shift+t",
        "command": "go.test.file",
        "args": {
            "flags": ["-short"]
        },
        "when": "editorTextFocus"
}
```

### DIFF
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -27,6 +27,10 @@ interface TestConfig {
 	 */
 	goConfig: vscode.WorkspaceConfiguration;
 	/**
+	 * Test flags to override the testFlags and buildFlags from goConfig.
+	 */
+	flags: string[];
+	/**
 	 * Specific function names to test.
 	 */
 	functions?: string[];
@@ -45,7 +49,7 @@ let lastTestConfig: TestConfig;
 * TODO: go test returns filenames with no path information for failures,
 * so output doesn't produce navigable line references.
 */
-export function testAtCursor(goConfig: vscode.WorkspaceConfiguration) {
+export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, args: any) {
 	let editor = vscode.window.activeTextEditor;
 	if (!editor) {
 		vscode.window.showInformationMessage('No editor is active.');
@@ -72,6 +76,7 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration) {
 		return goTest({
 			goConfig: goConfig,
 			dir: path.dirname(editor.document.fileName),
+			flags: getTestFlags(goConfig, args),
 			functions: [testFunction.name]
 		});
 	}).then(null, err => {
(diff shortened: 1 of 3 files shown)


src/goTest.ts
```diff
@@ -122,7 +122,8 @@ function goTest(config: TestConfig): Thenable<boolean> {
 		outputChannel.show(2);
 		let buildFlags: string[] = vscode.workspace.getConfiguration('go')['buildFlags'];
 		let buildTags: string = vscode.workspace.getConfiguration('go')['buildTags'];
-		let args = ['test', '-v', '-timeout', config.timeout, '-tags', buildTags, ...buildFlags];
+		let testFlags: string[] = vscode.workspace.getConfiguration('go')['testFlags'];
+		let args = ['test', ...testFlags, '-timeout', config.timeout, '-tags', buildTags, ...buildFlags];
```
ironcladlou: I like the idea of using `testFlags` for build/test flags to the test command. Should we remove support for `buildFlags` here, given there is no distinction in `go test` between build and test flags? Having both is redundant and confusing to me given your new option.

ramya-rao-a: Agreed. Ideally 
- `buildFlags` should be used for build only, 
- `testFlags` should be used for test only.
- `testTimeout` should not be a separate setting and should be included in the `testFlags` setting

If we do this, we won't be backward compatible and everyone who had configured buildFlags and/or testTimeout will have to update their settings to add the new `testFlags` 

@lukehoban Thoughts?

ironcladlou: @ramya-rao-a
>testTimeout should not be a separate setting and should be included in the testFlags setting

Hm. Wouldn't it then follow that `buildTags` should be merged into `buildFlags`? After all,  `-tag` is just another `buildFlag`. I do see value in adding convenient strongly typed wrappers around the most common flags, but those would need to be easily omitted so that users can have absolute control over flags.

ramya-rao-a:
cc @roblourens 
ping @ironcladlou

roblourens:
LGTM

ramya-rao-a:
@ironcladlou I wish you had a chance to look at this PR. I am planning to release an update in a few days, therefore will go ahead and merge this PR so that I can go ahead with some end to end testing.

When you get some time, please do add your thoughts here.

PR #491, opened 2016-09-20
Fixes #316 Replace full path for vendor packages with relative path

### LINKED ISSUE #316: Go: Add Imports adds full path to vendored dependency, instead of the vendored path
Let's say you have a dependency vendored at <$GOPATH/src/vendor/github.com/foo/bar>. If you use cmd+shift+p > Go: Add Import to import that dependency into your current working file, it will add the string "vendor/github.com/foo/bar" instead of "github.com/foo/bar" to that file's imports. Since vendoring is now officially suppported in Go 1.6 (and support for ignoring vendoring will be removed in Go 1.7), I think that the Go: Add Import functionality should conform to the standard.

ramya-rao-a:
PR to fix #316 

This PR uses the prefix from the full path of vendor packages, appends that to $GOPATH to get the path to the vendor folder, tries to match it with current file path to see if both belong to the same project. If yes, then the relative path of the vendor replaces the full path.

Another way to find relative paths for vendor packages is by using the govendor tool:
`govendor list -no-status +v` gives the list of vendor packages (with relative path) for current go project.
But using this adds a 1 second delay between executing the command `Go: Add Import` and showing the results in Windows. In Mac, the difference is much less.
The implementation using govendor is here: https://github.com/ramya-rao-a/vscode-go/blob/vendor-imports-using-govendor/src/goImport.ts

Thoughts on both approaches are welcome.

### DIFF
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -9,31 +9,91 @@ import vscode = require('vscode');
 import cp = require('child_process');
 import { getBinPath } from './goPath';
 import { parseFilePrelude } from './util';
-import { promptForMissingTool } from './goInstallTools';
 import { documentSymbols } from './goOutline';
+import { promptForMissingTool, isVendorSupported } from './goInstallTools';
+import path = require('path');
 
 export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document.fileName) : Promise.resolve([]);
-	let pkgsPromise = new Promise<string[]>((resolve, reject) => {
+	let vendorSupportPromise = isVendorSupported();
+	let goPkgsPromise = new Promise<string[]>((resolve, reject) => {
 		cp.execFile(getBinPath('gopkgs'), [], (err, stdout, stderr) => {
 			if (err && (<any>err).code === 'ENOENT') {
 				promptForMissingTool('gopkgs');
 				return reject();
 			}
 			let lines = stdout.toString().split('\n');
-			let sortedlines = lines.sort().slice(1); // Drop the empty entry from the final '\n'
-			return resolve(sortedlines);
+			if (lines[lines.length - 1] === '') {
+				// Drop the empty entry from the final '\n'
+				lines.pop();
+			}
+			return resolve(lines);
 		});
 	});
 
-	return Promise.all<string[]>([importsPromise, pkgsPromise]).then(values => {
-		let imports = values[0];
-		let pkgs = values[1];
-		if (imports.length === 0) {
-			return pkgs;
-		}
-		return pkgs.filter(element => {
-			return imports.indexOf(element) === -1;
+	return vendorSupportPromise.then((vendorSupport: boolean) => {
+		return Promise.all<string[]>([goPkgsPromise, importsPromise]).then(values => {
+			let pkgs = values[0];
+			let importedPkgs = values [1];
+
+			if (!vendorSupport) {
+				if (importedPkgs.length > 0) {
+					pkgs = pkgs.filter(element => {
+						return importedPkgs.indexOf(element) === -1;
+					});
+				}
+				return pkgs.sort();
+			}
+
+			let currentFileDirPath = path.dirname(vscode.window.activeTextEditor.document.fileName);
+			let workspaces: string[] = process.env['GOPATH'].split(path.delimiter);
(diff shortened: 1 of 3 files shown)


src/goImport.ts
```diff
@@ -9,31 +9,95 @@ import vscode = require('vscode');
...
+			let pkgs = values[0];
+			let importedPkgs = values [1];
+
+			if (!vendorSupport || !vscode.window.activeTextEditor) {
+				if (importedPkgs.length > 0) {
+					pkgs = pkgs.filter(element => {
+						return importedPkgs.indexOf(element) === -1;
+					});
+				}
+				return pkgs.sort().slice(1);
```
roblourens: Looks like this is removing a "" entry - do you know why that exists? Should this slice be a safer check? Maybe roll it into the filter?

src/goImport.ts
```diff
@@ -9,31 +9,95 @@ import vscode = require('vscode');
...
+					return;
+				}
+
+				let vendorIndex = pkg.indexOf('/vendor/');
+
+				// Check if current file and the vendor pkg belong to the same root project
+				// If yes, then vendor pkg can be replaced with its relative path to the "vendor" folder
+				if (vendorIndex > 0) {
+					let rootProjectForVendorPkg = path.join(currentWorkspace, pkg.substr(0, vendorIndex));
+					let relativePathForVendorPkg = pkg.substring(vendorIndex + 8);
```
roblourens: 8... '/vendor/'.length?

src/goImport.ts
```diff
@@ -9,31 +9,95 @@ import vscode = require('vscode');
...
+
+					if (relativePathForVendorPkg && currentFileDirPath.startsWith(rootProjectForVendorPkg)) {
+						// When a vendor package for a project/repo already exists outside the current project,
+						// pkgs list would already have an entry for it.
+						// In which case, we can skip adding the vendor package to the list
+						// Example: Assume package 'a' is installed from github.com/somerepo/a to $GOPATH
+						//          The current project github.com/someotherrepo/b may have a version of a as a vendor package
+						//          github.com/someotherrepo/b/vendor/github.com/somerepo/a
+						//          In this case, github.com/somerepo/a would already be returned by goPkgs
+						if (pkgs.indexOf(relativePathForVendorPkg) === -1) {
```
roblourens: Would it be worth it to get rid of this check for perf? It's only applied to vendored packages so maybe it's not a big deal but do people have 100s of vendored packages? You could make pkgs a Set, or do other things to de-dupe.

roblourens:
This seems pretty straightforward, so it doesn't seem like the delay to start up govendor is worth it?

src/goImport.ts
```diff
@@ -9,31 +9,95 @@ import vscode = require('vscode');
...
+				}
+			}
+
+			let updatedPackages: string[] = [];
+			pkgs.forEach(pkg => {
+				if (!pkg || importedPkgs.indexOf(pkg) > -1) {
+					return;
+				}
+
+				let vendorIndex = pkg.indexOf('/vendor/');
```
roblourens: Can packages not include '/vendor/' somewhere else in the name? Could I 'go get github.com/foo/vendor' then this might be a false positive.

src/goImport.ts
ramya-rao-a: the stdout from gopkgs is a `\n` separated pkg paths. So when you do a split, you end up with a trailing empty string. That's what this originally was, and I just moved the code around while refactoring. Will fix it nevertheless.

src/goImport.ts
ramya-rao-a: set is good idea, though in the end I'll have to convert it to array to have sorting

src/goImport.ts
ramya-rao-a: Below is from [Understanding and using the vendor folder](https://blog.gopheracademy.com/advent-2015/vendor-folder/)
>  However, if a package or a parent folder of a package contains folder named vendor it will be searched for dependencies using the vendor folder as an import path root. 

And while trying different combinations, I could confirm that by "parent", they mean "ancestor"
Since that is the exact check I make which is making sure that the "vendor" folder's ancestors are current package's ancestors too, we should be good.

Go doesn't stop you from creating a package by the name "vendor", so you could totally do a "go get github.com/foo/vendor". Any packages under this folder will be available for use to any Go projects under "github.com/foo/*"

src/goImport.ts
```diff
@@ -9,31 +9,91 @@ import vscode = require('vscode');
 import cp = require('child_process');
```
roblourens: LGTM

Issue #437, opened 2016-08-16
feature request: autocomplete non imported packages


trajber:
The Go plugin for IntelliJ has a very useful feature that allows code completion works even when the package is not imported yet.
On this example you can see **"fmt"** and **"os"** packages being used even with no imports at all.

![output](https://cloud.githubusercontent.com/assets/291612/17714200/2d2dc442-63d5-11e6-86bc-66b9f964cedb.gif)

derekperkins:
Duplicate of #407

ramya-rao-a:
This feature is now available in the latest version of the Go extension : 0.6.44

You need to set `go.autocompleteUnimportedPackages` to true to enable this feature.

Do give it a try and give feedback

We will turn it on by default in the next update

Issue #494, opened 2016-09-20
Go: Add Import command allows import for packages that are already imported


ramya-rao-a:
In the below sample go file, `fmt` is already imported

```
package main
import "fmt"

func main() {
     fmt.Println("Hello!")
}
```

Now run `Go: Add Import` command from the command palette.
The quick open shows `fmt` as an option as well.
Choose `fmt`, and an import to `fmt` gets added

```
package main
import "fmt"
import "fmt"

func main() {
     fmt.Println("Hello!")
}
```

Either, we shouldn't show an already imported package in the quick open or when chosen, show an info/error box saying that the package is already imported

lukehoban:
The current design of "Add Import" kind of relies on `gofmt`/`goimports` to clean up the imports instead of baking in too much fanciness to the `Add Import` command itself.  I'm not sure that's a bad thing.  But looks like the changes in https://github.com/Microsoft/vscode-go/pull/500 may add more of this formatting logic into `Add Import`.

Broader note - this extension tries as hard as it can to not have to interpret Go code at all, leaving that responsibility to utilities written in Go which can use the official parsers and type checking logic built into the Go runtime.  This `Add Import` feature is one of the places where the extension goes the furthest in trying to parse and interpret Go code.  It's already a bit brittle (regexps for parsing!), and I'm worried that making it more intelligent may lead to more additional cases where it breaks in unexpected ways.

ramya-rao-a:
@lukehoban 

Agreed that parsing is best left to utilities written in Go. Lucky for us you already have written one :)
`go-outline` returns imported packages along with rest of the declarations in a go file.

The need for removing duplicates from the packages list doesn't arise much for the `Add Import` command. Like you said, formatting would clean it up. Worst case, the duplicate would get marked with red squiggly lines getting the user's attention.

Personally, this ask is more for my work in #437 where I add importable packages in autocomplete suggestions. I make use of the [`listPackages()`](https://github.com/Microsoft/vscode-go/blob/master/src/goImport.ts#L14) method from "goImports.go" file. If a package is already imported then my autocomplete gives 2 suggestions instead of 1. Which is why I needed `listPackages()` to not return packages that are already imported.

[...]

PR #497, opened 2016-09-22
Add completion items for unimported packages

### LINKED ISSUE #407: Autocomplete Should Complete From Packages Which Haven't Been Imported Yet
Just thinking how vscode-go could be even more awesome for me.

I sometimes find myself hunting through APIs. For example...

- I know that in the `fmt` package, there is a function called `Prints` which formats to a string, but I can't remember the parameters.
- So, I type `fmt.Prints(`, but I don't see any autocompletion. This is because `fmt` hasn't been imported yet.
- I hit save, but it still doesn't import the package. I wonder why...
- I put a value into the method's parameters and hit save, but it still doesn't add the import automatically.
- I google the docs for the `fmt` package and realise that the function is actually called `Sprint` not `Prints`, so I fix it and hit save... the plugin now imports `fmt` and I can browse the parameters and return value in the popup.

I wanted to use vscode-go to help me browse the `fmt` package and choose the correct method, but it's not helping me to do that. Note how when I've typed `fmt.` I can't see anything...

autocomplete

It would be better if:

- I type `fmt.` and a list of possible functions and their parent packages appears. `fmt` is already in my path (but not listed in my imports), so all the methods from `fmt` are listed. I notice `Prints` isn't in the list.
- I notice that `fmt.Sprint` is there, I hit the down arrow a couple of times to choose it and press the open bracket. I'm in the parameters list, and I can see the popup detailing the parameters the method accepts.
- I hit `Save` and the `import` statement is updated as usual.

Does anyone else think this would be good?

ramya-rao-a:
This PR adds 2 features

1. **Auto-complete for package names that are not imported**. Example:
Assume you haven't imported `fmt`, type `f`, the auto-complete will give `fmt` as an option. Choose this option, the word will get completed and an import to `fmt` will be added to your go file

2. **Auto-complete for methods from packages that are not yet imported**. Example:
Assume you haven't imported `math`, type `math.`, the auto-complete will give all methods from `math` package as options. Choose any, the word gets completed and an import to `math` will be added to your go file

These features depend on updates to the CompletionItem API in VS Code 1.5

Fixes #407 and #437

### DIFF
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -61,36 +61,37 @@ function askUserForImport(): Thenable<string> {
 	});
 }
 
+export function getTextEditForAddImport(arg: string): vscode.TextEdit {
+	// Import name wasn't provided
+	if (arg === undefined) {
+		return null;
+	}
+
+	let {imports, pkg} = parseFilePrelude(vscode.window.activeTextEditor.document.getText());
+	let multis = imports.filter(x => x.kind === 'multi');
+	if (multis.length > 0) {
+		// There is a multiple import declaration, add to the last one
+		let closeParenLine = multis[multis.length - 1].end;
+		return vscode.TextEdit.insert(new vscode.Position(closeParenLine, 0), '\t"' + arg + '"\n');
+	} else if (imports.length > 0) {
+		// There are only single import declarations, add after the last one
+		let lastSingleImport = imports[imports.length - 1].end;
+		return vscode.TextEdit.insert(new vscode.Position(lastSingleImport + 1, 0), 'import "' + arg + '"\n');
+	} else if (pkg && pkg.start >= 0) {
+		// There are no import declarations, but there is a package declaration
+		return vscode.TextEdit.insert(new vscode.Position(pkg.start + 1, 0), '\nimport (\n\t"' + arg + '"\n)\n');
+	} else {
+		// There are no imports and no package declaration - give up
+		return null;
+	}
+}
+
 export function addImport(arg: string) {
 	let p = arg ? Promise.resolve(arg) : askUserForImport();
 	p.then(imp => {
-		// Import name wasn't provided
-		if (imp === undefined) {
-			return null;
-		}
-
-		let {imports, pkg} = parseFilePrelude(vscode.window.activeTextEditor.document.getText());
-		let multis = imports.filter(x => x.kind === 'multi');
-		if (multis.length > 0) {
-			// There is a multiple import declaration, add to the last one
-			let closeParenLine = multis[multis.length - 1].end;
-			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(closeParenLine, 0), '\t"' + imp + '"\n');
-			});
-		} else if (imports.length > 0) {
-			// There are only single import declarations, add after the last one
-			let lastSingleImport = imports[imports.length - 1].end;
-			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(lastSingleImport + 1, 0), 'import "' + imp + '"\n');
-			});
-		} else if (pkg && pkg.start >= 0) {
-			// There are no import declarations, but there is a package declaration
-			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(pkg.start + 1, 0), '\nimport (\n\t"' + imp + '"\n)\n');
-			});
-		} else {
-			// There are no imports and no package declaration - give up
-			return null;
-		}
+		let edit = getTextEditForAddImport(imp);
+		vscode.window.activeTextEditor.edit(editBuilder => {
+			editBuilder.insert(edit.range.start, edit.newText);
+		});
 	});
 }
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -90,8 +90,10 @@ export function addImport(arg: string) {
 	let p = arg ? Promise.resolve(arg) : askUserForImport();
 	p.then(imp => {
 		let edit = getTextEditForAddImport(imp);
-		vscode.window.activeTextEditor.edit(editBuilder => {
-			editBuilder.insert(edit.range.start, edit.newText);
-		});
+		if (edit) {
+			vscode.window.activeTextEditor.edit(editBuilder => {
+				editBuilder.insert(edit.range.start, edit.newText);
+			});
+		}
 	});
 }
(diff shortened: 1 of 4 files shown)


src/goSuggest.ts
```diff
@@ -153,5 +210,58 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+	}
+
+	// Return importable packages that match given word as Completion Items
+	private getMatchingPackages(word: string): vscode.CompletionItem[] {
+		if (!word) return [];
+		let completionItems = this.pkgsList.filter((pkgInfo: PackageInfo) => {
+			return pkgInfo.name.startsWith(word);
+		}).map((pkgInfo: PackageInfo) => {
+			let item = new vscode.CompletionItem(pkgInfo.name, vscode.CompletionItemKind.Keyword);
+			item.detail = 'Add import';
```
ramya-rao-a: I needed a place to suggest that choosing this completion item would add an import. But I guess we don't really need it

lukehoban:
Just trying this out - experience is overall really nice!

One thing I noticed, if you have `"go.useCodeSnippetsOnFunctionSuggest": true`, then the snippet generated when you complete a method call off of a package - like `bytes.Contains` don't cycle correctly with `tab` is the use of `bytes` trigged an import.

lukehoban:
Looks like you can get into cases where the completion list contains several copies of the same entry - referring to each of the fully-qualified but unique packages whose last sub-package is the selected name, as well as to the already imported name.

The latter is actually a bigger problem - if I already have imported the name, I now see it twice, once based on the actual name in scope and once as a suggestion for an Add Import.

I think this will be important to fix - at least before making this the default experience.

screen shot 2016-09-28 at 1 48 23 pm

src/goSuggest.ts
```diff
@@ -66,80 +74,129 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+				}
+			});
+			p.stdin.end(inputText);
+		});
+	}
 	// TODO: Shouldn't lib-path also be set?
 	private ensureGoCodeConfigured(): Thenable<void> {
-		return new Promise<void>((resolve, reject) => {
+		let pkgPromise = listPackages().then((pkgs: string[]) => {
+this.pkgsList = pkgs.map(pkg => {
```
lukehoban: Is it okay for this to be fixed at startup time, not up to date as new packages are installed into the GOPATH?

src/goSuggest.ts
lukehoban: I think this is going to need to include the pull import path name as additional context.  I think being clear that this will "Add an import" is also helpful.

src/goSuggest.ts
```diff
@@ -153,5 +210,58 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+
+		// There could be multiple dots in the line
+		// we are interested in the word preceeding the last one
+		let splits = line.split('.');
+		line = splits[splits.length - 2];
+
+		// There could be multiple words in the line
+		// we are interested in the last one
+		let wordmatches = null;
+		let pattern = /(\w+)/g;
```
lukehoban: Is this a correct regexp for Go identifies that could be packages?

lukehoban: Also - to avoid the while loop - can't you bind it to match at the end of the line?  In fact, can't you avoid most of this logic by just using the regexp to match the `.$` at the end of the line that you are sure is there?

src/goSuggest.ts
ramya-rao-a: I didnt want to run `listPackages` for every autocomplete operation and so added this basic placeholder for an actual caching mechanism. Plan was to see if users notice the missing packages and if they do, if a reload window is not too much of a hassle

ramya-rao-a:
Reagrding https://github.com/Microsoft/vscode-go/pull/497#issuecomment-250295676
This is why I had logged #494. If `listPackages()` does not return already imported packages, then autocomplete need not have to worry about them

src/goSuggest.ts
ramya-rao-a: Updated to use `/(\w+)\.$/g` and removed the while loop

ramya-rao-a:
Regarding https://github.com/Microsoft/vscode-go/pull/497#issuecomment-250294753

Using `additionalTextEdits` instead of `command` in the `CompletionItem` ensures that `useCodeSnippetsOnFunctionSuggest` feature is not broken

src/goImport.ts
```diff
@@ -31,36 +31,39 @@ function askUserForImport(): Thenable<string> {
...
-				editBuilder.insert(new vscode.Position(closeParenLine, 0), '\t"' + imp + '"\n');
-			});
-		} else if (imports.length > 0) {
-			// There are only single import declarations, add after the last one
-			let lastSingleImport = imports[imports.length - 1].end;
-			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(lastSingleImport + 1, 0), 'import "' + imp + '"\n');
+		let edit = getTextEditForAddImport(imp);
+		if (edit) {
+			vscode.window.activeTextEditor.edit(editBuilder => {
```
lukehoban: May not matter right now - but probably good to return the promise here so it get's chained into the `.then`.

src/goImport.ts
```diff
@@ -31,36 +31,39 @@ function askUserForImport(): Thenable<string> {
 	});
 }
 
+export function getTextEditForAddImport(arg: string): vscode.TextEdit {
```
lukehoban: `arg` should be `importPackageName`?

src/goImport.ts
```diff
@@ -31,36 +31,39 @@ function askUserForImport(): Thenable<string> {
...
-			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(pkg.start + 1, 0), '\nimport (\n\t"' + imp + '"\n)\n');
-			});
-		} else {
-			// There are no imports and no package declaration - give up
-			return null;
 		}
 	});
-}
+}
```
lukehoban: No need to remove final newlines.

ramya-rao-a:
@lukehoban 

> I do wonder if the lack of vendor packages in the listPackages list will be more noticeable with this feature.

Vendor packages do get returned by `listPackages()`. What are you referring to here?

>It's quite a visible change, so it'll be interesting to get feedback - if there's any way to do that in advance of a new release would be great to get some users trying this out.

I can add a setting to enable this feature, have it off by default, get feedback from users who specifically asked for this feature and anyone else you have in mind. Then, in the next release we make it a default feature.

>If I already have imported the name, I now see it twice, once based on the actual name in scope and once as a suggestion for an Add Import.

#508 and https://github.com/lukehoban/go-outline/pull/4 should take care of this. Can you take a look?

ramya-rao-a:
@lukehoban 
A Setting `go.autocomplteUnimportedPackages` has been added. Its false by default. This will control the Autocomplete Unimported Packages feature

#508 has been merged, so no more double entries for the same package

PR #498, opened 2016-09-22
Set environment variables for go test runs

### LINKED ISSUE #475: How to set env var before run Go test command
hi, Need use env var in go test , So  I want  to add Environment Variables just for go test command.
how to do it ?  thanks!
`Go: Run test at cursor`

ramya-rao-a:
Fixes #475


src/goTest.ts
```diff
@@ -94,19 +94,19 @@ export function testCurrentPackage(timeout: string) {
...
 		return;
 	}
-	getTestFunctions(editor.document).then(testFunctions => {
+	return getTestFunctions(editor.document).then(testFunctions => {
 		return goTest({
 			timeout: timeout,
 			dir: path.dirname(editor.document.fileName),
 			functions: testFunctions.map(func => { return func.name; })
-		});
-	}).then(null, err => {
```
lukehoban: Possibly minor - but this change will mean that any exceptions thrown during execution of `goTest` are not handled and reported to `console.error`.  A weirdness of `.then` on promises.  Possibly just leave the additional `.then` call.

src/goTest.ts
```diff
@@ -116,19 +116,23 @@ export function testCurrentFile(timeout: string) {
  *
  * @param config the test execution configuration.
  */
-function goTest(config: TestConfig): Thenable<boolean> {
+function goTest(config: TestConfig, goConfig?: vscode.WorkspaceConfiguration): Thenable<boolean> {
```
lukehoban: Any reason not to make this required and fix up callers?

src/goTest.ts
```diff
@@ -116,19 +116,23 @@ export function testCurrentFile(timeout: string) {
...
 		outputChannel.show(2);
-		let buildFlags: string[] = vscode.workspace.getConfiguration('go')['buildFlags'];
-		let buildTags: string = vscode.workspace.getConfiguration('go')['buildTags'];
+		if (!goConfig) {
+			goConfig = vscode.workspace.getConfiguration('go');
+		}
+		let buildFlags: string[] = goConfig['buildFlags'];
+		let buildTags: string = goConfig['buildTags'];
 		let args = ['test', '-v', '-timeout', config.timeout, '-tags', buildTags, ...buildFlags];
+		let testEnvVars = Object.assign({}, goConfig['testEnvVars'], process.env);
```
lukehoban: Doesn't the order need to be changed to ensure that the `testEnvVars` overrride the process environment settings?

src/goTest.ts
ramya-rao-a: Put the `console.error(err)` to back where it was and returned promise from there

src/goTest.ts
ramya-rao-a: None at all. Made the update to pass goConfig from goMain.ts

src/goTest.ts
ramya-rao-a: True, nice catch. Corrected

rynop:
For those trying to figure out how to use this, you have to set a value in your vs code `settings.json`:

Ex:
`"go.testEnvFile": "${workspaceFolder}/.env",`

henrahmagix:
Or
```json
"go.testEnvVars": {
  "APP_DEBUG": "1",
  "MY_ENV": "my-value"
},
```

PR #499, opened 2016-09-23
Send test output to a distinct output channel

ironcladlou:
Send Go test output to a dedicated output channel. Re-using the "Go"
channel from the status component causes a lot of confusing
interspersed unrelated output, and the test command often clears the
"Go" channel, obliterating things users might want to see.


ramya-rao-a:
@ironcladlou Just a heads up, travis builds most likely will fail due to https://github.com/Microsoft/vscode-debugadapter-node/issues/65

lukehoban:
Looks good.

Issue #374, opened 2016-07-05
Intentional choice to not collapse imports?


benclarkwood:
I love the auto import features of vscode-go, but it always bugs me that I can wind up in scenarios where the imports aren't collapsed into a 

```
import (
  // first
  // second
  // etc
)
```

statement instead. I'd like to change goImport.ts to handle this collapsing. I can wrap it in a setting if that is more amenable.

ramya-rao-a:
@benclarkwood 
In a go file with no imports, I added `fmt.Println("Hello")` and saved to trigger the auto import which gave me
```go
import "fmt"
```
Then I added `math.Abs(1)` and saved the file which updated the import block to
```go
import (
   "fmt"
   "math"
)
```

The auto imports do get collapsed. 
Is your scenario any different?

peterbourgon:
@ramya-rao-a While it does work like that _somtimes_, I can easily get vscode-go into a state where subsequent imports are added one-by-one, i.e.

```go
import "fmt"
import "math"
import "os"
```

IMO there's no reason not to use the import grouping form all the time. That is, even when I just have a single import, I want it to look like

```go
import (
    "fmt"
)
```

Perhaps we can change it to simply have this behavior?

benclarkwood:
@ramya-rao-a sorry I didn't include proper repro steps before. Here is a repro that always causes the failure to collapse for me:

1. Create a file (e.g. importsbreak.go) with an import:
```
package main

func test() {
	fmt.Println("test")
}
```
2. Save and close the file (the auto-import will add `fmt`
3. Reopen the file, use the command palette Go: Add Import command to add another import (e.g. net/http)
4. I at least end up with something like this then:
```
package main

import "fmt"
import "net/http"

func test() {
	fmt.Println("test")
	http.NewRequest("GET", "banana", nil)

}
```

As @peterbourgon suggested, the "simplest" (and maybe best?) solution is to just always use the ```import (
  ...
)
```
block style for imports. I'll take a look at what has to change in goImport.ts for that and submit a PR against this ticket.

ramya-rao-a:
@benclarkwood  @peterbourgon  Sorry, I saw the word "goImport" and my brain mapped it to "goimports" the formatting tool and thought the problem was with the imports that get added when formatting is triggered.

At the moment, any new import added via the `Go: Add Import` command gets added inside `()` if existing imports on the page are already inside a `()` or if there are no imports on the existing page. Else, they get added as separate `import` statements.

[...]

PR #500, opened 2016-09-23
Proposal to implement collapsing

benclarkwood:
1. Add new setting "go.blockImportsOnly"
2. Add parsing of package name for single import lines
3. When the user adds an import manually and there is already one or more single imports, collapse it all into a block if go.blockImportsOnly is true.

note: addresses #374

### DIFF
--- a/package.json
+++ b/package.json
@@ -900,6 +900,12 @@
           "default": [],
           "description": "Folder names (not paths) to ignore while using Go to Symbol in Workspace feature",
           "scope": "resource"
+        },
+        "go.blockImportsOnly": {
+          "type": "boolean",
+          "default": false,
+          "description": "If true, all imports in a file will always be in the block import ( ... ) style.",
+          "scope": "resource"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -900,12 +900,6 @@
           "default": [],
           "description": "Folder names (not paths) to ignore while using Go to Symbol in Workspace feature",
           "scope": "resource"
-        },
-        "go.blockImportsOnly": {
-          "type": "boolean",
-          "default": false,
-          "description": "If true, all imports in a file will always be in the block import ( ... ) style.",
-          "scope": "resource"
         }
       }
     },
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -11,7 +11,6 @@ import { documentSymbols } from './goOutline';
 import { promptForMissingTool } from './goInstallTools';
 import { getImportablePackages } from './goPackages';
 
-
 const missingToolMsg = 'Missing tool: ';
 
 export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
@@ -60,7 +59,6 @@ export function getTextEditForAddImport(arg: string): vscode.TextEdit[] {
 		return null;
 	}
 
-	const goConfig = vscode.workspace.getConfiguration('go');
 	let { imports, pkg } = parseFilePrelude(vscode.window.activeTextEditor.document.getText());
 	let multis = imports.filter(x => x.kind === 'multi');
 	if (multis.length > 0) {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -320,6 +320,11 @@
           "type": "string",
           "default": "",
           "description": "The Go build tags to use for all commands that support a `-tags '...'` argument"
+        },
+        "go.blockImportsOnly": {
```
benclarkwood: If anyone has a better name...

src/util.ts
```diff
@@ -31,10 +31,12 @@ export function parseFilePrelude(text: string): Prelude {
 			ret.pkg = { start: i, end: i };
 		}
 		if (line.match(/^(\s)*import(\s)+\(/)) {
-			ret.imports.push({ kind: 'multi', start: i, end: -1 });
+			ret.imports.push({ kind: 'multi', start: i, end: -1, package: 'multi-na' });
 		}
 		if (line.match(/^(\s)*import(\s)+[^\(]/)) {
-			ret.imports.push({ kind: 'single', start: i, end: i });
+			// Pull the package name
+			let match = line.match(/import "(.+)"/);
```
lukehoban: Need to make sure this never fails.  Probably better to just modify the `if` match to capture the name as well?  I ran into failures when a line looked like `import  "byte"` for example (two spaces instead of one).

package.json
lukehoban: I actually think we can just make this the default.  No need for a new setting here.

src/goImport.ts
```diff
@@ -47,12 +47,34 @@ export function addImport(arg: string) {
...
+				importBlock += '\t"' + element.package + '"\n';
+			});
+			importBlock += '\t"' + imp + '"\n'; // Add the user's new import
+			importBlock += ')\n';
+
+			return vscode.window.activeTextEditor.edit(editBuilder => {
+				// First, delete all the single packages.
+				editBuilder.delete(new vscode.Range(new vscode.Position(firstImport.start, 0), new vscode.Position(lastImport.end, 10000)));
+				// Now, insert the import block.
+				editBuilder.insert(new vscode.Position(pkg.start + 2, 0), importBlock);
```
lukehoban: Is anchoring this on `pkg.start` always safe?  What if there are comments in front of the imports or two groups of parenthesized imports?  Can this just be anchored at `firstImport.start` instead?

src/util.ts
ramya-rao-a: This will also fail in case of dot imports and named imports. Example:
```
import . "fmt"
import m "math"
```

src/goImport.ts
ramya-rao-a: Like @lukehoban  says, this would result in comments that were previously between packages and imports to appear after the new import block.
Also, comments in between single import statements are lost too. Example:
```
import "math"  //Helps me in math
// This is my custom package
import "mycustom"
```
How about doing this instead?
1. Insert `(\n\t` after the word "import" in the first import
2. Replace the word "import" with `\t` for rest of the imports
3. Insert `)` in the line after the last import

benclarkwood:
Oh my gosh this is such a better way of doing this!

Thanks!

src/goImport.ts
```diff
@@ -31,7 +31,7 @@ function askUserForImport(): Thenable<string> {
 	});
 }
 
-export function addImport(arg: string) {
+export function addImport(arg: string, goConfig: vscode.WorkspaceConfiguration) {
```
ramya-rao-a: you shouldn't be needing `goConfig` here anymore

src/goImport.ts
```diff
@@ -48,10 +48,46 @@ export function addImport(arg: string) {
...
+				5. The user then saves, and goimports will clean everything up
+				*/
+				let firstImport = imports[0];
+				let lastImport = imports[imports.length - 1];
+				const importLength = 'import'.length;
+
+				editBuilder.insert(new vscode.Position(firstImport.start, 0), 'import (\n');
+
+				imports.forEach(element => {
+					editBuilder.replace(new vscode.Range(new vscode.Position(element.start, 0), new vscode.Position(element.start, importLength)), '\t');
```
ramya-rao-a: Depending on `importLength` will cause issues when there are spaces/tabs before the word "import".
This shouldn't happen if formatting has been run on the file, but there might be cases where user has not yet saved the file or has turned formatOnSave off or has not formatted the doc at all

src/goImport.ts
```diff
@@ -48,10 +48,46 @@ export function addImport(arg: string) {
...
+			// then later the user manually imports something via this command.
 			return vscode.window.activeTextEditor.edit(editBuilder => {
-				editBuilder.insert(new vscode.Position(lastSingleImport + 1, 0), 'import "' + imp + '"\n');
+				/* The import statement(s) probably look something like this right now:
+
+				import "apackage"
+				// A comment
+				import . "dotinclude"
+				import "apackage2" // A comment explaining the package
+
```
ramya-rao-a: Can you add a unit test for this?

benclarkwood:
Haven't abandoned this, just been supes busy at work. Will add tests / address additional concerns this weekend hopefully.

ramya-rao-a:
@benclarkwood There has been some changes to `goImport.ts` recently, so make sure to get latest from `master` before you continue

mattetti:
This PR is stuck in a conflict, @benclarkwood would you mine rebasing?
On a separate note, I find the property name `go.blockImportsOnly` quite confusing since it reads like we are blocking imports only. Maybe `go.groupedImportsOnly`?

ramya-rao-a:
@benclarkwood I took the liberty of rebasing.

benclarkwood:
@ramya-rao-a thanks! Sorry, Golang became not my daily driver and I totally forgot about this PR. 🙇 .

Issue #467, opened 2016-09-02
Tool installation is broken


calmh:
While setting up a new installation, `lukehoban.Go-0.6.43`:

![screen shot 2016-09-02 at 07 10 03](https://cloud.githubusercontent.com/assets/125426/18193311/8a5dab14-70dc-11e6-8479-27528f4685ae.png)

For searchability:

`The "goimports" command is not available. Use "go get -v undefined" to install.`

("undefined" is not a good package path here.)

Obviously "Install" does nothing, and also not "Install all".

ramya-rao-a:
This fix is now available in the latest version of the Go extension : 0.6.44

PR #509, opened 2016-10-02
Update installTools logic to support optional tools

ramya-rao-a:
The previous fix to #467 added `goimports` to the list of tools to be installed.
For users who have not chosen `goimports` as their formatting tool, this change will prompt them to install `go imports` via the `Analysis Tools Missing` status bar item.

The current PR makes it so that user will be prompted to install `goimports` only if it has been chosen  as their formatting tool

`golint` has dropped support for Go 1.5 This would break installing of the `golint` tool on machines having Go 1.5 This also breaks travis

Ideally, `golint` should provide a tag/branch which works for Go 1.5 and we should be able to use that.
Until then, we suggest such users to use `gometalinter` instead


src/goInstallTools.ts
```diff
@@ -11,38 +11,67 @@ import path = require('path');
...
+		'gopkgs': 'github.com/tpng/gopkgs',
+		'godef': 'github.com/rogpeppe/godef',
+		'go-outline': 'github.com/lukehoban/go-outline',
+		'go-symbols': 'github.com/newhook/go-symbols',
+		'guru': 'golang.org/x/tools/cmd/guru',
+		'gorename': 'golang.org/x/tools/cmd/gorename'
+	};
+
+	// Install the formattool that was chosen by the user
+	if (goConfig['formatTool'] === 'goimports') {
```
Tyriar: Is `gofmt` included with Go?

src/goInstallTools.ts
```diff
@@ -11,38 +11,67 @@ import path = require('path');
...
+	// golint is no longer supported in go1.5
+	if (goVersion && (goVersion.major > 1 || (goVersion.major === 1 && goVersion.minor > 5))) {
+		tools['golint'] = 'github.com/golang/lint/golint';
+	}
+	return tools;
+}
 
 export function installAllTools() {
-	installTools(Object.keys(tools));
+	getGoVersion().then(() => installTools(Object.keys(getTools())));
```
Tyriar: `getTools` is called here and then again multiple times in `installTools`

Tyriar: Also you're not using the promise in `then`, why not do this?

```ts
getGoVersion().then(version => installTools(Object.keys(getTools()), version));
```

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.5
```
Tyriar: Can we amend the tests to not lint on 1.5 instead?

src/goInstallTools.ts
ramya-rao-a: Yes

.travis.yml
ramya-rao-a: Its not the tests that fail, its the `install` step where we install `golint`.
I could write a custom install script that looks at the go version and then takes the decision whether or not to install `golint`
But, thats for another day.
I am hoping that `golint` craetes a tag/branch that works for Go 1.5, then I can put 1.5 back here in travis, add a custom install script to target that tag/branch

Tyriar: I'm concerned about removing the 1.5 test and then never adding it back again. I think it's worth adding a conditional to https://github.com/Microsoft/vscode-go/blob/master/.travis.yml if that's all that needs to happen.

Tyriar: Does this work?

```bash
if [ "$(go --version)" != "1.5" ]; then go get -u -v github.com/golang/lint/golint; fi
```

src/goInstallTools.ts
ramya-rao-a: The version is used internally by `getTools()` and not `installTools` directly.

Refactored `installedTools` to not repeatedly call `getTools()` for each missing tool

.travis.yml
ramya-rao-a: am trying this
```bash
if [[ "$(go version)" =~ "go version go1.5" ]]; then echo cannot get golint; else go get -u -v github.com/golang/lint/golint; fi
```

ramya-rao-a:
@Tyriar All comments have been addressed, I think we are good to go

PR #510, opened 2016-10-04
deprecates Go 1.5 in .travis.yml

cedriclam:
This Extension is using golint (github.com/golang/lint/golint) that
requires at least Go 1.6.
It results as an error during the CI install step in travis-ci for the
target envs go.1.5 linux and osx.
error: ../../golang/lint/lint.go:250: undefined: types.ImportMode

modifications done:
- Remove envs "Go 1.5".
- Update the Readme in order to inform users that this Extension
  requires at least Go 1.6.


ramya-rao-a:
we don't need to drop support for Go 1.5 just because `golint` did so. Existing users of the extension on Go 1.5 can continue to use the extension without any issues. It is only on machines where `golint` has not yet been installed and uses Go 1.5, that users will have problems. And of course travis. I'll send out a change soon to redirect such users to use `gometalinter` instead and remove `golint` from the install steps in travis and move it to the unit test that tests listing

PR #518, opened 2016-10-09
minor fix hover doc unexcept line wrapping!

leaxoy:
Due to go source doc have some newline. It's may be conflict with the auto wrap in vscode hover action. So replace `\n` with ` ` to fix this.
If source doc is 
```
This code is a balblabla value. And
other
```
Before:
```
This code is a balblabla
value. And 
other
```
After:
```
This code is a balblabla
value. And other
```

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -49,7 +49,7 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 				let pkgPath = path.dirname(file);
 				let definitionInformation: GoDefinitionInformtation = {
 					file: file,
-					line: +line - 1,
+					line: + line - 1,
 					col: + col - 1,
 					lines,
 					doc: undefined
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@leaxoy I tried the below 2 cases

```go
// Line 1 comment
// Line 2 comment
func main() {
}
```

```go
/*Line 1 comment
 Line 2 comment*/
func main() {
}
```

In both cases, I don't see any issues with the hover info.
The hover info appears in a single line and wraps if too long.
I am using VS Code 1.6.1 in Mac OS X

Can you elaborate as to what exactly are you trying to fix?

ramya-rao-a:
@leaxoy As mentioned in https://github.com/Microsoft/vscode-go/pull/518#issuecomment-254024679, I don't get what is being fixed here. 
I am closing this PR, if you still see a problem, create an issue and we can take it from there

PR #535, opened 2016-10-16
Fix signature / hover provider (e.g. for struct fields)

### LINKED ISSUE #440: Hover info does not show doc string for structs
I'm not sure if this is expected behaviour, but here is a screenshot. The struct has a docstring but it is not shown in the hover info.

nochso:
This fixes #440.

`guru` is used to find the position of any definition instead of `godef` as it's more reliable.
The definition lines are read straight from source.

Also `godoc` is no longer used to fetch the documentation, because querying it did not work in all cases.

This is my first time working with TypeScript, so feel free to nitpick. I've tested all I could think of and it consistently provides useful hover information to me.

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -52,9 +52,10 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 					lines: [sig],
 					doc: undefined
 				};
+				let source = fs.readFileSync(definitionInformation.file, 'utf8');
+				lines = source.split('\n');
+				addLinesToDefinition(definitionInformation, lines);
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@nochso Thanks for the PR! Sorry it took so long to review this.

I tried out your changes and though it works for the hover scenario, other features like Goto Definition, Peek Definition and Signature Help are broken.

What your thoughts on #607 ?

ramya-rao-a:
@nochso #622 was used to fix #440 and 5 other issues.
Therefore closing this PR.
Thanks for your PR, hope to see more from you and happy coding!

PR #550, opened 2016-10-23
Use -imports-only option from go-outline

ramya-rao-a:
We use `go-outline` tool to get list of imported packages in the current file. `go-outline` internally parses the file to return all symbols. But since we need only imports, the whole file need not be parsed. https://github.com/lukehoban/go-outline/pull/4 added the option to parse only the initial part of the file that contains the imports.
This PR uses this option.

### DIFF
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -9,7 +9,7 @@ import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
 import { getBinPath } from './goPath';
-import { promptForMissingTool } from './goInstallTools';
+import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 
 // Keep in sync with https://github.com/lukehoban/go-outline
 interface GoOutlineRange {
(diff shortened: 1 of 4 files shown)


src/goOutline.ts
```diff
@@ -29,15 +29,31 @@ interface GoOutlineDeclaration {
...
-export function documentSymbols(filename: string): Promise<GoOutlineDeclaration[]> {
+interface GoOutlineOptions {
+	fileName: string;
+	importsOnly?: boolean;
+}
+
+export function documentSymbols(options: GoOutlineOptions): Promise<GoOutlineDeclaration[]> {
 	return new Promise<GoOutlineDeclaration[]>((resolve, reject) => {
 		let gooutline = getBinPath('go-outline');
+		let gooutlineFlags = ['-f', options.fileName];
```
roblourens: camelCase?

test/go.test.ts
```diff
@@ -97,6 +97,9 @@ encountered.
 					})
 				);
 				return Promise.all(promises);
+			}).then(() => {
+				vscode.commands.executeCommand('workbench.action.closeActiveEditor');
```
roblourens: executeCommand returns a promise - don't know if it matters though

PR #553, opened 2016-10-24
Select a random port to debug on

### LINKED ISSUE #552: Can't use two instances of the debugger at the same time
Because the port is hardcoded at '2345', unless otherwise specified.

roblourens:
Fix #552.

This is copied from node-debug/node-debug2. It doesn't check whether a port is already in use, but... there are lots of ports out there. Without this, debugging two Go processes with a 'composite' launch config won't work out of the box without specifying ports manually.


roblourens:
The failures look unrelated

ramya-rao-a:
About the failures...

The ones for Go 1.5 is because `gometalinter` just updated to use the latest `golint` which dropped support for Go1.5 I am looking into it and in conversation with `gometalinter` owner

The others are just linting errors, should be easy to fix

roblourens:
Didn't notice the lint, fixed that.

Issue #540, opened 2016-10-18
Adopt formatOnSave changes


jrieken:
Since VS Code now has the `editor.formatOnSave`-setting and the `onWillSaveTextDocument`-event, [this code](https://github.com/Microsoft/vscode-go/blob/master/src/goMain.ts#L208) should be revisited. I'd recommend to remove it and to rely on the built-in `formatOnSave` feature or to use the new event. Both will prevent unnecessary fs-operations and reduce the risk for endless write-event-write loops. 

See https://code.visualstudio.com/updates#_format-on-save, Microsoft/vscode#12449, and Microsoft/vscode#239 for more details

ksmithut:
Can that `editor.formatOnSave` be applied on a per-language basis? I would love that setting for `go`, but not for `javascript`

ramya-rao-a:
@jrieken In VS Code 1.6 only formatters implementing `DocumentRangeFormattingEditProvider` get to use the `editor.formatOnSave`. The Go extension formatter implements `DocumentFormattingEditProvider` which I see that we are working on for VS Code 1.7

Can you confirm?

jrieken:
Yeah - I have fixed a bug for the Oct release that makes sure we honour Document and DocumentRange formatters.

ramya-rao-a:
great, will wait for the next release then

thnt:
I think you should keep `go.formatOnSave` until Microsoft/vscode#13547 has been resolved

dbudworth:
agreed on it needing to be a language specific thing.
All go editors format on save as it's expected by the community that you have ran your code through "go fmt"

that is not expected by folks in other languages in general.  If I formatted javascript on save, i'd have a lot of really annoyed co-workers.

As it stands, I basically can't use vscode as my go editor as I'm so used to format-on-save that I don't even think of formatting.

And, being a child of the DOS era, I habitually save as I move around looking at files.  So having javascript code that I have no intention of changing get reformatted as a side effect of 25 year old muscle memory is kind of a non-starter.

ramya-rao-a:
Current State: The adoption of `editor.formatOnSave` was reverted in 0.6.49

jrieken:
why?

ramya-rao-a:
@jrieken `editor.formatOnSave` saves all files on save. There has been feedback from the community that they prefer formatting just go files on save rather than all files.
[...]

PR #578, opened 2016-10-28
Support editor.formatOnSave

ramya-rao-a:
#540

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "test": "node ./node_modules/vscode/bin/test",
     "lint": "node ./node_modules/tslint/bin/tslint ./src/*.ts ./src/debugAdapter/*.ts ./test/*.ts"
   },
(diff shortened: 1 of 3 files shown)


aphistic:
This is really inconvenient.  There are only certain languages I would want auto formatting for and I don't want to turn it on completely just to have my Go code auto-formatted.

chrisolsen:
The default value of `editor.formaOnSave` is `false` which overrides `go.formatOnSave` making things seem broken after getting the most recent update.

ramya-rao-a:
This change has been reverted due to popular demand in the latest update. Will wait for VS Code to come up with language specific formatOnSave setting

aphistic:
Thank you!

ramya-rao-a:
@aphistic No problem.
Thanks for reporting issues and pain points. I get to learn something new each day :)

Issue #566, opened 2016-10-26
Generating unit tests should open or link to generated test file


mousetraps:
[Smoke testing](https://github.com/Microsoft/vscode-go/wiki/Smoke-Test), Windows

Right now, the test file is generated, but then the user has to figure out where, exactly, they got dropped. There should either be a button in the infobar that opens the generated test file, or the file should open automatically in the editor.

ramya-rao-a:
Also, the message that is shown at the end of generating tests can be better.
Example:

- Say 2 out of 3 functions in the file `hello.go` already have the tests in `hello_test.go`, then the message should only mention the test for the 3rd function
- Say all functions in the file `hello.go` have the tests in `hello_test.go`, then the message should say that no new tests were created.

etc.

PR #584, opened 2016-10-30
Dont generate unit tests for non Go files

### LINKED ISSUE #563: Generating unit tests should only apply to go files
[Smoke testing](https://github.com/Microsoft/vscode-go/wiki/Smoke-Test), Windows

In this case, no test is actually generated, but a status bar messages appears claiming successful test generation. 
image

ramya-rao-a:
Fixes #563

### DIFF
--- a/src/goGenerateTests.ts
+++ b/src/goGenerateTests.ts
@@ -14,32 +14,45 @@ import { getBinPath } from './goPath';
 import { promptForMissingTool } from './goInstallTools';
 import { GoDocumentSymbolProvider } from './goOutline';
 
-export function generateTestCurrentPackage(): Thenable<boolean>  {
+/**
+ * If current active editor has a Go file, returns the editor.
+ */
+function checkActiveEditor(): vscode.TextEditor {
 	let editor = vscode.window.activeTextEditor;
 	if (!editor) {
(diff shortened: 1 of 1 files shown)


src/goGenerateTests.ts
```diff
@@ -14,32 +14,45 @@ import { getBinPath } from './goPath';
...
+ */
+function checkActiveEditor(): vscode.TextEditor {
 	let editor = vscode.window.activeTextEditor;
 	if (!editor) {
-		vscode.window.showInformationMessage('gotests: No editor selected');
+		vscode.window.showInformationMessage('Cannot generate unit tests. No editor selected.');
+		return;
+	}
+	if (!editor.document.fileName.endsWith('.go')) {
+		vscode.window.showInformationMessage('Cannot generate unit tests. File in the editor is not a Go file.');
```
roblourens: Is there definitely no case where a go file can have some other extension?

ramya-rao-a: @roblourens Not that I know of

ramya-rao-a:
@roblourens #566 is tracking your https://github.com/Microsoft/vscode-go/pull/584#pullrequestreview-6531416

Issue #432, opened 2016-08-16
feature request: switch to test file


mattetti:
Like probably a lot of you, I switch back and forth between my tests and my implementation.
I'd like to be able to:
- jump to the test file (current file name but with `_test` suffix)
- jump to the test implementation (based on name conventions)
- open test file in other pane (so I can have both files side to side)
- jump to implementation file when looking at a test file

Bonus:
- create a test file if missing
- generate stubbed test for a function/method (generic table test)

I have to find out where is the documentation to open files and to define targets, but I think that should be quite doable.

mattetti:
API to open a specific file using filename only: https://code.visualstudio.com/docs/extensionAPI/vscode-api#1108

ramya-rao-a:
@mattetti The latest update has these features:

Commands to switch to test file and back. You can add your own shortcuts for these commands.
        - `Go: Open Test File`
        -  `Go: Open Implementation for Test File`
Navigate to test file after generating unit tests using the `Go: Generate unit tests ...` commands that were added last month

This should cover most of your asks here :)

Happy Coding!

PR #590, opened 2016-11-01
Open Test Files

ramya-rao-a:
Implementing Feature Requests from #432 and #566

### DIFF
--- a/src/goGenerateTests.ts
+++ b/src/goGenerateTests.ts
@@ -14,6 +14,9 @@ import { getBinPath } from './goPath';
 import { promptForMissingTool } from './goInstallTools';
 import { GoDocumentSymbolProvider } from './goOutline';
 
+const generatedWord = 'Generated ';
+const commandTitle = 'Open generated test for current file';
+
 /**
  * If current active editor has a Go file, returns the editor.
  */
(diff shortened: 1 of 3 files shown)


src/goGenerateTests.ts
```diff
@@ -112,8 +152,29 @@ function generateTests(conf: Config): Thenable<boolean> {
 				if (err) {
 					return reject('Cannot generate test due to errors: ' + stderr);
 				}
-				let message = 'gotests: ' + conf.msg;
-				vscode.window.showInformationMessage(message);
+
+				let message = stdout;
+				let testsGenerated = false;
+				if (stdout.startsWith(generatedWord)) {
```
roblourens: A comment about the output you expect and what you want to transform it to will make this easier to debug if it changes some day.

src/goGenerateTests.ts
```diff
@@ -112,8 +152,29 @@ function generateTests(conf: Config): Thenable<boolean> {
...
+					}).map((element) => {
+						return element.substr(generatedWord.length);
+					});
+					message = `Generated ${lines.join(', ')}`;
+					testsGenerated = true;
+				}
+
+				if (testsGenerated) {
+					vscode.window.showInformationMessage(message, 'Open Generated Test for current file').then(selected => {
+						if (selected === 'Open Generated Test for current file') {
```
roblourens: Extract this string? And i'm not sure 'Generated Test' should be capitalized.

ramya-rao-a: will do

kwiesmueller:
As I keep being annoyed by this feature in my daily work I was looking for a setting to disable it. As there is none I want to ask here.
When I work with tests in go, most of the time I got my editor in split. Left is code, right is test.
When then using the add test feature, my code gets closed and the test file (i already have open) gets opened there.
This is gets kinda annoying when generating a lot of tests.
My suggestion would be a setting for this to enable or disable it.
As an addition to that, I would also appreciate a setting to disable the "test created" modal, which blocks my view every time even if I don't care about.

What do you think?

PR #594, opened 2016-11-01
Fixes #592 Use go from GOROOT while installing tools

### LINKED ISSUE #592: Install tools does not use go.goroot project setting.
When I setup a go.goroot project setting I expect that the install tools would use the go command installed in that location, and not try to find it from the $PATH environment variable.

I have a setup where I want different tooling and different versions of the go environment for different projects and they should all be self contained. 

Output:

```
Installing 1 tools
  goreturns

Installing goreturns FAILED

1 tools failed to install.

goreturns:
Error: Command failed: go get -u -v sourcegraph.com/sqs/goreturns
/bin/sh: go: command not found
/bin/sh: go: command not found
```

ramya-rao-a:
We have been using Go from GOROOT (falling back to PATH if not found) everywhere while running the `go` command except while installing tools.
This PR will fix that
Fixes #592

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -11,7 +11,7 @@ import path = require('path');
 import os = require('os');
 import cp = require('child_process');
 import { showGoStatus, hideGoStatus } from './goStatus';
-import { getBinPath } from './goPath';
+import { getBinPath, getGoRuntimePath } from './goPath';
 import { outputChannel } from './goStatus';
 import { getGoVersion, SemVersion, isVendorSupported } from './util';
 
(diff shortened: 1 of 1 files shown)


src/goInstallTools.ts
```diff
@@ -110,7 +110,8 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 
 	missing.reduce((res: Promise<string[]>, tool: string) => {
 		return res.then(sofar => new Promise<string[]>((resolve, reject) => {
-			cp.exec('go get -u -v ' + tools[tool], { env: process.env }, (err, stdout, stderr) => {
+			let goRuntimePath = getGoRuntimePath();
```
tw1nk: You would probably only need to lookup the goRuntimePath once during the install phase so to improve performance move it above the reduce function.

Issue #225, opened 2016-03-04
package main completion issue


freeformz:
Main a new file, start typing "package main" + Enter and you get this:

![sadmain](https://cloud.githubusercontent.com/assets/103/13542235/3ba3dbe2-e216-11e5-9949-95676ca84eaf.gif)

lukehoban:
Do you have the latest version of the extension installed?  There was a fix yesterday which should have made this less likely to happen.

freeformz:
I updated to the latest released version, restarted vscode and the same thing happens.

lukehoban:
Yeah - the issue is the `main` snippet - which is handy when you want to use it, but can get in the way when you don't.  Not sure what to do about that without removing the snippet.

The one thing that did change recently is that you should get a completion suggestion as you start typing `package` which will offer to complete `package main\n` for you.  If you select that before you get to typing the `main` part, it will autocomplete what you want.

But that doesn't fully solve the issue.

I think the options are:
* Leave this behaviour as is.
* Remove all snippets.
* Rename the `main` snippet to something less likely to interfere with normal typing (like `im` and `ims` which are used for imports).

I don't love any of the options though.

freeformz:
I find the existing auto-completion/snippet for `package` to be off. It suggests the name (seemingly anyway) based on the directory you are in. IMO it should look at other files in the same directory and default to the same package, unless it's a _test.go file where it could look at other test files in the same directory and default to the same package name. For what I do anyway, if there are no other go files it should default to `package main`. But maybe I'm off and my work flow isn't the same as other people's?

freeformz:
Is there any way to squash the auto-complete if the line starts with `package`? This way when I type `package main` I never get the main function auto-complete?

patbaker82:
This is a problem in the atom editor as well.  It would be nice if we could stop the evaluation of the snippet.

You could also:

- Remove main as a snippet prefix
- Create snippet prefix `pkgm` that will create both the package main designation and the main function

lukehoban:
@patbaker82  I like that suggestion - it replaces the `main` snippet with something more useful when you want to create a `main` file, plus avoids the name clash.

PR #602, opened 2016-11-03
add a few useful snippets

marklap:
Useful snippets to me that is.

### DIFF
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -155,6 +155,18 @@
 		"test function": {
 			"prefix": "tf",
 			"body": "func Test$1(t *testing.T) {\n\t$0\n}"
+		},
+		"init function": {
+			"prefix": "finit",
+			"body": "func init() {\n\t$1\n}"
+		},
+		"main function": {
+			"prefix": "fmain",
+			"body": "func main() {\n\t$1\n}"
+		},
+		"method declaration": {
+			"prefix": "meth",
+			"body": "func (${1:var} ${2:receiver}) ${3:method}($4) $5 {\n\t$0\n}"
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


snippets/go.json
```diff
@@ -155,6 +155,18 @@
 		"test function": {
 			"prefix": "tf",
 			"body": "func Test$1(t *testing.T) {\n\t$0\n}"
+		},
+		"init function": {
+			"prefix": "finit",
```
kulshekhar: All things considered, `fmain` and `init` seem like the most apt options

marklap: `fmain` and `init` work for me.

marklap: Someone give me the nod and I'll update my PR to change `finit` to `init`.

bennyscetbun: Hi my thought on this one.
I like consistency and many dev does too.
main/init should be good except that the #225 is a real pain when using it
So fmain/finit works fine for me :)
But please. No fmain/init or main/finit :)

patbaker82: I agree with @bennyscetbun.

ramya-rao-a: Consistency does make sense and drive usability as well. Lets see if there is any other feedback over the weekend. I'll merge the PR (`fmain` and `finit`) on Monday if we don't hear anything more.

kulshekhar: Given that 'test function' uses `tf` and 'goroutine function' uses `gf`, maybe the main function can use `mf` while the 'init' function can use `init` (`if` might not be the best choice :slightly_smiling_face:)

ramya-rao-a: sigh.. This takes me back to a programmer's hardest task. http://imgur.com/y3dhJFQ : Naming things.

Going with `fmain` and `finit`

PR #603, opened 2016-11-04
pkg suggestions to appear after gocode suggestions

ramya-rao-a:
Ever since we introduced importable packages as suggestions in the auto-complete, people have found it annoying that these package suggestions are interspersed among variables/function names etc.

Excerpt from Gitter:

> After last plugin update in autocompletion list I got list of all available packages, it's pretty cool but little annoying because variables with corresponding names appears in the end of a list. It would be nice to sort this list and show variables on top

This PR is to group all the package suggestions and make them appear AFTER the usual/normal suggestions from `gocode`.

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -184,6 +184,8 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 								}
 								item.insertText = suggest.name + '(' + paramSnippets.join(', ') + ') {{}}';
 							}
+							// Add same sortText to all suggestions from gocode so that they appear before the unimported packages
+							item.sortText = 'a';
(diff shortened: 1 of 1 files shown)


src/goSuggest.ts
```diff
@@ -184,6 +184,8 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 								}
 								item.insertText = suggest.name + '(' + paramSnippets.join(', ') + ') {{}}';
 							}
+							// Add same sortText to all suggestions from gocode so that they appear before the unimported packages
```
roblourens: Will they still be in the right order within the 'a' and 'z' groups, or should it be like `'a' + item.name`?

PR #610, opened 2016-11-09
Open test file after the Generate Unit Tests command

ramya-rao-a:


kwiesmueller:
As I keep being annoyed by this feature in my daily work I was looking for a setting to disable it. As there is none I want to ask here.
When I work with tests in go, most of the time I got my editor in split. Left is code, right is test.
When then using the add test feature, my code gets closed and the test file (i already have open) gets opened there.
This is gets kinda annoying when generating a lot of tests.
My suggestion would be a setting for this to enable or disable it.
As an addition to that, I would also appreciate a setting to disable the "test created" modal, which blocks my view every time even if I don't care about.

What do you think?

ramya-rao-a:
For the opening of test file after generating tests, we can add a check so that it doesn't happen if the test file is already open. 
For the notification, instead of a setting we can have a "Don't show again" option. We have too many settings already in the extension and I'd like to avoid adding a new one.

PR for both these improvements are welcome

kwiesmueller:
Sounds good to me.
I already disabled the opening of tests locally, so I might as well do this as a pr with your suggestion.

PR #614, opened 2016-11-11
always pass in the -srcdir option when using goimports

abduld:


ramya-rao-a:
@abduld Can you explain the problem you are trying to solve?

Issue #567, opened 2016-10-26
Signature Help and Quick Info do not show function comments for unexported functions 


mousetraps:
[Smoke testing](https://github.com/Microsoft/vscode-go/wiki/Smoke-Test), Windows

In the example below, I would expect the "hello" comment to appear in addition to just the function signature.
![image](https://cloud.githubusercontent.com/assets/762848/19744622/23bd4dc0-9b83-11e6-8e5f-747c50b8b90c.png)

This works fine for cases like stringutil.Reverse
![image](https://cloud.githubusercontent.com/assets/762848/19744722/913ae61e-9b83-11e6-8c63-8f469f03fa9a.png)

ramya-rao-a:
Functions from shared libraries seem to show the comments, but not the ones in main.go which is a special go file meant for executables vs shared code

ramya-rao-a:
We use `godoc` to get the function documentation to display in Signature Help and Quick Info on hover.

`godoc` is meant to show documentation for exported members of packages.

Which is why no member in main.go or unexported members of packages do not get the documentation info.

Replacing bug label with enhancement

ramya-rao-a:
In the latest update of the Go extension (0.6.50), we now support the use of `gogetdoc` for **Goto/Peek Definition, Hover info and Signature Help** features. Add `"go.docsTool": "gogetdoc"` to your setting to enable the use of `gogetdoc` and that should fix this issue.

Issue #515, opened 2016-10-06
Go to definition and type info doesn't work with mux.Vars or anything else from gorilla/mux


DmitriyMV:
I have no idea idea why - maybe because of the build tags? Other packages work fine.

ironcladlou:
What platform and version of Go? I'm using `gorilla/mux` (and other Gorilla packages) and completion works fine for me (OSX 10.11.6 / Go 1.7).

DmitriyMV:
`go version go1.7.1 linux/amd64`
`vs_code_1.6.0-1476123947_amd64`
Auto-complete works (suddenly, for some reason), but try to hover over mux.Vars function for example. In normal case, this should show function info, but it doesn't. Go to definition doesn't work either.

ramya-rao-a:
I can repro this. We use `godef` for Goto definition and the hover features. When I hover over `mux.Router` it  returns an error `godef: no declaration found for mux.Router` 

This is an issue with `godef`. Can you open an issue in the repo for `godef` ?

jadrio:
I have the same issue
OS Ubuntu 16.10
go version go1.7.1 linux/amd64
vscode 1.7.1 02611b40b24c9df2726ad8b33f5ef5f67ac30b44

ramya-rao-a:
In the latest update of the Go extension (0.6.50), we now support the use of `gogetdoc` for **Goto/Peek Definition, Hover info and Signature Help** features. Add `"go.docsTool": "gogetdoc"` to your setting to enable the use of `gogetdoc` and that should fix this issue.

Issue #496, opened 2016-09-21
Goto Definition, Hover, Signature Help do not work for Dot imported functions 


Alexendoo:
Using 0.6.43

The function details show correctly

![image](https://cloud.githubusercontent.com/assets/1830331/18731770/00853dba-8056-11e6-8717-c1f06b61ad63.png)

but argument details do not

![image](https://cloud.githubusercontent.com/assets/1830331/18731779/0a47fe3c-8056-11e6-87e4-dddf7081d17e.png)

Becomes an issue in testing when using a different package for the tests or a project such as [goconvery](https://github.com/smartystreets/goconvey)

ramya-rao-a:
The signature help feature uses the `godef` to get package info and then uses `godoc` to get the signature help. The `godef` tool does not support dot imports. See https://github.com/rogpeppe/godef/issues/19

Other features of the Go extension that use the `godef` and so will not work with dot imports are:

- Extra Info on hover
- Goto and Peek definition

ramya-rao-a:
Closing this as won't fix, as the tools we use don't support the dot imported functions

ramya-rao-a:
#622 can fix this, so re-opening

ramya-rao-a:
In the latest update of the Go extension (0.6.50), we now support the use of `gogetdoc` for **Goto/Peek Definition, Hover info and Signature Help** features. Add `"go.docsTool": "gogetdoc"` to your setting to enable the use of `gogetdoc` and that should fix this issue.

paulcalcraft:
Make sure you have gogetdoc installed by running:
`go get github.com/zmb3/gogetdoc`

It's not installed as standard by go or vscode's go extension, and the above setting ("go.docsTool": "gogetdoc") will fail silently if it is not installed.

unix2dos:
in 1.92 mac , do not work?
@ramya-rao-a

Issue #459, opened 2016-08-29
"Go to Definition" not working for structs


apkrieg:
This is a feature my colleagues and I use very often when programming in Go (and other languages). I'm not sure if this is already implemented and I have not enabled it or if it is simply missing.

apkrieg:
Correction, it seems this feature does work, but only for functions and not `struct`s, fields, or `interface`s. It should work for all types.

ramya-rao-a:
@zmb3 @abarisain Would integrating `gogetdoc` help here?

abarisain:
I think so !

abarisain:
Actually, it already works on my gogetdoc2 branch.

@apkrieg if you can do a small repro file that fails with the current version, I could confirm that gogetdoc fills this gap

apkrieg:
```
package main

import (
	"database/sql"
	"log"

	"github.com/go-sql-driver/mysql"
)

var (
	DBConn *sql.DB
)

func init() {
	log.Println("mcrm: Initializing...")
}

func main() {
	log.Println("mcrm: Starting...")
	// Establish a database connection
	dbconfig := mysql.Config{
		Addr:   "localhost:3066",
		User:   "mcrm_usr",
		Passwd: "---",
		DBName: "mcrm_db",
		Params: map[string]string{"parseTime": "true"},
	}
	var err error
	DBConn, err = sql.Open("mysql", dbconfig.FormatDSN())
	if err != nil {
		log.Fatalf("mcrm: error: Could not create DB connection.\n\tMore Info: %s\n", err.Error())
	}
	err = DBConn.Ping()
	if err != nil {
		log.Fatalf("mcrm: error: Unable to connect to the DB.\n\tMore Info: %s\n", err.Error())
	}
	// ---
	defer func() {
		log.Println("mcrm: Exiting...")
		err = DBConn.Close()
		if err != nil {
			log.Fatalf("mcrm: error: Could not close the DB connection. Exiting anyway.\n\tMore Info %s\n", err.Error())
		}
	}()
}
```
If you try to go to the definition of `mysql.Config` on line 21, it goes nowhere.

apkrieg:
I wrote this just to test the issue and it seems "Go to Definition" works for `structs`. Maybe it's just the sql library I'm using?

```
// Testing "Go to Definition"
package main

import (
	"github.com/satori/go.uuid"
	"fmt"
)

var (
	id uuid.UUID
)

func main() {
	id = uuid.NewV4()
[...]

PR #622, opened 2016-11-17
Use gogetdoc instead of godef and godoc

ramya-rao-a:
Currently we use `godef` for Goto/Peek Definition.
`godef` + `godoc` together provide the Hover Info and Signature Help.

These tools have shortcomings which we hope to overcome by using `gogetdoc` instead.
#440 #442 #515 #567 #459 #496

This PR is work on top of what @leaxoy has done in #607 
What has been added are the below
- A setting where user can choose `godoc` (old way of doing things `godef` + `godoc`)  vs `gogetdoc`
- If User is using Go 1.5, then we use the old way regardless of what is set
- The setting will default to the old way. There are some 3 or 4 issues logged by users which using `gogetdoc` will solve. We will reach out to those users and ask them to change the setting to use `gogetdoc` and give feedback
- Some refactoring
- Fix the failing  tests run by travis
- Add unit tests to cover the scenarios mentioned in the above mentioned issues


ramya-rao-a:
the unit tests are failing for 1.7 but pass in 1.6 mac... which is baffling

@leaxoy if you have some time to look at the test failures, I'd appreciate it

test/go.test.ts
```diff
@@ -53,28 +53,38 @@ encountered.
...
+			if (version.major > 1 || (version.major === 1 && version.minor > 5)) {
+				testCases[0][1] = 'func main()';
+				testCases[1][1] = 'package fmt';
+				testCases[2][1] = 'func Println(a ...interface{}) (n int, err error)';
+				testCases[3][1] = 'func print(txt string)';
+			}
+			return vscode.workspace.openTextDocument(uri).then((textDocument) => {
+				let promises = testCases.map(([position, expectedSignature, expectedDocumentation]) =>
+					provider.provideHover(textDocument, position, null).then(res => {
+						// TODO: Documentation appears to currently be broken on Go 1.7, so disabling these tests for now
```
abarisain: Can we give reenabling these tests a shot? When this was commented, they worked on my machine

src/goSignature.ts
```diff
@@ -28,10 +28,25 @@ export class GoSignatureHelpProvider implements SignatureHelpProvider {
...
+				let sigStart = nameEnd + 5; // ' func'
+				let funcName = text.substring(0, nameEnd);
+				sig = text.substring(sigStart);
+				si = new SignatureInformation(funcName + sig, res.doc);
+			} else {
+				// declaration is of the form "func Add(a int, b int) int"
+				text = res.declarationlines[0].substring(5);
+				si = new SignatureInformation(text, res.doc);
+				let braceStart = text.indexOf('(');
+				sig = text.substring(braceStart);
```
abarisain: Does that work for methods that have a signature like `func (l *Struct) Add(a int, b int) {`? I did it somewhat differently for that reason https://github.com/abarisain/vscode-go/blob/gogetdoc2/src/goSignature.ts#L33-L40

If it does, then sorry! Haven't had the time to test this branch yet.

ramya-rao-a:
All tests are a pass!

src/goSignature.ts
ramya-rao-a: Good catch, ya it doesnt work for that case

ramya-rao-a: In such case it thinks the receiver is the first param, thats a bug.

ramya-rao-a: Fixed it, Took your code and tested

ramya-rao-a: @abarisain The changes are in

test/go.test.ts
ramya-rao-a: @abarisain Done!

ramya-rao-a:
Added additional tests to cover various scenarios where using `gogetdoc` wins vs `godef`

src/goInstallTools.ts
```diff
@@ -22,13 +22,19 @@ function getTools(goVersion: SemVersion): { [key: string]: string } {
...
 		'go-outline': 'github.com/lukehoban/go-outline',
 		'go-symbols': 'github.com/newhook/go-symbols',
 		'guru': 'golang.org/x/tools/cmd/guru',
 		'gorename': 'golang.org/x/tools/cmd/gorename'
 	};
 
+	// Install the doc/def tool that was chosen by the user
+	if (goConfig['docsTool'] === 'godoc') {
+		tools['godef'] = 'github.com/rogpeppe/godef';
+	} else if (goConfig['formatTool'] === 'goreturns') {
```
abarisain: I think that should be `} else if (goConfig['docsTool'] === 'gogetdoc') {` :)

zmb3:
LGTM :-)

abarisain:
Ditto

abarisain:
Great addition to the plugin ! Congrats @zmb3

PR #639, opened 2016-11-22
Add support for reading http.proxy setting

ibigbug:
Make `go get` honor http.proxy setting so that user can not only provide their proxy settings via environment variables

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -115,9 +115,11 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 
 	outputChannel.appendLine(''); // Blank line for spacing.
 
+	let http_proxy = vscode.workspace.getConfiguration('http').get('proxy');
+	let https_proxy = http_proxy;  // default proxy for http & https
 	missing.reduce((res: Promise<string[]>, tool: string) => {
 		return res.then(sofar => new Promise<string[]>((resolve, reject) => {
-			cp.execFile(goRuntimePath, ['get', '-u', '-v', tools[tool]], { env: process.env }, (err, stdout, stderr) => {
+			cp.execFile(goRuntimePath, ['get', '-u', '-v', tools[tool]], { env: Object.assign({}, process.env, { http_proxy, https_proxy }) }, (err, stdout, stderr) => {
 				if (err) {
 					outputChannel.appendLine('Installing ' + tool + ' FAILED');
 					let failureReason = tool + ';;' + err + stdout.toString() + stderr.toString();
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -115,11 +115,20 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 
 	outputChannel.appendLine(''); // Blank line for spacing.
 
-	let http_proxy = vscode.workspace.getConfiguration('http').get('proxy');
-	let https_proxy = http_proxy;  // default proxy for http & https
+	// http.proxy setting takes precedence over environment variables
+	let httpProxy = vscode.workspace.getConfiguration('http').get('proxy');
+	let env = process.env;
+	if (httpProxy) {
+		env = Object.assign({}, process.env, {
+			http_proxy: httpProxy,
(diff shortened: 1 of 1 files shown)


ibigbug:
Oh ci complains.

src/goInstallTools.ts
```diff
@@ -115,9 +115,11 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 
 	outputChannel.appendLine(''); // Blank line for spacing.
 
+	let http_proxy = vscode.workspace.getConfiguration('http').get('proxy');
+	let https_proxy = http_proxy;  // default proxy for http & https
```
ramya-rao-a: @ibigbug  Sorry for the very late review.

Have you tested the case where http_proxy is not set in settings , but is set in the environment variable?
In that case will the value set in the environment variable make its way to the child process running the `go get` command?

ibigbug: @ramya-rao-a Yeah. I'm sure http_proxy in environment variables and open VS code from this environment using `$ code` can make `go get` work through proxy, though it's limited use case. User may often not open VS code through terminal and for OS X users (like me) and other Linux users, environment variables for terminal and GUI are always isolated, this is the main consideration for this PR.

ramya-rao-a: @ibigbug I understand the use case you are trying to solve. My question was for the use case where a user might have set the env var for http_proxy, but not in settings. What happens then, with your code changes?

I tested your code in 2 scenarios in Windows, this is what I see for the `env` that gets passed to the child process running the `go get`

1. When `http_proxy` is set in env var but not in VS Code settings
image

2. When different values are set for `http_proxy` in env var and VS Code settings
image

You will have to do 2 changes I believe
1. Use upper case while setting the http_proxy variable
2. If there is no VS Code setting for http_proxy, then do not change the process.env

src/goInstallTools.ts
```diff
@@ -115,9 +115,20 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 
 	outputChannel.appendLine(''); // Blank line for spacing.
 
+	// http.proxy setting takes precedence over environment variables
+	let httpProxy = vscode.workspace.getConfiguration('http').get('proxy');
+	let env = process.env;
+	if (httpProxy) {
+		env = Object.assign({}, process.env, {
+			http_proxy: httpProxy,
```
ibigbug: @ramya-rao-a Thanks for pointing out the problems which I did't test carefully.

1. I believe golang follows both lower and upper case  proxy env vars: https://github.com/golang/go/blob/master/src/net/http/transport.go#L246 , do you think that it's OK to set both lower and upper case env vars here?
2. I've added an detection if http.proxy was present.

ramya-rao-a: any benefits of setting both upper and lower case?
Setting just the upper case should do so that if any that was set outside vs code get overridden.

ibigbug: Yes I believe there are.

In case `go get` are calling some external tools such as `wget` to download third party resources or other tools like `wget` which read lowercase setting from environment variables:  https://www.gnu.org/software/wget/manual/wget.html#Proxies

So I think setting both can be safer.

ramya-rao-a: Sounds good. Only thing pending is the failing tests in travis. Can you debug and see if it fails in your local set up as well?

ibigbug: I found the failed test case is `Test diffUtils.getEdits` however I have no idea how does this commit make a difference on that. I'll have a local debug to see what happened.

ibigbug: @ramya-rao-a It's interesting that running 'npm t` on current master branch also gets 6 failed tests. I wonder if there are any dependencies have had a breaking change?

ramya-rao-a: I have fixed the tests in master. Re-running the travis tests for this PR keeps using the version of master before my fix.
So will go ahead and merge this without waiting for tests to turn green for this PR.

I was out on winter holidays, so couldn't help wrap this up sooner. 
Thanks for your PR and I look forward to more in the future

PR #648, opened 2016-11-24
Add benchmark function snippet

viru:
Analogical to existing `tf` snippet.

### DIFF
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -156,6 +156,10 @@
 			"prefix": "tf",
 			"body": "func Test$1(t *testing.T) {\n\t$0\n}"
 		},
+		"benchmark function": {
+			"prefix": "bf",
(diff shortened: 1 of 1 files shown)


snippets/go.json
```diff
@@ -156,6 +156,10 @@
 			"prefix": "tf",
 			"body": "func Test$1(t *testing.T) {\n\t$0\n}"
 		},
+		"benchmark function": {
+			"prefix": "bf",
+			"body": "func Benchmark$1(b *testing.B) {\n\tfor n := 0; n < b.N; n++ {\n\t\t$0\n\t}\n}"
```
ramya-rao-a: Allow to change the variable `n` in the for loop here just like the for loop snippet. And name it `i` or `index` to begin with.

viru: I will change `n` to `i`, but allowing to change that variable doesn't make sense for benchmark functions as Go chooses value of `b.N` on its own.

ramya-rao-a:
@viru The latest update 0.6.51 has your snippet. Thanks for your contribution!

PR #655, opened 2016-11-28
Fix "Go to Definition" when running Go HEAD

fiam:
Don't return a failure in definitionLocation() when getGoVersion()
returns a null version. Instead, assume it's newer than Go 1.5.


captncraig:
I'd love to have this change in the marketplace version. Any chance of pushing an update soon?

ramya-rao-a:
@captncraig Just did.
@fiam Thanks for the fix!

PR #657, opened 2016-11-28
Fix #656 "gotests" when generates test for current function

### LINKED ISSUE #656: Fix "gotests" when generates test for current function
Issue: "Generate unit tests for current function" fails with recent vscode-go version. This due to the fact that the GoDocumentSymbolProvider returns now also the receiverType in the name function.

example: 
- before ```DecodeJSON```
- last vscode-go version: ```(*Volume).DecodeJSON``` 

```gotests``` bin is not able to understand this string with the option -only since it is not a valid regular expression.

Fix: check is the function name contains a ".", if yes: returns only the second part of the name that is corresponding to the function name without the receiverType.

### PR DESCRIPTION
**Pull request by cedriclam**
Issue:
"Generate unit tests for current function" fails with recent
vscode-go version. This due to the fact that the GoDocumentSymbolProvider
returns now also the receiverType in the name function.
example: before "DecodeJSON", after (*Volume).DecodeJSON
gotests bin is not able to understand this string with the option -only
since it is not a valid regular expression.

Fixes: #656
check is the function name contains a ".", if yes: returns only the
second part of the name that is corresponding to the function name
without the receiverType.

### CODE REVIEW

File: src/goGenerateTests.ts [hunk 1]
```diff
@@ -86,33 +85,41 @@ export function generateTestCurrentFile(): Thenable<boolean> {
...
+			};
+			if (!currentFunction) {
+				vscode.window.setStatusBarMessage('No function found at cursor.', 5000);
+				return;
+			}
+			let funcName = currentFunction.name;
+			if (funcName.includes('.')) {
+				funcName = funcName.split('.')[1];
+			}
+			generateTests({ dir: file, func: funcName }).then((success: boolean) => {
```
ramya-rao-a: You don't need a new promise at line 92. The below will do
```
return getFunctions(editor.document).then(functions => {
...
...
      return generateTests({dir: file, fun: funcName});
}

File: test/go.test.ts [hunk 2]
```diff
@@ -19,19 +19,25 @@ import { testCurrentFile } from '../src/goTest';
...
-import { generateTestCurrentFile, generateTestCurrentPackage } from '../src/goGenerateTests';
+import { generateTestCurrentFile, generateTestCurrentPackage, generateTestCurrentFunction } from '../src/goGenerateTests';
 import { getBinPath } from '../src/goPath';
 
 suite('Go Extension Tests', () => {
 	let gopath = process.env['GOPATH'];
 	let repoPath = path.join(gopath, 'src', 'test');
 	let fixturePath = path.join(repoPath, 'testfixture');
 	let fixtureSourcePath = path.join(__dirname, '..', '..', 'test', 'fixtures');
+	let generateTestsSourcePath = path.join(repoPath, 'generatetests');
```
ramya-rao-a: why not just re-use `fixturePath` ?

test/go.test.ts
```diff
@@ -298,14 +304,39 @@ It returns the number of bytes written and any write error encountered.
...
+			}
+
+			let uri = vscode.Uri.file(path.join(generateFunctionTestSourcePath, 'generatetests.go'));
+			return vscode.workspace.openTextDocument(uri).then(document => {
+				return vscode.window.showTextDocument(document).then((editor: vscode.TextEditor) => {
+					assert(vscode.window.activeTextEditor, 'No active editor');
+					let selection = new vscode.Selection(5, 0, 6, 0);
+					editor.selection = selection;
+					return generateTestCurrentFunction().then((result: boolean) => {
+						assert.equal(result, true);
```
ramya-rao-a: Ideally, I'd like to check whether the test file was created with the test function. You don't have do it now, but if you have some time, I'd appreciate it.

test/fixtures/test.go
```diff
@@ -9,4 +9,4 @@ func print(txt string) {
 }
 func main() {
 	print("Hello")
-}
+}
```
ramya-rao-a: Removal of this line at the end is breaking one of the tests which tries to add some text there

test/go.test.ts
cedriclam: if the tests use the same file, 2 tests will failed, since the all tests will be already generated by the "generate test in current package"

src/goGenerateTests.ts
cedriclam: Yes, thanks

test/fixtures/test.go
cedriclam: test updated

test/go.test.ts
ramya-rao-a: Got it, thanks!

test/go.test.ts
cedriclam: Good idea and done.

cedriclam:
@ramya-rao-a  thanks for your quick review !

ramya-rao-a:
My pleasure @cedriclam. Thanks for your contributions

PR #658, opened 2016-11-28
Fixed wrong workspace when one gopath is the substring of another.

PotterDai:
Fixed wrong workspace when one gopath is the substring of another in “canonicalizeGOPATHPrefix” of “util.ts”.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -98,12 +98,17 @@ export function canonicalizeGOPATHPrefix(filename: string): string {
 	if (!gopath) return filename;
 	let workspaces = gopath.split(path.delimiter);
 	let filenameLowercase = filename.toLowerCase();
+
+	let currentWorkspace : string = null
 	for (let workspace of workspaces) {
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -98,12 +98,17 @@ export function canonicalizeGOPATHPrefix(filename: string): string {
 	if (!gopath) return filename;
 	let workspaces = gopath.split(path.delimiter);
 	let filenameLowercase = filename.toLowerCase();
+
```
ramya-rao-a: Can you add a comment here about the logic being used, so that the next person knows whats going on?
The unit tests are failing in travis due to linting issues. Run `npm run lint` locally and fix the reported errors.

ramya-rao-a:
Thanks @PotterDai ! Will merge as soon as all the tests pass.

ramya-rao-a:
The latest update 0.6.51 has this fix.

PR #693, opened 2016-12-18
fix coverage display when editor changes to undefined

nwolber:
getCodeCoverage gets triggered when vscode.window.onDidChangeActiveTextEditor fires. onDidChangeActiveTextEditor also fires when the active editor changes to undefined.


ramya-rao-a:
Thanks for the fix!

Issue #465, opened 2016-09-01
"gotoSymbol" failed when Chinese words are in comments


dbaeumer:
_From @cjqpker on September 1, 2016 3:24_
- VSCode Version: Code 1.4.0 (6276dcb0ae497766056b4c09ea75be1d76a8b679, 2016-08-04T16:41:29.513Z)
- OS Version: Darwin x64 15.6.0

Steps to Reproduce:
1. shift+cmd+o, input some keywords
   2.vscode locates a strange position

![img_20160901_111603](https://cloud.githubusercontent.com/assets/14192091/18154361/7c8c89f2-7036-11e6-8bd7-7e14a2afd379.jpg)

_Copied from original issue: Microsoft/vscode#11363_

dbaeumer:
Moving to the vscode-go repository.

cjqpker:
ha?

lukehoban:
What is the issue here?  Could you share repro steps?

cjqpker:
language : golang
1.shift+cmd+o, input some keywords
2.select a function and press enter
2.vscode will locate a strange position

ramya-rao-a:
@cjqpker Can you share a sample code/repo where you see this?

I followed your repro steps and cannot see the issue. For me, after Step 1 and 2, cursor goes to the line where the selected function is defined.

Can you also share information about your OS platform, VS Code version and the Go extension version?

cjqpker:
I tested some codes and find out this problem appears when annotations include Chinese words.

cjqpker:
@ramya-rao-a

ramya-rao-a:
@cjqpker We were able to repro this issue by using Chinese words in comments. Thanks for reporting!

ramya-rao-a:
@cjqpker The new release of the extension (0.6.52) has the fix for this issue

PR #699, opened 2016-12-27
Fix outline offset problem in No-English language

bylevel:


ramya-rao-a:
@bylevel Would this fix #465 ?

bylevel:
@ramya-rao-a Yes, this will fix #465

ramya-rao-a:
Great! Thanks for the PR @bylevel

Issue #673, opened 2016-12-01
`go build -o` should be using `-i`


mattetti:
I noticed that we rebuild all packages every time a user saves a project with build on save on. I think it would be best to use the `-i` flag to only build packages that changed.

> The -i flag installs the packages that are dependencies of the target.

My understanding is that installed packages aren't rebuilt. I can see that building a project with a slow to build dependency such as `github.com/mattn/go-sqlite3` is much much faster when the project is built with `-i`.

@ramya-rao-a what do you think?

ramya-rao-a:
@mattetti 

> My understanding is that installed packages aren't rebuilt.

Is there any official documentation for this?

mattetti:
https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies

The -i flag installs the packages that are dependencies of the target.
Unfortunately I am not aware of documentation explaining the ramifications
of having packages installed.
Here is a Quora explaination:
https://www.quora.com/What-is-the-difference-between-build-and-install-in-Go
Basically what I saw is that when using -i in build or test, the dependency
packages are cached and reused if they haven't changed. You can test that
using an expensive pack to build such as sqlite3.
Once you build it once with -i, other builds can use the cache.

ramya-rao-a:
@derekperkins mentioned https://github.com/joshuarubin/zb in another issue. This claims to run faster builds. It comes down to build vs install there as well

mattetti:
I don't know if we should go all the way to pushing something like zb but
yes zb uses some of the same techniques. -i really shouldn't have any down
sides, besides making for faster builds, especially on save.

mattetti:
I'll look at adding this flag and send a PR, @ramya-rao-a would you mind allocating this issue to me?

ramya-rao-a:
Done

ramya-rao-a:
@mattetti 

I was doing some reading up on `go build` in general.
One of the build flags is `-a` which according to the docs will `force rebuilding of packages that are already up-to-date.`

If there is such a flag, then it doesnt make sense that `go build -o` is rebuilding all packages while building current package (Save triggers `go build -o` on current package)

I'd prefer to see some documentation on why `go build -i` is faster than `go build -o` before making this a default experience. Can we add setting for this change so that people can opt-in to use `go build -i`?

mattetti:
[...]

PR #718, opened 2017-01-11
if buildOnSave is on: build using -i

mattetti:
This PR addresses the compilation performance optimization discussed in #673.

Installing the built packages improves following compilations since packages are only rebuilt if they changed.
If the current package is a main package, the installation is skipped to avoid installing executable binaries.

I had to tweak a util method that was scanning text files to find the package and the imports. Unfortunately this util wasn't tracking the name of the package. I needed the package name to check if the `-i` flag should be used or not.

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -14,6 +14,7 @@ import { getBinPath, getGoRuntimePath } from './goPath';
 import { getCoverage } from './goCover';
 import { outputChannel } from './goStatus';
 import { promptForMissingTool } from './goInstallTools';
+import { parseFilePrelude } from './util';
 
 export interface ICheckResult {
 	file: string;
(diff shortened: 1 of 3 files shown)


mattetti:
@ramya-rao-a would you mind helping me understand what's wrong with the test suite? I seem to have broken something on macOS. The failures don't make too much sense to me :(

ramya-rao-a:
@mattetti The 2 tests are failing because they are not getting back expected errors from `go build`.
You are reading the file using `fs.readFile` and calling the `runTool` in the callback. By the time the callback is called, the test has already completed, and so does not get the results of `go build`.

Here is what you can try:
```
let buildPromise = new Promise<ICheckResult[]>((resolve, reject) => {
    fs.readFile(filename. 'utf8', (err, data) => {
        .....
        runTool(....).then(result => return resolve(result), err => return reject(err));
    }
});
runningToolsPromises.push(buildPromise);
```

mattetti:
perfect thank you! I'll give it a try

mattetti:
@ramya-rao-a PR updated, tests are now passing, thanks for your help!

Note, this PR is pending the resolution of the discussion happening there: https://github.com/Microsoft/vscode-go/issues/673

PR #720, opened 2017-01-13
Use proper plurality in installation console output

verygoodsoftwarenotvirus:
Relatively meaningless PR to fix an extremely minor annoyance


ramya-rao-a:
:) Thanks @verygoodsoftwarenotvirus

PR #725, opened 2017-01-16
Use GOPATH from env if dlv not found

ramya-rao-a:


src/debugAdapter/goDebug.ts
```diff
@@ -169,6 +169,14 @@ class Delve {
 				return;
 			}
 			let dlv = getBinPath('dlv');
+
+			// If dlv not found, try using the GOPATH that was set in env in launch.json
+			if (!existsSync(dlv) && env['GOPATH'] && !process.env['GOPATH']) {
+				process.env['GOPATH'] = env['GOPATH'];
+				dlv = getBinPath('dlv');
+				process.env['GOPATH'] = undefined;
```
tw1nk: Wouldn't it be nicer if the getBinPath took an (optional) second argument called gopath and used that instead of changing the process.env['GOPATH']?

ramya-rao-a: good point, I like it :)

ramya-rao-a:
@tw1nk Updated as per your suggestion. I agree, its much cleaner this way. Thanks!

guogeer:
image
I update vscode-go version to newest and there is a error. like above image, if launch config miss env option,it happen Failed to continue: "TypeError: Cannot read property 'GOPATH' of undefined.

ramya-rao-a:
Thanks @guogeer, I found this the other day too. The fix is already in master. The next update of the Go extension will have the fix.

PR #731, opened 2017-01-16
Replace VSCODE_GOTOOLS env var with go.toolsGopath config

samherrmann:


roblourens:
Right, since goPath is shared with goDebug, it can't depend on the vscode module.

samherrmann:
Just a heads up that I unfortunately do not have time to work on this for the next few weeks, so if anyone wants to take over feel free to do so.

ramya-rao-a:
Thanks for the heads up @samherrmann 
Closing this PR in favor of #737

PR #737, opened 2017-01-21
Use settings for separate location for installing Go tools

ramya-rao-a:
See #5 for the discussion on the need for a separate location to install Go tools instead of the user's GOPATH.

See #351 for the implementation using env vars and further talk of how a VS Code setting is better instead of env var

This PR is to use VS Code setting where user can provide a location for installing the Go tools.
`gometalinter` is an exception to this rule and will get installed in the user's GOPATH always.
This is because `gometalinter` expects all the linters it uses to be in the user's GOPATH.

Users need to install `dlv` manually as always and is out of this disucssion

cc @roblourens

### DIFF
--- a/package.json
+++ b/package.json
@@ -388,6 +388,11 @@
             "godoc",
             "gogetdoc"
           ]
+        },
+        "go.toolsGopath": {
+          "type": "string",
+          "default": "",
+          "description": "Location to install the Go tools that the extension depends on if you don't want them in your GOPATH."
         }
       }
     }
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -388,6 +388,11 @@
             "godoc",
             "gogetdoc"
           ]
+        },
+        "go.toolsGopath": {
```
roblourens: Capital P?

roblourens: Or maybe gopath is like one word, either way

ramya-rao-a: I feel different everytime I think about it :(

PR #740, opened 2017-01-22
Map unexpected errors to first line

### LINKED ISSUE #664: No build errors when code fails on import cycle
When I get an accidental import cycle, I get zero build errors in VSCode. But manually building, I of course get:

```
$ go build -v github.com/sanity-io/gradient/pkg/search/queryengine
can't load package: import cycle not allowed
[etc.]
```

This is frustrating because when this happens, it appears that compilation/linting has just stopped working.

(Funnily enough, Atom's `go-plus` package has the exact same issue.)

My Go settings:

```
	"go.buildOnSave": true,
	"go.lintOnSave": true,
	"go.vetOnSave": true,
	"go.lintTool": "gometalinter",
	"go.lintFlags": [
		"--disable-all",
		// Linters
		"--enable=deadcode",
		"--enable=vetshadow",
		"--enable=golint",
		//"--enable=errcheck",
		"--enable=ineffassign",
		"--enable=goconst",
		// Exclude stuff
		"--exclude=\\.pb\\.go",
		"--exclude=or be unexported",
		"--exclude=that.stutters",
		"--exclude=don't use underscores",
		"--exclude=\\.nolint\\.go",
		"--vendor",
		"--tests"
	],
	"go.useCodeSnippetsOnFunctionSuggest": false,
	"go.formatOnSave": true,
	"go.formatTool": "goreturns",
```

ramya-rao-a:
Fixes #664

cc @roblourens

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -55,6 +55,8 @@ function runTool(args: string[], cwd: string, severity: string, useStdErr: boole
 				outputChannel.appendLine(['Finished running tool:', cmd, ...args].join(' '));
 
 				let ret: ICheckResult[] = [];
+				let unexpectedOutput = false;
+				let atleastSingleMatch = false;
 				for (let i = 0; i < lines.length; i++) {
 					if (lines[i][0] === '\t' && ret.length > 0) {
 						ret[ret.length - 1].msg += '\n' + lines[i];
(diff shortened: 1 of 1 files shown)


src/goCheck.ts
```diff
@@ -55,22 +55,34 @@ function runTool(args: string[], cwd: string, severity: string, useStdErr: boole
...
 					file = path.resolve(cwd, file);
 					ret.push({ file, line, msg, severity });
 					outputChannel.appendLine(`${file}:${line}: ${msg}`);
 				}
+				if (!atleastSingleMatch && unexpectedOutput && useStdErr && vscode.window.activeTextEditor) {
+					outputChannel.appendLine(stderr);
+					ret.push({
+						file: vscode.window.activeTextEditor.document.fileName,
+						line: 1,
+						msg: 'Something went wrong. Check the Go channel in the output window',
```
ramya-rao-a: Good point. Done.

roblourens: Oh man I did that thing again where I typed that this morning before standup but didn't submit the review. Just had to say that since we talked about this at standup.

Issue #689, opened 2016-12-15
Can't set breakpoins in libraries when remote debugging


f0zi:
- What version of Delve are you using (dlv version)?
  - 0.11.0-alpha on both Windows + Linux
- What version of Go are you using? (go version)?
  - go version go1.7 windows/amd64
  - go version go1.7.4 linux/amd64
- What operating system and processor architecture are you using?
  - Windows 10 64 bit running VS code
  - Any linux (tested with Ubuntu 16.4) running the test subject & headless dlv
- What did you do?
  - At debug time, add/remove breakpoint in a library file
- What did you expect to see?
  - Breakpoint accepted and debugee paused when hit
- What did you see instead?
  - Unverified breakpoint, breakpoint not set.

In my launch.json I have the remotePath set to `/home/fozi/go/src/test/main`, If I set it to anything else (e.g. opening the `test` folder and remove `/main`) none of the breakpoints work. The library I tested with is in `/home/fozi/go/src/test/lib`, but It would be nice if this would work at the very least in anything under `$GOPATH/src` (assuming it's reachable) including vendor folders. Windows (and VS Code) accessed these files on a samba share so the paths look like `\\vmmachine\fozi\go\src\test\main`.

I tried this in dlv command line mode directly and it seems to work fine. I think it's simply that the file name translation needs to be more sophisticated.

f0zi:
Attaching a diff that solves that for me.

[goDebug.ts.diff.txt](https://github.com/Microsoft/vscode-go/files/717247/goDebug.ts.diff.txt)

Note: I did `findPathSeperator` because I often have something like `"program": "${workspaceRoot}/test"` in my launch.json which gives the wrong results on Windows.

f0zi:
Use this instead, fixed "completely different" case:

[goDebug.ts.diff.txt](https://github.com/Microsoft/vscode-go/files/724908/goDebug.ts.diff.txt)

ramya-rao-a:
@f0zi Can you submit a PR with details on the problem you are trying to solve?
cc @roblourens

ramya-rao-a:
Fixed by #742

ramya-rao-a:
Fixed in the latest update (0.6.54) to the Go extension.

PR #742, opened 2017-01-24
Removing common parts from program and remotePath

f0zi:
also returning '\\' as a separator on Windows paths with mixed separators.

PR as requested in #689.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -339,7 +339,8 @@ class GoDebugSession extends DebugSession {
 
 			let llist = localPath.split(/\/|\\(?! )/).reverse();
 			let rlist = remotePath.split(/\/|\\(?! )/).reverse();
-			for(var i = 0; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
+			let i = 0;
+			for(; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
 
 			if (i) {
 				localPath = llist.reverse().slice(0, -i).join(this.localPathSeparator);
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -340,7 +340,7 @@ class GoDebugSession extends DebugSession {
 			let llist = localPath.split(/\/|\\(?! )/).reverse();
 			let rlist = remotePath.split(/\/|\\(?! )/).reverse();
 			let i = 0;
-			for(; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
+			for (; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
 
 			if (i) {
 				localPath = llist.reverse().slice(0, -i).join(this.localPathSeparator);
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -321,21 +321,35 @@ class GoDebugSession extends DebugSession {
 		log('InitializeEvent');
 	}
 
+	protected findPathSeperator(path) {
+		if (/^(\w:[\\/]|\\\\)/.test(path)) return '\\';
+		return path.includes('/') ? '/' : '\\';
+	}
+
 	protected launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
 		// Launch the Delve debugger on the program
+		let localPath = args.program;
 		let remotePath = args.remotePath || '';
 		let port = args.port || random(2000, 50000);
 		let host = args.host || '127.0.0.1';
 
 		if (remotePath.length > 0) {
-			this.localPathSeparator = args.program.includes('/') ? '/' : '\\';
-			this.remotePathSeparator = remotePath.includes('/') ? '/' : '\\';
-			if ((remotePath.endsWith('\\')) || (remotePath.endsWith('/'))) {
+			this.localPathSeparator = this.findPathSeperator(localPath);
+			this.remotePathSeparator = this.findPathSeperator(remotePath);
+
+			let llist = localPath.split(/\/|\\(?! )/).reverse();
+			let rlist = remotePath.split(/\/|\\(?! )/).reverse();
+			for(var i = 0; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
+
+			if (i) {
+				localPath = llist.reverse().slice(0, -i).join(this.localPathSeparator);
+				remotePath = rlist.reverse().slice(0, -i).join(this.remotePathSeparator);
+			} else if ((remotePath.endsWith('\\')) || (remotePath.endsWith('/'))) {
 				remotePath = remotePath.substring(0, remotePath.length - 1);
 			}
 		}
 
-		this.delve = new Delve(args.mode, remotePath, port, host, args.program, args.args, args.showLog, args.cwd, args.env, args.buildFlags, args.init);
+		this.delve = new Delve(args.mode, remotePath, port, host, localPath, args.args, args.showLog, args.cwd, args.env, args.buildFlags, args.init);
 		this.delve.onstdout = (str: string) => {
 			this.sendEvent(new OutputEvent(str, 'stdout'));
 		};


src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,35 @@ class GoDebugSession extends DebugSession {
...
+
 	protected launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
 		// Launch the Delve debugger on the program
+		let localPath = args.program;
 		let remotePath = args.remotePath || '';
 		let port = args.port || random(2000, 50000);
 		let host = args.host || '127.0.0.1';
 
 		if (remotePath.length > 0) {
-			this.localPathSeparator = args.program.includes('/') ? '/' : '\\';
```
f0zi: This fails for the (for me) common case where I share a project (and it's launch.json) and open it on different platforms and having a setting like `"program": "${workspaceRoot}/test"` because `${workspaceRoot}` has backslashes on Windows but slashes on Unix.

src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,35 @@ class GoDebugSession extends DebugSession {
...
-			this.remotePathSeparator = remotePath.includes('/') ? '/' : '\\';
-			if ((remotePath.endsWith('\\')) || (remotePath.endsWith('/'))) {
+			this.localPathSeparator = this.findPathSeperator(localPath);
+			this.remotePathSeparator = this.findPathSeperator(remotePath);
+
+			let llist = localPath.split(/\/|\\(?! )/).reverse();
+			let rlist = remotePath.split(/\/|\\(?! )/).reverse();
+			for(var i = 0; i < llist.length; i++) if (llist[i] !== rlist[i]) break;
+
```
f0zi: This finds (and counts) common path elements. Think:
`/home/user/go/src/foo/bar`
`C:\Users\user\gohome\src\foo\bar`

src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,35 @@ class GoDebugSession extends DebugSession {
...
+			if (i) {
+				localPath = llist.reverse().slice(0, -i).join(this.localPathSeparator);
+				remotePath = rlist.reverse().slice(0, -i).join(this.remotePathSeparator);
+			} else if ((remotePath.endsWith('\\')) || (remotePath.endsWith('/'))) {
```
f0zi: This removes the common elements found above.

src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,35 @@ class GoDebugSession extends DebugSession {
...
 				remotePath = remotePath.substring(0, remotePath.length - 1);
 			}
 		}
 
-		this.delve = new Delve(args.mode, remotePath, port, host, args.program, args.args, args.showLog, args.cwd, args.env, args.buildFlags, args.init);
+		this.delve = new Delve(args.mode, remotePath, port, host, localPath, args.args, args.showLog, args.cwd, args.env, args.buildFlags, args.init);
 		this.delve.onstdout = (str: string) => {
 			this.sendEvent(new OutputEvent(str, 'stdout'));
 		};
```
f0zi: The result is that you can now set breakpoints and debug libraries in e.g. `/home/user/go/src/foo/baz` even though your main package is in `/home/user/go/src/foo/bar`

ramya-rao-a:
@roblourens Can you take a look at this?

src/debugAdapter/goDebug.ts
roblourens: Is that true, since foo != bar? Isn't it like `/home/foo1/go/src/bar` will match `/home/foo2/go/src/bar` because the two paths will be trimmed to `/home/foo1` and `/home/foo2`.

src/debugAdapter/goDebug.ts
roblourens: How about this, based on how we do it for node debugging. The program field isn't used for remote debugging, right? So instead of specifying the program, the user could specify the final 'localPath' to match 'remotePath'. Then they are responsible for doing this in their launch config - finding the common path of the two.

Would that end up being more confusing? I'm not opposed to this, just thinking about it.

Other debug adapters have 'launch'-type and 'attach'-type configs, where 'launch' is when it has to start executing the binary, and 'attach' is when it has already been started and is in debug mode. We could introduce that concept here too. Then the 'program' field would not be allowed for remote debug scenarios, just 'remotePath' and 'localPath' to provide the mapping.

src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,36 @@ class GoDebugSession extends DebugSession {
 		log('InitializeEvent');
 	}
 
+	protected findPathSeperator(path) {
+		if (/^(\w:[\\/]|\\\\)/.test(path)) return '\\';
```
f0zi: Not sure how that would help. Trying to match absolute Windows paths like C:\ or C:/ and share paths like \\server\share. It also matches \\?\, so it does not need it's own check.

src/debugAdapter/goDebug.ts
```diff
@@ -321,21 +321,36 @@ class GoDebugSession extends DebugSession {
...
 
 		if (remotePath.length > 0) {
-			this.localPathSeparator = args.program.includes('/') ? '/' : '\\';
-			this.remotePathSeparator = remotePath.includes('/') ? '/' : '\\';
-			if ((remotePath.endsWith('\\')) || (remotePath.endsWith('/'))) {
+			this.localPathSeparator = this.findPathSeperator(localPath);
+			this.remotePathSeparator = this.findPathSeperator(remotePath);
+
+			let llist = localPath.split(/\/|\\(?! )/).reverse();
+			let rlist = remotePath.split(/\/|\\(?! )/).reverse();
```
f0zi: Yes, trying to skip an escaped space, which is the only backslash I regularly encounter on unix systems. Not sure how delve and this plugin handles filenames on unix, so you might be right, it might not be necessary. I can't try this right now, but I'll give it a try once I have the chance to. That said, I think its save to leave it in as spaces are not allowed at the beginning of a filename on Windows.

src/debugAdapter/goDebug.ts
f0zi: I think an alternative would be to use the local GOPATH and add a setting for the remote GOPATH. Everything after the $GOPATH\src\ could be used, however this solution makes some assumptions that are not always true (like that the code actually is in $GOPATH and that there is a /src/ folder in there, which can be false especially when remote debugging) while the current solution "just works".

src/debugAdapter/goDebug.ts
f0zi: Remember this is initialization. The actual paths we get from remote or send to remote are the `path` arguments to `toDebuggerPath` and `toLocalPath`. Now we have four components: 
- The main module path, not visible here, but worth noting as it played a role in the old code because localPath was equal to main module path, e.g. `C:\Users\foo\go\src\foo/bar`. Note the mixed separators come from `${workspace}/bar`. Launch.json is shared between Windows & Unix.
- The `localPath`, now shortened to e.g. `C:\Users\foo`
- The `remotePath`, shortened to e.g. `/home/foo2`
- The argument to the function, e.g. `/home/foo2/go/src/foo/baz` for `toLocalPath` or (think set BP) `C:\User\foo\go\src\foo\baz` for `toDebuggerPath`.

The functions `toLocalPath` and `toDebuggerPath` just replace the beginning of the paths and change the path separators.

src/debugAdapter/goDebug.ts
roblourens: It would just be simpler. But I guess if the path is like `C:/foo/bar` you still want to return a backslash

f0zi: also for \\\\server\\share

src/debugAdapter/goDebug.ts
roblourens: Yeah, I think that escaped characters will have been resolved by this point so when you have `a\ b` you'll have a string which contains `a`, literal space, `b`.

f0zi:
@roblourens So, to move things forward, I don't mind to remove the negative forward assertion, are there any other changes you would like to see? 

Again, please make sure you click 'view changes' as not all my comments appear here on this page.

src/debugAdapter/goDebug.ts
roblourens: That's not a bad idea either, but I don't want to make assumptions like that. I think that for now, we should keep the current approach with your changes.

roblourens:
Thanks @f0zi!

PR #750, opened 2017-01-27
Add option to use Go language server for certain features

ramya-rao-a:


netroby:
May be this language server enabled by default?

ramya-rao-a:
@netroby Language server is currently an experimental feature, so we wont be enabling this by default until we have ironed out current issues like Windows support

PR #752, opened 2017-01-29
Clear cache for tools when toolspath changes

ramya-rao-a:
When `go.toolsGopath` changes, we need to clear the cache (for tools path), so that the extensions starts looking at the new `go.toolsGopath` without needing a VS Code window reload

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -30,12 +30,13 @@ import { addImport } from './goImport';
 import { installAllTools, checkLanguageServer } from './goInstallTools';
 import { isGoPathSet, getBinPath } from './util';
 import { LanguageClient } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


src/goMain.ts
```diff
@@ -30,12 +30,15 @@ import { addImport } from './goImport';
 import { installAllTools, checkLanguageServer } from './goInstallTools';
 import { isGoPathSet, getBinPath } from './util';
 import { LanguageClient } from 'vscode-languageclient';
+import { clearCacheForTools } from './goPath';
 
 let diagnosticCollection: vscode.DiagnosticCollection;
 let useLangServer: boolean;
+let toolsGopath: string;
```
roblourens: Could be declared inside the function?

ramya-rao-a: Done

Issue #757, opened 2017-01-31
[Feature Request]  go.inferGopath. A GoSublime's use_gs_gopath like feature


codmajik:
in GoSublime specifying a $GS_GOPATH or use_gs_gopath cause it to lookup as follows:

- the current working directory, e.g. ~/go/src/pkg then $GS_GOPATH will be ~/go/
- or the path the current .go file (or last activated .go file if the current file is not .go) e.g. if your file path is /tmp/go/src/hello/main.go then it will be /tmp/go

in vscode-go:
- setting ```go.inferGopath: true``` should use the current workspace to resolve a suitable gopath, 
       e.g. **$HOME/go/src/github.com/myuser/demoapp** then **go.gopath** will be **$HOME/go/**

More info:
    [GoSublime Usage and Tips](https://github.com/DisposaBoy/GoSublime/blob/master/USAGE.md)

I personally prefer working with multiple windows, which is what is provided right now anyways. 
Having this would de-clutter explorer sidebar showing just files relevant to the project. Also prevent accidentally modifying files not related to your project.

OneOfOne:
+1 this is a handy feature if you wanna test a PoC in /tmp real quick without setting up a proper workspace.

ramya-rao-a:
PR is merged. This should be out in the next update.
Thanks for the work @codmajik, looking forward to work more with you on this extension.

ramya-rao-a:
This feature is now available in the latest update (0.6.54) to the Go extension.

PR #762, opened 2017-02-02
added go.inferGopath. Similar to GoSublime's use_gs_gopath

codmajik:
I would like to propose this implementation for the feature request raised in #757

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.formatFlags": [],
 	"go.goroot": "/usr/local/go",
 	"go.gopath": "/Users/lukeh/go",
+	"go.inferGopath": false,
 	"go.gocodeAutoBuild": false
 }
 ```
--- a/package.json
+++ b/package.json
@@ -337,6 +337,11 @@
           "default": false,
           "description": "Complete functions with their parameter signature"
         },
+        "go.inferGopath": {
+          "type": "boolean",
+          "default": false,
+          "description": "use workspace root to find appropriate GOPATH"
+        },
         "go.gopath": {
           "type": [
             "string",
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -212,12 +212,12 @@ export function updateGoPathGoRootFromConfig() {
 
 	let inferGoPath = vscode.workspace.getConfiguration('go')['inferGopath'];
 	if (inferGoPath) {
-		let dirs = vscode.workspace.rootPath.toLowerCase().split(path.sep)
+		let dirs = vscode.workspace.rootPath.toLowerCase().split(path.sep);
 		// find src directory closest to workspace root
-		let srcIdx = dirs.lastIndexOf("src")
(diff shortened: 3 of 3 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

ramya-rao-a:
@codmajik Thanks for the PR!

A few notes:

Take a case where the user has set `go.gopath` to a particular path, and also set `go.inferGopath` to true. Then according to your current logic:

- the inferred path wins. I was wondering if this should be the other way around. 
- If the user then sets `go.inferGopath` to false, you will need to reload the VS Code window for the change to take effect. If we only infer the go path, when there is no `go.gopath`, then you won't need the reload

Also, can you write a unit test for this as well? You will have to change `updateGoPathGoRootFromConfig` to take the `goConfig` as a parameter to do that.

codmajik:
@ramya-rao-a Thanks for your feedback.

The case you provided is the actual use case for this and also the current implementation in GoSublime. 

Where `go.gopath` is set in the global config and `go.inferGopath` is set in workspace config. Though
I am adding code to allow the reserve when `go.inferGopath` is global and  `go.gopath` is set in workspace

ramya-rao-a:
@codmajik I am assuming that when you say "global config", you mean the "User settings" in VS Code.

All configurations defined in the `package.json` are available for changes in both Workspace and User settings.

By the time you fetch the settings by `vscode.workspace.getConfiguration()`,  Workspace and User settings are already merged (workspace overrides user), so as an extension author, we have no idea if the value for a config was set in global/user settings or workspace settings.

Therefore, in our case here both `go.gopath` and `go.infergopath` can be set either in User or Workspace settings. We need to adapt for both.

Thinking more on this, it makes sense for the inferred path to win when both `go.gopath` and `go.inferGopath` are set.

codmajik:
@ramya-rao-a Thanks. 

I made those changes on my local checkout. I know it is not needed but I can create a new pull request with that as improvements. I just need to figure out how to test this with your current test suite.

btw `vscode.workspace.getConfiguration('go').inspect('gopath')` would give you the User Setting and Workspace Setting separately.

ramya-rao-a:
@codmajik Never mind the tests, I was thinking of doing a major refactoring in the unit test area soon, I can take care of that.

What changes are you referring to exactly?

codmajik:
@ramya-rao-a, changes to make sure that if `go.gopath` is set in the workspace config and but `go.inferGopath` is set in User Settings then workspace `go.path` would take precedence.

PS: should I delete the branch?

ramya-rao-a:
@codmajik I think its fine the way it is. You can delete your branch.

I have updated the descriptions for both `go.gopath` and `go.inferGopath` to reflect that the latter overrides the former. 

There is command `Go: Current GOPATH` that shows the GOPATH being used by the extension. It says: "Current GOPATH: ..."
We can update that to say "Current GOPATH is inferred from the workspace root: ..." when the GOPATH is inferred.

You can send a separate PR for that.

PR #768, opened 2017-02-05
Fixes #623 Expand ~ to os.homedir() in gopath

### LINKED ISSUE #623: Go plugin doesn't expand  ~ in GoPath
- VSCode Version: 1.7.1
- OS Version: Mac OSX Sierra 10.12.1
- Go PlugIn: 0.6.39
Steps to Reproduce:

1. set goPATH=~/gosrc and install plugins there using 'go get' 
2. Open VsCode and add following setting 

"go.gopath": "~/gosrc"

Save and exit.

3. Open VS code - VS code still prompts for missing plugin such as golint, etc
4. If you change ~ to the full path '/users/yizhang/gosrc',  go plugin would properly find the path

ramya-rao-a:


src/goPath.ts
```diff
@@ -95,4 +95,15 @@ function fileExists(filePath: string): boolean {
...
+
+/**
+ * Exapnds ~ to homedir in non-Windows platform and replaces ${workspaceRoot} token with given workspaceroot
+ */
+export function resolvePath(inputPath: string, workspaceRoot?: string): string {
+	if (!inputPath || !inputPath.trim()) return inputPath;
+	if (workspaceRoot) {
+		inputPath = inputPath.replace(/\${workspaceRoot}/g, workspaceRoot);
+	}
+	return (process.platform === 'win32' || !inputPath.startsWith('~')) ? inputPath : path.join(os.homedir(), inputPath.substr(1));
```
roblourens: Why not resolve ~ on windows?

ramya-rao-a: Because ~ being the home dir is a POSIX thing. Doesn't apply to Windows.

But I see your point, if there was a workspace setting for go path using ~ checked in to source control, then the same workspace should work on Windows machine.

Thoughts?

roblourens: Windows has it too now - ~ is expanded (only in powershell), and `os.homedir()` returns the right thing.

ramya-rao-a: Done

ramya-rao-a:
Sure mr. nitpicker :) Will do.

PR #774, opened 2017-02-06
When delve fails, stop the debug sessions

### LINKED ISSUE #492: Debugging a non-compiling app doesn't automatically end the debug session
## Repro
1. Apply the workaround in #474 on: VSCode 1.5.2, Go 1.7.1, Windows 10 Anniversary.
2. Create a simple hello world `main.go`.
3. Press <kbd>F5</kbd> and accept the defaults for `launch.json`.
4. Press <kbd>F5</kbd> and make sure that debugging works.
5. Add a simple compilation error (e.g. unused variable).
6. Press <kbd>F5</kbd>.
### Outcome

The app fails to compile, the debugger toolbar appears and the stop button must be clicked. The debug log shows `exit status 2`. The debugging session must be manually terminated.

Debug session is not terminated
### Expected Outcome

The app fails to compile and no debug session is started, or the debug session is automatically terminated.

ramya-rao-a:
Fixes #492

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -172,6 +172,7 @@ class Delve {
 	connection: Promise<RPCConnection>;
 	onstdout: (str: string) => void;
 	onstderr: (str: string) => void;
+	onclose: () => void;
 
 	constructor(mode: string, remotePath: string, port: number, host: string, program: string, args: string[], showLog: boolean, cwd: string, env: { [key: string]: string }, buildFlags: string, init: string) {
 		this.program = program;
@@ -268,9 +269,11 @@ class Delve {
 					connectClient(port, host);
 				}
 			});
+			let that = this;
 			this.debugProcess.on('close', function(code) {
 				// TODO: Report `dlv` crash to user.
 				logError('Process exiting with code: ' + code);
+				if (code !== 0 && that.onclose) { that.onclose(); }
 			});
 			this.debugProcess.on('error', function(err) {
 				reject(err);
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -268,9 +269,11 @@ class Delve {
 					connectClient(port, host);
 				}
 			});
+			let that = this;
```
ramya-rao-a: I am not sure why, but "this" in line 275 does not point to Delve object, it points to the ChildProcess which is why I had to resort to `that = this`

roblourens: You can use a fat arrow function (`code => { ... }`) instead of `function`, then `this` will be correct

src/debugAdapter/goDebug.ts
```diff
@@ -388,6 +391,10 @@ class GoDebugSession extends DebugSession {
 		this.delve.onstderr = (str: string) => {
 			this.sendEvent(new OutputEvent(str, 'stderr'));
 		};
+		this.delve.onclose = () => {
+			this.sendErrorResponse(response, 3000, 'Failed to continue: Check the debug console for details.');
```
roblourens: Why 'failed to continue'?

src/debugAdapter/goDebug.ts
ramya-rao-a: Oh ya! doh!

src/debugAdapter/goDebug.ts
ramya-rao-a: Cos it matches with the other message for when the delve.connection fails?

PR #780, opened 2017-02-08
fix offsets for files with multibyte characters

neelance:
When using multibyte characters like 🚨 in comments, actions like hover and go-to-def were using wrong offsets after that character.


ramya-rao-a:
Thanks @neelance!

ramya-rao-a:
The fix is now in the latest update (0.6.54) to the Go extension.

PR #794, opened 2017-02-14
Fixes #744 Adding configuration snippets

### LINKED ISSUE #744: Support configurationSnippets
In order to be cool as vanila ice chrome debugger should specify `configurationSnippets`
Here's how node is doing it https://github.com/Microsoft/vscode-node-debug/blob/master/package.json#L98

This will make the whole setup experience smother. It would be great if we could tackle this in january

ramya-rao-a:


package.json
```diff
@@ -155,6 +155,56 @@
...
+          },
+          {
+            "label": "Go: Launch file",
+            "description": "Debug the file in the program attribute",
+            "body":{
+              "name": "${2:Launch file}",
+              "type": "go",
+              "request": "launch",
+              "mode": "debug",
+              "program": "^\"\\${workspaceRoot}${1:/main.go}\""
```
roblourens: You might want to put the slash before the `${}` so they can replace `main.go` without removing the slash.

ramya-rao-a: good idea, this was a remnant of when I had combined debugggin package and file in which case one might want to remove the / as well

Issue #581, opened 2016-10-28
Add test on save feature


ramya-rao-a:
From the discussion in #451 , we should have a feature to run tests on save

`go.testOnSave` : true or false

This should run all the tests in the current package on saving any file in the current package

The test output should be shown in the output channel

ramya-rao-a:
This feature is now available in the latest update (0.6.54) to the Go extension.

PR #810, opened 2017-02-22
Add go.testOnSave option

dmage:
This is a very straightforward implementation of the testOnSave feature (#581).

I suppose it would be much better to compile sources only one time if buildOnSave, testOnSave, and coverOnSave are on: `go test -c` includes everything from `go build`, cover is obtained from tests.

Also it would be nice to have some sort of progress bar, but I have no idea where it should be placed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ The following Visual Studio Code settings are available for the Go extension.  T
 	"go.lintTool": "golint",
 	"go.lintFlags": [],
 	"go.vetFlags": [],
+	"go.testOnSave": false,
 	"go.coverOnSave": false,
 	"go.useCodeSnippetsOnFunctionSuggest": false,
 	"go.formatOnSave": true, 
--- a/package.json
+++ b/package.json
@@ -420,6 +420,11 @@
           "default": null,
           "description": "Specifies the GOROOT to use when no environment variable is set."
         },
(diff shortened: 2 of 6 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

dmage:
As I can see there is no way to use same command and same handler for all three cases (the output of `go test -c -cover` differs from the output of `go test -c`/`go build`). I think it will be better to leave buildOnSave handler as is, later it can be replaced by go-langserver.

dmage:
Thank you, I didn't expect you to merge this without any changes :)

Please take a look at https://github.com/dmage/vscode-go/commit/d4a9ff39cdd291ba85702342e827af7dcfb2d6e2

Now I think showHideTestStatus is unnecessary. There is no reason to hide test status when a tab switched. And test status should be invalidated if any file was changed (it may be a fixture, for example).

ramya-rao-a:
To be honest I wanted to include this feature in the next update which I wanted to release in a day or two. Didn't know if that was enough time for you to respond to any additional refactoring comments. So I thought I'd do it myself :)

Thanks for the update, that is definitely more cleaner. Feel free to send a PR, and I'll get it merged right away :)

PR #820, opened 2017-02-26
Use default gopath supported in Go 1.8

### LINKED ISSUE #813: Add support for default GOPATH (Go 1.8)
Go 1.8 introduced a [default GOPATH](https://beta.golang.org/doc/go1.8#gopath) that is used when users do not set one themselves. That means that `go env` will return a valid GOPATH even when the GOPATH environment variable is not set. As such, the [isGoPathSet](https://github.com/Microsoft/vscode-go/blob/master/src/util.ts#L213) function returns a false negative in that scenario.

**Proposal:**
Instead of checking the GOPATH environment variable, use the output from the `go env GOPATH` command.

ramya-rao-a:
Fixes #813


src/goInstallTools.ts
```diff
@@ -220,10 +220,28 @@ export function updateGoPathGoRootFromConfig() {
...
+	}
+
+	// From Go 1.8 onwards, when there is no GOPATH set, there is default GOPATH used which can be got from running `go env`
+	let goRuntimePath = getGoRuntimePath();
+	return new Promise<void>((resolve, reject) => {
+		cp.execFile(goRuntimePath, ['env'], (err, stdout, stderr) => {
+			if (err) {
+				return reject();
+			}
+			let gopathOutput = stdout.split('\n').find((value, index) => { return value.startsWith('GOPATH="') && value.endsWith('"'); });
```
roblourens: Maybe could leave a \r after the split - you can split on `/\r\n|\n/` or trim

ramya-rao-a: Good catch, done

PR #821, opened 2017-02-26
Show test status until a file is changed

dmage:
Small addition for #810.

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -21,34 +21,11 @@ let statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.
 statusBarItem.command = 'go.test.showOutput';
 
 export function removeTestStatus(e: vscode.TextDocumentChangeEvent) {
-	if (e.document.isUntitled || e.document.languageId !== 'go') {
+	if (e.document.isUntitled) {
 		return;
 	}
 	statusBarItem.hide();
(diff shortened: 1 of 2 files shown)


dmage:
Ugly fix for CI: Microsoft/vscode-extension-vscode#52

ramya-rao-a:
What does this have to do with Microsoft/vscode-extension-vscode#52?

dmage:
I'm about https://travis-ci.org/Microsoft/vscode-go/builds/205602472
Tests for this PR are red without changes in vscode-extension-vscode.

ramya-rao-a:
All the Mac related tests were previously failing to connect to github: https://travis-ci.org/Microsoft/vscode-go/builds/205432409, so I wrongly assumed the current failure is due to the same reason.

vscode-extension-vscode hasn't been updated in months. Why do you think that the current failure `'base must be a non-empty string, or null/undefined.'` is related to vscode-extension-code?

dmage:
https://github.com/ddliu/gulp-remote-src/commit/02953a1429023bffa418b4174a73282faa4018d9
This is the reason why tests will fail on Mac since yesterday.
I was able to reproduce this error on my macbook and found that vscode-extension-vscode uses empty string as a base.

ramya-rao-a:
Thanks @dmage, looks like your PR to vscode-extension-vscode was merged.

PR #824, opened 2017-02-28
Log all settings in a telemetry event

ramya-rao-a:


src/goMain.ts
```diff
@@ -287,3 +290,32 @@ function startBuildOnSaveWatcher(subscriptions: vscode.Disposable[]) {
 
 }
 
+function sendTelemetryEventForConfig(goConfig: vscode.WorkspaceConfiguration) {
+	sendTelemetryEvent('goConfig', {
+		buildOnSave: goConfig['buildOnSave'] ? 'true' : 'false',
```
roblourens: Or like `goConfig['buildOnSave'] + ''` for these if they have to be a string.

ramya-rao-a: Good tip, done

Issue #589, opened 2016-11-01
Add setting to lint whole workspace rather than just the current package


atombender:
Example steps:

1. Edit `foo.go` (package `bar`) with some syntax error.
2. Panel correctly shows error.
3. Edit `foo_test.go` (package `bar_test` in same folder) with some syntax error and save.
4. Fix error in `foo_test.go` and save.
5. Error panel is confusingly empty, even though `foo.go` still has syntax errors.
6. Switch to `foo.go`. Panel stays empty.
7. Make a dummy modification and save file.
8. Panel now correctly shows errors for `foo.go`.
9. That's terrible.

In other words, the panel only shows errors for current package, and requires a dummy modification + re-save of a failing file in order to re-display errors if it i's in a different package from the one that previously failed.

More generally, I want VSCode to behave like any IDE: Show errors across my entire project.

For example: Let's say I rename a field in a public struct that's used by other packages. I want to see _all subsequent errors_ from _all packages in my project_ that are using that struct and no longer work. I don't want to build the project to see those errors, and I don't want to randomly save files just to trigger linting.

Surely this must be an extremely common desire.

(Apologies if there's an existing issue for this; I couldn't find it.)

ramya-rao-a:
Funnily enough, there is an issue asking for linting ONLY in current package
#505 

`golint` does have the option of linting a whole directory. Would adding a setting to lint the whole workspace work for you?

atombender:
Yeah, I don't understand that issue.

So I just realized I was using `golint` instead of `gometalinter`. When I enable the latter, I seem to be getting workspace-wide errors. Maybe close this, then.

ramya-rao-a:
That's interesting, I didnt know that.

If `gometalinter` does the job, and you don't mind using it (it runs a bunch of linters including `golint`) then yes, go ahead and close this.

atombender:
Thanks! It would be a good idea to document this behaviour. I find `golint` quite limited, myself.

ramya-rao-a:
@atombender Are you sure you are getting workspace-wide errors even when your workspace has multiple packages?

I tried `gometalinter` on a workspace with multiple packages, and I see linting errors (on save) scoped to just the package to which the current file that is open belongs to

atombender:
I spoke too soon. I'm actually seeing the same behaviour as described originally.

[...]

Issue #816, opened 2017-02-23
Code Coverage for multiple packages


ajackson-cpi:
'go test cover' doesn't consider coverage via other packages' tests. 
Please make it possible to use coverage tools that do, like:
 https://github.com/haya14busa/goverage

mattetti:
I would personally recommend to keep it simple for now. Also, the go recommended way is to have the tests with the code, what are the scenarios where you'd write tests in other packages that the original one?

ajackson-cpi:
I'm building a simple REST server that reads from a database. Instead of reinventing DB processing for each call, most of the logic is in a "db-helpers" package. 

To test both packages, I would need to copy-paste tests to both. 
It would be far cleaner if the API tests could indicate that they tested the db-helpers package.

Gaffney:
Agreed, this is annoying... it has resulted in writing tests for the sake of coverage even though in reality the code in question is indirectly covered all over the place.  Really slows down productivity and is an obstacle towards refactoring when you know you not only have to extract out all the common code, but you have to pull out all the common tests as well.

ramya-rao-a:
We could add support for https://github.com/haya14busa/goverage, but keep it behind a setting, so that this a feature that users opt in to and is not the default one.

AlekSi:
Go 1.10 supports `go test -cover` for multiple packages: https://golang.org/doc/go1.10#test Extending existing code to pass `-coverpkg=all` should be easy, no external tool is needed now.

ramya-rao-a:
@mattetti, @AlekSi Could either of you try passing the `-coverpkg=all` flag via the `go.testflags` setting and share if any extra work is needed from the extension to achieve this?

jirfag:
@ramya-rao-a I've added `"go.testFlags": ["-coverpkg=all"],` to settings, tests are running with this flag, but coverage is shown only in the tested package.

ShoshinNikita:
> @ramya-rao-a I've added `"go.testFlags": ["-coverpkg=all"],` to settings, tests are running with this flag, but coverage is shown only in the tested package.

I found the reason why coverage isn't shown. The bug is in `applyCodeCoverageToAllEditors` function in `src/goCovers.ts` file

## Setup

Let's use this package structure (package is `github.com/user/repo`, root is `/home/usr/go/src/github.com/user/repo`)

```
|-- main.go
|-- first
    |-- first.go
|--second
    |-- second.go
    |-- second_test.go
```

[...]

PR #830, opened 2017-03-02
Add basic multi-package mode

mattblagden:
Add an option to run lint/build/test on the entire workspace.

This is a very basic implementation that checks for the `go.multiPackage` flag when inspecting the code. If in multi-package mode:

- The working directory is set to the workspace root instead of the current file's directory
- Build options that aren't supported for multiple packages are omitted
- The target is switched from `.` to `./...`

See issue #589

### DIFF
--- a/package.json
+++ b/package.json
@@ -484,6 +484,11 @@
           "type": "boolean",
           "default": false,
           "description": "If false, the import statements will be excluded while using the Go to Symbol in File feature"
+        },
+        "go.multiPackage": {
+          "type": "boolean",
+          "default": false,
+          "description": "Invoke build/lint/test commands recursively from the workspace root."
         }
       }
     }
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -107,7 +107,8 @@ function runTool(args: string[], cwd: string, severity: string, useStdErr: boole
 export function check(filename: string, goConfig: vscode.WorkspaceConfiguration): Promise<ICheckResult[]> {
 	outputChannel.clear();
 	let runningToolsPromises = [];
-	let cwd = path.dirname(filename);
+	let multiPackage = !!goConfig['multiPackage'];
+	let cwd = multiPackage ? vscode.workspace.rootPath : path.dirname(filename);
 	let goRuntimePath = getGoRuntimePath();
 
 	if (!goRuntimePath) {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -484,6 +484,11 @@
           "type": "boolean",
           "default": false,
           "description": "If false, the import statements will be excluded while using the Go to Symbol in File feature"
+        },
+        "go.multiPackage": {
+          "type": "boolean",
```
ramya-rao-a: Currently, this flag affects build, lint, vet and test. There might be cases when users might want just a few of these to run on whole workspace. Example: There might be a lot of tests in the workspace and they might take time, so user might not want test to run on whole workspace but only build.

So instead of just having this setting as a simple boolean, we can have it as an object
```
"go.multiPackage": {
          "type": "object",
          "default": {
            "build": false,
            "test": false,
            "lint": false,
            "vet": false
          },
          "description": "Invoke build/lint/test commands recursively from the workspace root."
        }
```

Thoughts?

package.json
```diff
@@ -484,6 +484,11 @@
...
+          "default": false,
+          "description": "Invoke build/lint/test commands recursively from the workspace root."
```
ramya-rao-a: vet is missing in the description

src/goCheck.ts
```diff
@@ -160,9 +165,19 @@ export function check(filename: string, goConfig: vscode.WorkspaceConfiguration)
...
-					args = ['test', '-copybinary', '-o', tmppath, '-c', '-tags', buildTags, ...buildFlags, '.'];
+					args = ['test', '-tags', buildTags, ...buildFlags];
+					if (multiPackage) {
+						args = args.concat('-run=^$');
+					} else {
+						args = args.concat('-copybinary', '-c');
+					}
+				}
+				if (multiPackage) {
+					args = args.concat('./...');
```
ramya-rao-a: coverOnSave is broken when multiPackage is set to true
To begin with, we can skip test from the multiPackage feature, and continue working on it in a separate PR

https://github.com/Microsoft/vscode-go/issues/816 talks about using test results from multiple packages for getting code coverage. We can use that to track multiPackage feature for test and code coverage

coccyx:
Would love to see this go in.  Looks like some work is remaining, but +1 from me for this feature.

ramya-rao-a:
Closing this in favor of #1023

Issue #834, opened 2017-03-03
When hitting Stop during Debug, it does not kill debug process


kisPocok:
"debug" process is not being killed after Stop button is pressed. I have to kill it manually.

vscode 1.10.1
go extension 0.6.54
macOS 10.12.3

Reference https://github.com/Microsoft/vscode-go/issues/438

fasa123:
Same problem here (on Linux).  
I assume it stems from having multiple goroutines being executed (a server running) and as thus a simple SIGTERM is not enough. 

Just force close the process by sending a SIGKILL. 

If you are debugging your application and are clicking on the Close button the expected behaviour should be a SIGKILL (-9) and not a SIGTERM (-15).

fasa123:
So turns out bash/sh cannot parse the file because of Windows line endings (it uses CRLF). Just change the line endings to LF and it works perfectly again on Linux (and should as well on Mac obviously). 

One thing I'd love to know is why a Linux specific file has Windows specific formatting.

ramya-rao-a:
This should be fixed with the latest version 0.6.55

kisPocok:
"Eureka, it works" :)

Issue #438, opened 2016-08-16
When hitting Stop during Debug, it does not always kill debug process


intdel:
I am running a web server in Go (1.6) with visual studio code and the most recent plugin:

```
func main() {

    http.HandleFunc("/data", htmlHandler)

    err := http.ListenAndServe(":8080", nil)
    if err != nil {
        log.Fatal(err)
    }
    os.Exit(0)
}
```

When I hit the stop button on the small debug window, it does not kill the webserver, because after I hit F5 again I get:

>  2016/08/16 15:34:13 server.go:71: Using API v1
> 2016/08/16 15:34:13 debugger.go:61: launching process with args: [./debug]
> API server listening at: 127.0.0.1:2345
> 2016/08/16 15:34:13 debugger.go:393: continuing
> 2016/08/16 15:34:13 listen tcp :8080: bind: address already in use

So I will have to look manually for the PID and kill it. Then it works again.

lukehoban:
What OS, and do you have a full precise repro?  I tried a few times on Windows and OSX with the closest full example I could derive from the above and couldn't reproduce this.

Happy to reopen if you have more info to investigate.

intdel:
Sure. I am using Debian Testing with Google Chrome:

```
lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux testing (stretch)
Release:	testing
Codename:	stretch
```

Here is an example code that I used to successfully verify the issue:
web.go:
```
package main

import "net/http"
import "fmt"
import "log"

func welcomeHandler(w http.ResponseWriter, r *http.Request) {

	fmt.Fprintln(w, "hello")
}

func main() {

	http.HandleFunc("/", welcomeHandler)

	err := http.ListenAndServe(":8080", nil)

	if err != nil {
		log.Fatal(err)
	}

}
```

launch.json:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "remotePath": "",
            "port": 2345,
            "host": "127.0.0.1",
            "program": "${workspaceRoot}",
            "env": {},
            "args": [],
            "showLog": true
        }
    ]
}
```

Steps to reproduce:

1. Press F5 to launch the debug process
2. Open terminal and do pidof debug:
```
$ pidof debug                                                                                                                                                       
5441     
```                          
\3. Open browser and go to http://localhost:8080. You will see "hello
[...]

PR #838, opened 2017-03-03
Fixed line endings for scripts/terminateProcess.sh

fasa123:
Since the file has Windows line endings bash/sh cannot execute the file under Linux which is illogical since it is only ran under Linux (/Mac). Changed the endings from CRLF to LF now the debugger works again fine 

see https://github.com/Microsoft/vscode-go/issues/834 and https://github.com/Microsoft/vscode-go/issues/438


roblourens:
Are you sure you didn't just check out CRLF line endings? They are LF on my machine

fasa123:
Hi, 
Yes you are right the git source uses LF you can discard this pull request.

Though the extension that you install from the extension explorer of vscode appearently uses a version which uses Windows endings. The pull request is from my ~/.vscode/extensions and that one used CRLF which is also why so many people did have this problem.  Just update that one and everything will work.

roblourens:
@fasa123 I see it now, yeah it was published with CRLF endings. But when I fix that locally, it actually still reproes for me...

kisPocok:
I've also fixed the CRLF ending locally but the bug still there.

roblourens:
Actually, it wasn't working when I just resaved the file with LF, but I deleted and recreated it with LF, and now it works. That's weird. Also, doing the opposite (running out of sources, changing LF to CRLF) broke it in the same way. So I think this is likely the problem.

kisPocok:
The current master version works for me. No errors so far.

roblourens:
@fasa123 Did you also have to `chmod +x terminateProcess.sh`?

roblourens:
To make this work I have to chmod +x, and fix the line endings. When I publish an extension with a .sh from Mac, it comes out with the x flag automatically. Apparently that's not the case from Windows, so maybe we just have to always publish from mac/linux?

ramya-rao-a:
Just published from a Mac

@kisPocok @fasal123 @YCF Can you get the latest version and confirm if things work for you?

kisPocok:
Eureka, it works! 👍

YCF:
@ramya-rao-a 
It works well!

ramya-rao-a:
woohoo!

fasa123:
Yes I did have to chmod as well.
And yes the extension works now.

roblourens:
Great! Thanks for pointing out the line ending issue.

PR #854, opened 2017-03-10
Fix godoc command not found

netroby:


ramya-rao-a:
Can you add it here as well? https://github.com/Microsoft/vscode-go/blob/master/src/goInstallTools.ts#L33

netroby:
done. can you take a look ? @ramya-rao-a

src/goInstallTools.ts
```diff
@@ -25,7 +25,8 @@ function getTools(goVersion: SemVersion): { [key: string]: string } {
 		'go-outline': 'github.com/lukehoban/go-outline',
 		'go-symbols': 'github.com/newhook/go-symbols',
 		'guru': 'golang.org/x/tools/cmd/guru',
-		'gorename': 'golang.org/x/tools/cmd/gorename'
+		'gorename': 'golang.org/x/tools/cmd/gorename',
+		'godoc': 'golang.org/x/tools/cmd/godoc'
```
ramya-rao-a: This should go inside the if block `if (goConfig['docsTool'] === 'godoc')`, because when user choose `gogetdoc`, they need not install `godoc`

README.md
```diff
@@ -171,6 +172,7 @@ go get -u -v github.com/tpng/gopkgs
 go get -u -v github.com/newhook/go-symbols
 go get -u -v golang.org/x/tools/cmd/guru
 go get -u -v github.com/cweill/gotests/...
+go get golang.org/x/tools/cmd/godoc
```
ramya-rao-a: Missing `-u -v`

netroby:
@ramya-rao-a Fix as your request. thanks for guide me

ramya-rao-a:
Fix is out in the latest update (0.6.56) to the Go extension

Issue #868, opened 2017-03-16
Long strings in variable inspection get truncated


Sirikon:
When trying to debug a long string, it gets truncated, and trying to copy it using the right-click context menu, returns the actual truncated value, with "...74 more" string.

VSCode itself adds ellipsis in that element when it's too long, maybe should directly print here the full length variable. Or, at least, when using the "Copy value" context menu, copy the real (and full length) one.

![image](https://cloud.githubusercontent.com/assets/1157382/24012680/203c3736-0a7f-11e7-87cb-141b9f2390c7.png)

Seems like this happens because Delve returns a 'Len' prop shorter than the actual one, and this is like this by design: https://github.com/derekparker/delve/issues/659#issuecomment-256945278

michaelmello:
So it looks like vscode uses v1 of the delve API: https://github.com/derekparker/delve/blob/master/service/rpc1/server.go

Unfortunately, v1 does not allow the client to change the default length of the variable value, which is set to 64 bytes.
![image](https://cloud.githubusercontent.com/assets/1448193/24074271/b7c635a4-0bdb-11e7-89df-0a4ee9cea63f.png)
![image](https://cloud.githubusercontent.com/assets/1448193/24074274/cb3f011a-0bdb-11e7-92a5-7b0a8ed04545.png)
![image](https://cloud.githubusercontent.com/assets/1448193/24074292/21230dd8-0bdc-11e7-84fa-ac626c50de87.png)

As a workaround, we can simply change the source of delve to return more than 64 bytes.

1.) Navigate to %GOPATH%\src\github.com\derekparker\delve\service\rpc1\server.go
2.) Look for the defaultLoadConfig variable and change the third argument value of 64, which is the field: MaxStringLen. I set mine to 500.
![image](https://cloud.githubusercontent.com/assets/1448193/24074274/cb3f011a-0bdb-11e7-92a5-7b0a8ed04545.png)
3.) Compile and install your modified version of delve. Navigate to %GOPATH%\src\github.com\derekparker\delve\cmd\dlv, and then run the command: go install.

edburns:
Any idea why this hasn't been merged yet?

lggomez:
@edburns this is a hack in the delve debugger rather than a fix on the extension side

To fix this on the extension we have to use the delve v2 api call instead of the old one:
[...]

PR #870, opened 2017-03-16
Long variable values in debug mode get truncated

michaelmello:
Visual Studio Code truncates the values of local debug variables today, so they fit nicely in the sidebar.  vscode-go seems to apply additional measures on top of the default behavior for truncating variable values with string types.  As a result, we cannot access the full value of the variable if it exceeds the limit in place (appears to be 64: https://github.com/derekparker/delve/issues/659#issuecomment-256945278), while debugging.

What I've done here, is simply remove that limit and return the entire value of the variable that we're debugging.

#868

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -641,12 +641,8 @@ class GoDebugSession extends DebugSession {
 				variablesReference: this._variableHandles.create(v)
 			};
 		} else if (v.kind === GoReflectKind.String) {
-			let val = v.value;
-			if (v.value && v.value.length < v.len) {
-				val += `...+${v.len - v.value.length} more`;
-			}
 			return {
-				result: v.unreadable ? ('<' + v.unreadable + '>') : ('"' + val + '"'),
+				result: v.unreadable ? ('<' + v.unreadable + '>') : ('"' + v.value + '"'),
 				variablesReference: 0
 			};
 		} else {


roblourens:
@michaelmello I don't understand, it doesn't look like vscode-go is applying a limit here. It's just tacking on the warning if the string is truncated by delve.

Previously we return `v.value` + extra thing. Now you return `v.value`. How does this help?

Sirikon:
https://github.com/derekparker/delve/issues/659#issuecomment-256945278

"The default for maximum string length is 64, we picked this value so that variables would fit into a small terminal window, other clients should pick a value that suits them."

The fix should include increasing that value, the maximum string length. I think.

michaelmello:
@roblourens You are correct. Though this minor edit may still be necessary once the full value is returned.

After more research, I think the problem lies with this call:

image


The Delve API method of interest is: ListLocalVars: https://godoc.org/github.com/derekparker/delve/service/rpc2#RPCServer.ListLocalVars

image

It takes a ListLocalVarsIn: https://godoc.org/github.com/derekparker/delve/service/rpc2#ListLocalVarsIn

image

vs-code is sending it the scope but not the Cfg which contains MaxStringLen. 
image

image

Since Delve defaults the MaxStringLen to 64, we'll need to explicitly ask for everything. I believe this is where the problem lies but haven't a chance to test.

roblourens:
Would be great if you can figure out whether we can override that max length! Closing this for now.

PR #880, opened 2017-03-21
Use gomodifytags to add/remove tags from struct fields

### LINKED ISSUE #433: feature request: generate json tags
@fatih's excellent vim-go has a great feature allowing you to generate struct field tags. The behavior is defined in this section: https://github.com/fatih/vim-go-tutorial#snippets

Select a field, after the type, you type `json` and use tab completion and it generates a json tag.
This is quite an useful feature when creating a lot of structs that marshall to json.

ramya-rao-a:
Fixes #433

### DIFF
--- a/src/goModifytags.ts
+++ b/src/goModifytags.ts
@@ -1,0 +1,136 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------*/
+
+'use strict';
+
+import vscode = require('vscode');
+import { byteOffsetAt, getBinPath } from './util';
+import cp = require('child_process');
+import { promptForMissingTool } from './goInstallTools';
+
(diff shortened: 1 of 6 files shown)


src/goModifytags.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+function runGomodifytags(args: string[]) {
+	let gomodifytags = getBinPath('gomodifytags');
+	let editor = vscode.window.activeTextEditor;
+	let fileContents = editor.document.getText();
+	let input = editor.document.fileName + '\n' + fileContents.length + '\n' + fileContents;
+	let p = cp.execFile(gomodifytags, args, (err, stdout, stderr) => {
+		if (err && (<any>err).code === 'ENOENT') {
+			promptForMissingTool('gomodifytags');
```
roblourens: Can you detect whether the tool is installed earlier, before the user has entered tag names and everything?

ramya-rao-a: We could, but this is the model we follow for all other tools, so will not be changing this unless we change the rest of the places

PR #885, opened 2017-03-26
Fixes #241 Use absolute path in output pane to support links

### LINKED ISSUE #241: Goto line on double click of build/test output w/ file & line #
I have no idea if this is possible in VSCode (I would hope so)... but it's super handy to be able to just double click a build or test error in the output and jump to the right line in the editor so you can read the exact code.  The filename and line number are right there in the output, so hopefully that's possible.

ramya-rao-a:
Fixes #241

The output from `go test -v` in case of test failures look like this:

image

This PR is to expand the filename to its absolute path, so that it is clickable and acts as a link to the test file

### DIFF
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -244,7 +244,7 @@ function getTestFlags(goConfig: vscode.WorkspaceConfiguration, args: any): strin
 function expandFilePathInOutput(output: string, cwd: string): string {
 	let lines = output.split('\n');
 	for (let i = 0; i < lines.length; i++) {
-		let matches = lines[i].match(/^\t(\w+_test.go):(\d+):/);
+		let matches = lines[i].match(/^\t(\S+_test.go):(\d+):/);
 		if (matches) {
 			lines[i] = lines[i].replace(matches[1], path.join(cwd, matches[1]));
 		}
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -39,10 +39,6 @@ interface TestConfig {
 	 * Test was not requested explicitly. The output should not appear in the UI.
 	 */
 	background?: boolean;
-	/**
-	 * Path of the file that contains the tests being run. Doesn't apply when running all tests in package.
-	 */
-	filePath?: string;
 }
 
 // lastTestConfig holds a reference to the last executed TestConfig which allows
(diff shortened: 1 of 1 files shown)


src/goTest.ts
```diff
@@ -191,7 +191,11 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 		outputChannel.appendLine('');
 
 		let proc = cp.spawn(goRuntimePath, args, { env: testEnvVars, cwd: testconfig.dir });
-		proc.stdout.on('data', chunk => outputChannel.append(chunk.toString()));
+		proc.stdout.on('data', chunk => {
+			let testOutput = expandFilePathInOutput(chunk.toString(), testconfig.dir);
```
roblourens: Is there some option for this tool that could make it output absolute paths?

src/goTest.ts
```diff
@@ -236,3 +240,14 @@ function getTestFlags(goConfig: vscode.WorkspaceConfiguration, args: any): strin
 	let testFlags = goConfig['testFlags'] ? goConfig['testFlags'] : goConfig['buildFlags'];
 	return (args && args.hasOwnProperty('flags') && Array.isArray(args['flags'])) ? args['flags'] : testFlags;
 }
+
+function expandFilePathInOutput(output: string, cwd: string): string {
+	let lines = output.split('\n');
+	for (let i = 0; i < lines.length; i++) {
+		let matches = lines[i].match(/^\t(\w+_test.go):(\d+):/);
```
roblourens: You might not want to use `\w` here because it might not match all characters that are valid in a path, like `.`. Maybe `\S+` (non whitespace char) or just `.+` because it's searching between \t and `_test` already?

src/goTest.ts
ramya-rao-a: Not that I know of.

@rakyll @campoy Do you know if `go test -v` can print out absoulte path of the test file with failing tests?

src/goTest.ts
ramya-rao-a: updated to use `\S`

src/goTest.ts
campoy: I don't think there's an option for this, but if it's really needed we could try to add a PR

rakyll: I think it requires a proposal to merge such a feature. Have you tried the `-exec` flag? It allows you to intercept the go test with any program you want. You might be able to access to the test file's absolute path with a custom executable.

A sample exec program: https://github.com/golang/go/blob/master/misc/android/go_android_exec.go

ramya-rao-a: Thanks @rakyll and @campoy 

The current needs isn't worth the trouble of changing the current behavior.

We will go with expanding the file name from the extensions's side

joebehymer:
Did you ever know that you're my hero

PR #903, opened 2017-03-31
Add live error feedback as you type

### LINKED ISSUE #883: Feedback as you type
One of the things I miss about Sublime Text is the linter setup I had. I created a sublime linter package that called `gotype` to display errors, then changed `gotype` to accept the contents of a file via stdin, so the in-memory edited file could be used when linting. This allowed `gotype` to provide as-you-type feedback.

To enable this functionality, you simply call `gotype` exactly as you would before, but add `-lf <path/to/file.go>` and send the contents of that file via stdin.

I would love to have this functionality in `vscode-go`, so I am opening an issue here to get some discussion going around it. I'm happy to do the work myself, but I am not sure where to start.

The questions that would need to be answered in order for this to work are:

1. Can VSCode/vscode-go execute the linter automatically as you type, preferably after a set timeout?
2. Can VSCode/vscode-go send the path of the currently focused file as an argument to the command?
3. Can VSCode/vscode-go send the contents of the currently focused file via stdin to the command?

If these things are possible, then we should be able to integrate this lint-as-you-type functionality.

I suppose it would be best to also ask: would anyone else even be interested in this functionality? I find it extremely useful to get feedback as I type to catch all sorts of simple mistakes. I am able to code more quickly this way.

And, lastly, here is a link to [gotype-live](https://github.com/tylerb/gotype-live). (It may be a good idea to try and get this integrated into `gotype` proper, of course.)

tylerb:
This PR adds an optional (disabled by default) live error reporting system.

Closes #883

## Motivation

I've used many editors in which a live as-you-type feedback system was available. I've found that it helps me catch errors as they happen and fix them immediately. This constant feedback allows me to code more quickly, as I am still in the context of an error when I see it, rather than having to come back to it after the build has failed.

## Implementation

To implement this feature, I have taken advantage of my custom fork of [gotype](https://godoc.org/golang.org/x/tools/cmd/gotype), called [gotype-live](https://github.com/tylerb/gotype-live).

My fork enables the processing of an in-memory version of a file by accepting the contents via stdin. This allows us to get errors back from `gotype` as we code, rather than having to save the file first.

## Configuration

The default configuration for this feature, as of now, is as follows:

```json
"go.liveErrors": {
    "enabled": false,
    "delay": 500
}
```

The `delay` value configures a timer that waits for `n`ms after a keystroke before running the process. If a keystroke occurs during that window, the timer is reset.

## Example

Below is a short video showing the feature in action. This video was recorded with a 250ms delay time.

feedback

### DIFF
--- a/package.json
+++ b/package.json
@@ -542,16 +542,19 @@
           "description": "Tags and options configured here will be used by the Add Tags command to add tags to struct fields. If promptForTags is true, then user will be prompted for tags and options. By default, json tags are added."
         },
         "go.liveErrors": {
-          "enabled": {
+          "type": "object",
+          "properties": {
+            "enabled": {
             "type":"boolean",
             "default":false,
             "description": "If true, runs gotype on the file currently being edited and reports any semantic or syntactic errors found."
-          },
-          "delay": {
-            "type":"number",
-            "default":500,
-            "description": "The number of milliseconds to delay before execution. Resets with each keystroke."
-          },
+            },
+            "delay": {
+              "type":"number",
+              "default":500,
+              "description": "The number of milliseconds to delay before execution. Resets with each keystroke."
+            }
+          },          
           "default": {
             "enabled": false,
             "delay": 500
--- a/package.json
+++ b/package.json
@@ -559,7 +559,7 @@
             "enabled": false,
             "delay": 500
           },
-          "description": "Tags and options configured here will be used to configure how the live errors system behaves."
+          "description": "Use gotype on the file currently being edited and report any semantic or syntactic errors found after configured delay."
         },
         "go.removeTags": {
           "type": "object",
--- a/package.json
+++ b/package.json
@@ -545,13 +545,13 @@
           "type": "object",
           "properties": {
             "enabled": {
-            "type":"boolean",
-            "default":false,
-            "description": "If true, runs gotype on the file currently being edited and reports any semantic or syntactic errors found."
+              "type": "boolean",
+              "default": false,
+              "description": "If true, runs gotype on the file currently being edited and reports any semantic or syntactic errors found."
             },
             "delay": {
-              "type":"number",
-              "default":500,
+              "type": "number",
+              "default": 500,
               "description": "The number of milliseconds to delay before execution. Resets with each keystroke."
             }
           },          
--- a/package.json
+++ b/package.json
@@ -541,6 +541,23 @@
           },
           "description": "Tags and options configured here will be used by the Add Tags command to add tags to struct fields. If promptForTags is true, then user will be prompted for tags and options. By default, json tags are added."
         },
+        "go.liveErrors": {
+          "enabled": {
+            "type":"boolean",
+            "default":false,
+            "description": "If true, runs gotype on the file currently being edited and reports any semantic or syntactic errors found."
+          },
+          "delay": {
+            "type":"number",
+            "default":500,
+            "description": "The number of milliseconds to delay before execution. Resets with each keystroke."
+          },
+          "default": {
+            "enabled": false,
+            "delay": 500
+          },
+          "description": "Tags and options configured here will be used to configure how the live errors system behaves."
+        },
         "go.removeTags": {
           "type": "object",
           "properties": {
--- a/src/goLiveErrors.ts
+++ b/src/goLiveErrors.ts
@@ -40,7 +40,7 @@ export function parseLiveFile(e: vscode.TextDocumentChangeEvent) {
 }
 
 export function goLiveErrorsEnabled() {
-	return <GoLiveErrorsConfig>vscode.workspace.getConfiguration('go')['liveErrors'].enabled;	
+	return <GoLiveErrorsConfig>vscode.workspace.getConfiguration('go')['liveErrors'].enabled;
 }
 
 // processFile does the actual work once the timeout has fired
@@ -76,9 +76,9 @@ function processFile(e: vscode.TextDocumentChangeEvent) {
 		// we want to keep all non-error level diagnostics that were previously
 		// reported, so add them back in
 		oldDiagnostics.forEach((value) => {
-			if (value.severity != vscode.DiagnosticSeverity.Error) {
-				newDiagnostics.push(value)
-			} 
+			if (value.severity !== vscode.DiagnosticSeverity.Error) {
+				newDiagnostics.push(value);
+			}
 		});
 
 		if (err) {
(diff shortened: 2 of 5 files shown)


tylerb:
@ramya-rao-a I would greatly appreciate a code review, as I have never worked in typescript before. I want to make sure my code is clean and correct. 😄

ramya-rao-a:
@tylerb Great job! Your typescript is spot on :)

I have a left a few comments, have a look.

tylerb:
@ramya-rao-a 

I've made the changes you requested.

Regarding the changes for preserving existing diagnostics, here is the approach I've used:

1. Load existing diagnostics for this file
2. Delete existing diagnostics for this file
3. Iterate through the previous diagnostics and re-add any non-error level diagnostics
4. Iterate through errors reported by `gotype-live`, if any, and add them to the array
5. Set the new diagnostic array into the `diagnosticCollection`

The goal is to preserve all non-error level diagnostics, because `gotype-live` should be reporting any error-level diagnostics that we'd see during a build, so we are safe to clear those out. This lets us avoid duplicate reports while preserving diagnostics that already existed.

 Please review and let me know if you need anything else.

src/goLiveErrors.ts
```diff
@@ -0,0 +1,100 @@
...
+// semantic errors and reports them immediately
+export function parseLiveFile(e: vscode.TextDocumentChangeEvent) {
+	if (e.document.isUntitled) {
+		return;
+	}
+	if (e.document.languageId !== 'go') {
+		return;
+	}
+
+	let config = <GoLiveErrorsConfig>vscode.workspace.getConfiguration('go')['liveErrors'];
```
ramya-rao-a: why not re-use `goLiveErrorsEnabled()` ?

src/goLiveErrors.ts
```diff
@@ -0,0 +1,100 @@
...
+		clearTimeout(runner);
+	}
+	runner = setTimeout(function(){
+		processFile(e);
+		runner = null;
+	}, config.delay);
+}
+
+export function goLiveErrorsEnabled() {
+	return <GoLiveErrorsConfig>vscode.workspace.getConfiguration('go')['liveErrors'].enabled;
```
ramya-rao-a: add a null/undefined check for `vscode.workspace.getConfiguration('go')['liveErrors']` before accessing the `enabled` property

src/goLiveErrors.ts
```diff
@@ -0,0 +1,100 @@
...
+
+		// grab a copy of the existing diagnostics that are being reported for the
+		// current file
+		let oldDiagnostics = diagnosticCollection.get(uri);
+
+		// delete the existing diagnostics for the current file
+		//
+		// error-level diagnostics will be reported by this process, so we want to
+		// clear out the existing errors to avoid getting duplicates
+		diagnosticCollection.delete(uri);
```
ramya-rao-a: > gotype-live should be reporting any error-level diagnostics that we'd see during a build

If the above assumption is indeed true, I'd still suggest having 2 separate diagnostic collections.

One for warnings (this will be fed by the linting and vetting), the other for errors (this will be fed by build and live errors). 

This way you don't have to copy the warnings from currentDiagnostics. 
The rest of your code here (delete diagnostics for uri etc.) can remain

src/goLiveErrors.ts
ramya-rao-a: Also, I think that when auto save is enabled, we shouldn't be enabling live errors. That would just be redundant work being done by build and gotype.

Can you add that check here?

src/goLiveErrors.ts
ramya-rao-a: I am not too picky about this one though. Perf wise, it shouldn't be a problem because you are only touching diagnostics for a single file. But it just feels wrong (gut wise) to be copying array items everytime liveError is run

package.json
```diff
@@ -541,6 +541,26 @@
...
+              "type":"number",
+              "default":500,
+              "description": "The number of milliseconds to delay before execution. Resets with each keystroke."
+            }
+          },          
+          "default": {
+            "enabled": false,
+            "delay": 500
+          },
+          "description": "Tags and options configured here will be used to configure how the live errors system behaves."
```
tylerb: Hmm.. I don't see an error. What am I missing? :)

src/goLiveErrors.ts
tylerb: I use `config.delay` a few lines below, otherwise I would have reused that function, yeah.

src/goLiveErrors.ts
tylerb: I tried using two separate diagnostics, but when I called `.set` on the `live` diagnostic object, it erased the warning diagnostics that were there. It appeared to me that you can only have one diagnostic object in use, per file, at a time.

I may have done something wrong, though. Is it possible for me to use two separate diagnostic objects on the same file and the same time and the diagnostics from *both* objects will be displayed?

src/goLiveErrors.ts
tylerb: I'm re-using it now, since we are also checking for `autoSave`.

package.json
ramya-rao-a: Oh I meant that the description needs an update :)

ramya-rao-a:
This is great work @tylerb!

I just added 1 commit with some description update, edit to README etc.

Will merge as soon as the tests are green.

tylerb:
Thanks! And thanks for cleaning up that last bit. I appreciate your feedback and help on this!

tylerb:
Looks like the gometalinter test failed. I can't retry, though, it seems.

src/goLiveErrors.ts
```diff
@@ -0,0 +1,88 @@
...
+
+let runner;
+
+export function goLiveErrorsEnabled() {
+	let goConfig = <GoLiveErrorsConfig>vscode.workspace.getConfiguration('go')['liveErrors'];
+	if (goConfig === null || goConfig === undefined || !goConfig.enabled) {
+		return false;
+	}
+	let autoSave = vscode.workspace.getConfiguration('files')['autoSave'];
+	if (autoSave !== null && autoSave !== undefined && autoSave !== 'off') {
```
23doors: I know it is closed but to be honest, I don't see a reason why autosave is checked at all. I use autosave onFocusChange and why would it be clashing with a linter running after some delay? Even autosave afterDelay makes little sense to check as it may just be some much higher value than linter delay.

ramya-rao-a: I was erring on the side of caution.

Good point on the fact that auto save onFocusChange and onWindowChange has no affect whatsoever.

I was concerned with the autosave on delay. Then you have build and gotype both running and giving essentially similar results and overriding each other.

On second thoughts, it is definitely being over cautious. We can remove the check in the next update.

23doors: Thanks!

keithballdotnet:
@tylerb just wanted to come along and say that this is a great feature.  Love it!!!  Thank you!!

tylerb:
@keithballdotnet glad you're enjoying it! You're most welcome!

PR #904, opened 2017-04-01
set GOPATH for debug adaptor automatically

esjeon:
This patch modifies `go.debug.startSession` command to add  environment variable 'GOPATH' to launch configuration JSON object. A proper `GOPATH` is searched using `getBinPath`, which utilizes VSCode settings.

This allows more versatile configuration and provides better out-of-box experience.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -196,6 +196,14 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				'program': '${file}'
 			});
 		}
+
+		if (!config.env || !config.env['GOPATH']) {
+			let binpath = getBinPath('dlv');
+			config.env = config.env || {};
+			if (binpath != 'dlv')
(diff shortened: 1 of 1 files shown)


src/goMain.ts
```diff
@@ -196,6 +196,14 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				'program': '${file}'
 			});
 		}
+
+		if (!config.env || !config.env['GOPATH']) {
+			let binpath = getBinPath('dlv');
```
ramya-rao-a: have you tried `config.env['GOPATH'] = process.env['GOPATH']` ?

ramya-rao-a:
@esjeon Good catch in finding the bug 

I have updated your branch with a simpler way of getting the GOPATH.

Will merge the PR as soon as the tests are green

PR #906, opened 2017-04-02
Add goenv support.

yageek:
Developing standard Go app and AppEngine GoApp is quite common.
Using the [goenv](https://github.com/crsmithdev/goenv) tool is quite usefull to switch between golang version.

`goenv` is using the same strategy as the well known `rbenv` and
`pyenv` and include the different versions in $HOME. It also
aliases the `go` runtime binary to `GOENV_ROOT/shims/go`.

Using `goenv` was preventing the extension to find the golang binary.


yageek:
This is unnecessary. It would be better to fix this on `goenv` part.

PR #909, opened 2017-04-03
Fixes #825 Support ${workspaceRoot} and ~ in --config flag

### LINKED ISSUE #825: gometalinter with relative config path
Is there a way to setup the gometalinter within vscode with a relative path to the config file?

When i use

```
  "go.lintTool": "gometalinter",
  "go.lintFlags": [
    "--config=golint.json"
  ],
```
then the relative path is not resolved from the project root and the file can not be found. Ofc. it works with an absolute path but i'd like to avoid that.

ramya-rao-a:


uchoa:
Is there a way to avoid it from running tests in the vendor/ directory?

PR #923, opened 2017-04-13
Fix gotype-live errors showing in wrong file

tylerb:
There exists an issue with the live error system in that `gotype-live` will report errors for all files in the package, but the document change event is only for a single file. As a result, all the errors, for all the files, were being displayed as errors in the current file. This PR fixes that issue.

It may be worthwhile to investigate the possibility of displaying those errors for the correct file.

### DIFF
--- a/src/goLiveErrors.ts
+++ b/src/goLiveErrors.ts
@@ -51,7 +51,6 @@ export function parseLiveFile(e: vscode.TextDocumentChangeEvent) {
 
 // processFile does the actual work once the timeout has fired
 function processFile(e: vscode.TextDocumentChangeEvent) {
-	let uri = e.document.uri;
 	let gotypeLive = getBinPath('gotype-live');
 	let fileContents = e.document.getText();
 	let fileName = e.document.fileName;
@@ -62,12 +61,12 @@ function processFile(e: vscode.TextDocumentChangeEvent) {
 			return;
 		}
 
-		errorDiagnosticCollection.delete(uri);
+		errorDiagnosticCollection.clear();
 
 		if (err) {
 			// we want to take the error path here because the command we are calling
 			// returns a non-zero exit status if the checks fail
-			let diagnostics = [];
+			let diagnosticMap: Map<string, vscode.Diagnostic[]> = new Map();
 
 			stderr.split('\n').forEach(error => {
 				if (error === null || error.length === 0) {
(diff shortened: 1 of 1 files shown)


tylerb:
@ramya-rao-a I've just updated the PR to parse the errors from all the files and add them to the appropriate diagnostic set. This allows the live error system to now clear all errors from the errors diagnostic collection and replace them with the live errors coming from gotype.

ramya-rao-a:
>  gotype-live will report errors for all files in the package

So in that case, shouldn't we refrain from clearing the errors for files from other packages?

Also, lets enable the live error feature when `files.autoSave != 'afterDelay' || files.autoSaveDelay > liv-error delay *1.5` and listen for feedback on how that feels for users

tylerb:
Apologies. I misspoke. It will report errors for all files as though you were attempting a build. The messages will be slightly different, but you will get all the same errors (at least in my experience).

As such, I believe it is safe to clear all errors and let gotype repopulate them as we go.

I'll get that change made also.

tylerb:
@ramya-rao-a as requested, live errors now works under the specified autoSave conditions.

Please let me know if you need anything else before merging. Thanks!

src/goLiveErrors.ts
```diff
@@ -20,8 +20,11 @@ export function goLiveErrorsEnabled() {
 	if (goConfig === null || goConfig === undefined || !goConfig.enabled) {
 		return false;
 	}
-	let autoSave = vscode.workspace.getConfiguration('files')['autoSave'];
-	if (autoSave !== null && autoSave !== undefined && autoSave !== 'off') {
+	let files = vscode.workspace.getConfiguration('files');
+	let autoSave = files['autoSave'];
+	let autoSaveDelay = files['autoSaveDelay'];
+	if (autoSave !== null && autoSave !== undefined &&
```
ramya-rao-a: `autoSave !== 'afterDelay' || autoSaveDelay < goConfig.delay * 1.5` would be simpler

src/goLiveErrors.ts
```diff
@@ -62,26 +64,33 @@ function processFile(e: vscode.TextDocumentChangeEvent) {
...
+				if (!diagnostics) {
+					diagnostics = [];
+				}
 				diagnostics.push(diagnostic);
+				diagnosticMap.set(file, diagnostics);
 			});
 
-			errorDiagnosticCollection.set(uri, diagnostics);
+			diagnosticMap.forEach((diagnostics, file) => {
+				errorDiagnosticCollection.set(vscode.Uri.parse('file://' + file), diagnostics);
```
ramya-rao-a: just curious... why add the prefix `file://` before parsing the file to get the uri? `vscode.Uri.parse` should work with just the file path as well

tylerb:
When I tried it without the prefix, it didn't work. It would report the errors, and I clicked on one, it would say the file could not be found.

ramya-rao-a:
thats interesting, i'll look into that separately.

Thanks for the fix!

I'll combine this one with a few other bug fixes and release in a few days

tylerb:
Sounds good! Thanks!

ramya-rao-a:
@tylerb I was testing this in Windows and the errors were not getting mapped to the right file. Probably due to issues with path separators while trying to parse `file:\\' + file`.

https://github.com/Microsoft/vscode-go/commit/08e9c10e0d232c8f7e07077e91f93c05d0e717c9 to fix the issue.

Just FYI, in case you came across the fix next time you were looking at this file

tylerb:
Excellent thanks! I don't have a windows box for testing. Sorry about that!

ramya-rao-a:
No problem 😊

Issue #632, opened 2016-11-19
Add go.goos to the extension options


hickeng:
Currently the only way I've found to set GOOS for the Go tools in vscode is to set the environment before launching the editor - this doesn't work well if you have multiple workspaces that may need different GOOS settings.

While I appreciate the prompt to ensure that the code at least compiles on multiple platforms, not all 3rd party packages do so and that causes real issues with guru and the other navigation tools.

ramya-rao-a:
@hickeng On the same line, wouldn't you also need GOARCH?
@mattetti You have any thoughts here?

mattetti:
It is a bit of an edge case but I agree that it would be a useful feature to have. @hickeng can you explain a bit more your edge case, are you trying to check that your code compiles properly against multiple os/arch targets?

hickeng:
@ramya-rao-a For completeness yes, but it's GOOS that's the most awkward currently.

@mattetti The multi-target case is definitely one that I hit, but it's not actually the one I experience on a daily basis.
I'm using Windows for my OS for various reasons but core portions of code I'm developing build only on Linux. It's easy enough to build that portion of the code in a VM or cross-compile but the Go tools are not happy in vscode. For example `godefs` cannot find call targets in the Linux only code. I have tried setting GOOS to linux in the past but never had it propagate to the tools unless I set it at a system level which causes other issues. I realize that I'd have issue navigating the Windows portions of the code if I change GOOS at a workspace level but the portions of the code that are not multi-platform are Linux so that's where the pain primarily presents.

ramya-rao-a:
Thanks @hickeng. This should be easy to do. 

PRs are welcome if anyone wants to take a crack at this.

Hints:
- Add a new configuration property in [package.json](https://github.com/Microsoft/vscode-go/blob/master/package.json#L255)
- Update [updateGoPathGoRootFromConfig](https://github.com/Microsoft/vscode-go/blob/master/src/goInstallTools.ts#L171) which updates GOPATH and GOROOT from settings to include updating the GOOS as well. Probably rename the method as well :)

ramya-rao-a:
This is fixed by https://github.com/Microsoft/vscode-go/pull/932.

The update containing the fix will be out soon

ramya-rao-a:
Fix is out in the latest update (0.6.60)

@hickeng Can you add your env vars in the new setting `go.toolsEnvVars` and confirm if your issue is fixed?

jackie-qiu:
[...]

PR #932, opened 2017-04-19
Set environment variables for go tools

bancek:
This will let users set env variables for `go build` (e.g. `CGO_CFLAGS`, `CGO_LDFLAGS` and `LD_LIBRARY_PATH`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -471,6 +471,11 @@
           "default": null,
           "description": "Flags to pass to `go test`. If null, then buildFlags will be used."
         },
+        "go.toolsEnvVars": {
+          "type": "object",
+          "default": {},
+          "description": "Environment variables that will passed to the processes that run the Go tools (e.g. CGO_CFLAGS)"
+        },
         "go.gocodeAutoBuild": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@bancek Why only build/lint/vet? Wouldn't other tools/features benefit from this as well?

If yes, then it would make more sense to update `process.env` with these env variables once during extension activations, and it will get used everywhere.

@hickeng This can help your GOOS scenario also I believe. #632

hickeng:
@ramya-rao-a Yes! Looking forward to the merge.

bancek:
@ramya-rao-a env variables are mostly needed for building.

I don't think it's a good idea to change `process.env` because it would effect other extensions and especially when config can be changed while VS Code is running.

ramya-rao-a:
@bancek See https://github.com/Microsoft/vscode-go/issues/632#issuecomment-273282269 for scenario where the env vars need to be used by some of the Go tools that we use the Go extension.

About your concerns on changes to `process.env` affecting other extensions: 

VS Code runs each extension in a separate process, and so any changes we do to the `process.env` is reflected only in the operations carried out by the Go extension.

See https://code.visualstudio.com/docs/extensionAPI/patterns-and-principles#_stability-extension-isolation

Issue #726, opened 2017-01-16
Feature request: always display references


mattetti:
This is probably a hard one but something that is super useful when writing C# in Visual Studio

The feature is defined here: https://code.visualstudio.com/docs/editor/editingevolved#_reference-information
![](https://code.visualstudio.com/images/editingevolved_referenceinfo.png)

VSCode supports this feature when asking for references of a specific value, but the data isn't cached and it isn't always displayed as in the image above.

https://code.visualstudio.com/Docs/extensionAPI/vscode-api#_a-namecodelensaspan-classcodeitem-id471codelensspan

redpixr:
++ this is like a must have time saver to me.

theSoenke:
This would be really awesome. Codelens is also now available for Typescript https://code.visualstudio.com/updates/v1_9#_typescript-references-codelens It's has been a very useful feature for me in C# in VS

ramya-rao-a:
Yes, it is a very good feature indeed. PRs are welcome.

theSoenke:
I've build a go codelens prototype. It shows references for functions, interfaces and structs. It takes some time to find all references, but i don't think there is much i can do about it. https://github.com/thesoenke/vscode-go/tree/codelens

![go-codelens](https://cloud.githubusercontent.com/assets/1757301/25073820/fd20c324-22ee-11e7-9bae-546e07cc8cdf.png)

One issue i ran into is that go-outline seems to return always 1 as the start character instead as of at least 6 for functions in a line, but for structs it's correct.

Another issue is that the console is filled with errors from guru when editing the code. Not sure what's the best solution here

ramya-rao-a:
@theSoenke great work! Please send a PR and we can try and get this in the next release :)

Few points
- I have a case where a function is used in another Go project outside the workspace opened in VS Code. The Codelens lists this external reference, but when I do a "Find all references", the external reference is not listed. Any idea how that happened? 
- About `go-outline` returning different offset for `func` vs `type`, looks like that is how the Go parser works. See https://github.com/lukehoban/go-outline/blob/master/main.go#L53
- About console filled with errors, codelens tries to update itself while you edit the file. And with every edit, the offset changes, but codelens still looks at the old offsets, thus causing the errors. Well, that's my current theory anyway, am digging deeper to see how other languages have done this and see if I can come up with a solution

ramya-rao-a:
@theSoenke 

About the errors in the console on edit, I see 2 types

[...]

PR #933, opened 2017-04-19
Add references codelens support

theSoenke:
This PR adds support for reference codelenses as discussed in https://github.com/Microsoft/vscode-go/issues/726

There are still a few issues:

- [x] Debug console shows guru errors when file contains syntax errors and is not saved
- [ ] Does not work with method receivers. Issue here is probably that the go parser returns 1 as a start character in a line for functions. It works for functions without receivers because i add an offset of 5 characters
- [ ] Slow for big projects. This is more an issue with guru

### DIFF
--- a/src/goCodelens.ts
+++ b/src/goCodelens.ts
@@ -1,0 +1,68 @@
+'use strict';
+
+import vscode = require('vscode');
+import { CodeLensProvider, SymbolInformation, SymbolKind, TextDocument, CancellationToken, CodeLens, Range, Command, Location, commands } from 'vscode';
+import { documentSymbols, GoDocumentSymbolProvider } from './goOutline';
+import { GoReferenceProvider } from './goReferences';
+
+export class GoCodeLensProvider implements CodeLensProvider {
+	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
+		let codelensEnabled = vscode.workspace.getConfiguration('go').get('referencesCodeLens.enabled');
+		if (!codelensEnabled) {
+			return Promise.resolve([]);
+		}
+
+		return this.provideDocumentSymbols(document, token).then(symbols => {
+			let symbolReferences = symbols.map(symbol => this.provideSymbolReferences(document, symbol, token));
+			return Promise.all(symbolReferences).then(values => {
+				let codelenses = [];
+				values.forEach(lens => {
+					if (lens) {
+						codelenses.push(lens);
+					}
+				});
+				return codelenses;
+			});
+		});
+	}
+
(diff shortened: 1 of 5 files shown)


src/goCodelens.ts
```diff
@@ -0,0 +1,66 @@
...
+    private provideSymbolReferences(document: TextDocument, symbol: SymbolInformation, token: CancellationToken): Thenable<CodeLens> {
+        if (token.isCancellationRequested) {
+            return Promise.resolve(null);
+        }
+
+        let options = {
+            includeDeclaration: false
+        };
+        let position = symbol.location.range.start;
+        if (symbol.kind === vscode.SymbolKind.Function) {
```
ramya-rao-a: Can you add a comment here as to why 5 is being added? Helps the next person reading this code :)

src/goReferences.ts
```diff
@@ -19,7 +19,7 @@ export class GoReferenceProvider implements vscode.ReferenceProvider {
 		});
 	}
 
-	private doFindReferences(document: vscode.TextDocument, position: vscode.Position, options: { includeDeclaration: boolean }, token: vscode.CancellationToken): Thenable<vscode.Location[]> {
+	public doFindReferences(document: vscode.TextDocument, position: vscode.Position, options: { includeDeclaration: boolean }, token: vscode.CancellationToken): Thenable<vscode.Location[]> {
```
ramya-rao-a: I am guessing the reason you are using `doFindReferences` instead of `provideReferences` is because the latter does a `saveAll` before finding the references.

My another guess for why `saveAll` was put in place was because `guru` was made to run on saved files.

Now that we have `guru` working on unsaved files as well (with the -modified flag), lets remove the `saveAll` that happens in `provideReferences` and you can use it here.

ramya-rao-a:
For the slowness issue, have you considered using [resolveCodeLens](https://code.visualstudio.com/docs/extensionAPI/vscode-api#CodeLensProvider.resolveCodeLens) ?

src/goCodelens.ts
theSoenke: Good point :)

src/goReferences.ts
theSoenke: Yes, saveAll was the reason for that. Removed `saveAll` and set `provideReferences` back to private

theSoenke:
Looks like resolveCodeLens could make a lot sense here. I will look into it.

ramya-rao-a:
@theSoenke I have merged the PR, but please continue looking at the `resolveCodeLens` option and send a different PR for it ?

theSoenke:
@ramya-rao-a Great! I've submitted a PR to use `resolveCodeLens` https://github.com/Microsoft/vscode-go/pull/938

PR #935, opened 2017-04-19
move env consolidation after loading envFile

### LINKED ISSUE #934: envFile does not load .env file into Delve
VSCode Version: 1.11.2
OS Version: macOS 10.12.3
Extension: Go 0.6.59

I have the following `launch.json` and content from `.test.env` is not loaded into Delve.

```
{
	"name": "Test",
	"type": "go",
	"request": "launch",
	"mode": "test",
	"remotePath": "${workspaceRoot}",
	"port": 2345,
	"host": "127.0.0.1",
	"program": "${workspaceRoot}/",
	"envFile": "${workspaceRoot}/.test.env",
	"env": { },
	"args": [ ]
}
```

goenning:
Works fine for me after I moved this :)
Fixes   #934

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -181,7 +181,7 @@ class Delve {
 		this.program = program;
 		this.remotePath = remotePath;
 		let mode = launchArgs.mode;
-		let env = launchArgs.env || {};
+		let env = Object.assign({}, launchArgs.env, process.env);
 		let dlvCwd = dirname(program);
 		let isProgramDirectory = false;
 		this.connection = new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -250,25 +269,6 @@ class Delve {
...
-				try {
-					const buffer = stripBOM(FS.readFileSync(launchArgs.envFile, 'utf8'));
-					buffer.split('\n').forEach( line => {
-						const r = line.match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/);
-						if (r !== null) {
-							let value = r[2] || '';
-							if (value.length > 0 && value.charAt(0) === '"' && value.charAt(value.length - 1) === '"') {
-								value = value.replace(/\\n/gm, '\n');
-							}
-							env[r[1]] = value.replace(/(^['"]|['"]$)/g, '');
```
ramya-rao-a: Actually, just changing the `env` in this line to `finalEnv` should also do the trick. Can you try that?

ramya-rao-a: Or we can get rid of `finalEnv` altogether and instead just replace that with 

`env = Object.assign({}, process.env, env)`

goenning: Great idea! I was focusing only on getting it done 😄 Can you see how it looks now?

goenning: By the way, I'm suggesting that we use the following precedence order 
`launch env -> file env -> process env` so we can override file env values with launch specific values.

What do you think? Check the last commit.

src/debugAdapter/goDebug.ts
```diff
@@ -181,7 +181,7 @@ class Delve {
 		this.program = program;
 		this.remotePath = remotePath;
 		let mode = launchArgs.mode;
-		let env = launchArgs.env || {};
+		let env = Object.assign({}, launchArgs.env, process.env);
```
ramya-rao-a: this way the `process.env` overrides `launchArgs.env` 
It should be the other way around

goenning: Check last commit, I changed it.

src/debugAdapter/goDebug.ts
ramya-rao-a: agreed on the order.

ramya-rao-a:
great, thanks @goenning!

Issue #879, opened 2017-03-21
Create launch configuration from selected Test function


dolmen:
In files named `*_test.go` a right click on a function named `Test*` with signature `func (t *testing.T)` would show a menu entry allowing to inject a Launch Configuration or to run it if it already exists.

ramya-rao-a:
Today you can use the configuration snippets in the launch.json file to add a configuration for debugging a test function. Just start typing "Go" and you will get suggestion items for config snippets.

It should be pretty straightforward to add a command that makes use of the config snippet, and plugs in the name of the test function under the cursor.

As far as debugging the test under the cursor, I think we can use the command "go.debug.startSession" and pass in the config

Am a little busy this week and the next week, so might not be able to get to this.
PRs are welcome though, and I am available to help out if anyone wants to take stab at this and needs help

ramya-rao-a:
With the latest update (0.6.60), you can now debug tests via the codelens.

This is for simple cases only.

If you have cases like env vars being set in your launch.json, the codelens will not be using it.

PR #937, opened 2017-04-20
RunTest with CodeLens

goenning:
Same feature as Test at Cursor, but in this case there is no need to select anything.

It's based on C# VSCode extension.

What you think about it?

I didn't provide any flag to disable it, because it's possible to do so with standard editor flag: `"editor.codeLens": false`.

npz78pjlim

### DIFF
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -27,10 +27,10 @@ export class GoRunTestCodeLensProvider implements CodeLensProvider {
 	private getCodeLensForPackage(document: TextDocument): Thenable<CodeLens[]> {
 		let documentSymbolProvider = new GoDocumentSymbolProvider();
 		return documentSymbolProvider.provideDocumentSymbols(document, null)
-				.then(symbols => symbols.filter(sym => sym.kind === vscode.SymbolKind.Package))
-				.then(pkgs => {
-					if (pkgs.length > 0 && pkgs[0].name) {
-						const range = pkgs[0].location.range;
+				.then(symbols => symbols.find(sym => sym.kind === vscode.SymbolKind.Package && !!sym.name))
+				.then(pkg => {
+					if (pkg) {
+						const range = pkg.location.range;
 						return [
 							new CodeLens(range, {
 								title: 'run package tests',
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -9,7 +9,7 @@ import vscode = require('vscode');
 import { CodeLensProvider, TextDocument, CancellationToken, CodeLens, Command } from 'vscode';
 import { getTestFunctions } from './goTest';
 
-export class GoCodeLensProvider implements CodeLensProvider {
+export class GoRunTestCodeLensProvider implements CodeLensProvider {
 	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
 		return getTestFunctions(document).then(testFunctions => {
 			return testFunctions.map(func => {
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -12,6 +12,10 @@ import { GoDocumentSymbolProvider } from './goOutline';
 
 export class GoRunTestCodeLensProvider implements CodeLensProvider {
 	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
+		if (!document.fileName.endsWith('_test.go')) {
+			return;
+		}
+
 		return Promise.all([
 			this.getCodeLensForPackage(document),
 			this.getCodeLensForFunctions(document)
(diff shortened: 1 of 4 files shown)


src/goRunTestCodelens.ts
```diff
@@ -0,0 +1,25 @@
...
+import { CodeLensProvider, TextDocument, CancellationToken, CodeLens, Command } from 'vscode';
+import { getTestFunctions } from './goTest';
+
+export class GoRunTestCodeLensProvider implements CodeLensProvider {
+	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
+		return getTestFunctions(document).then(testFunctions => {
+			return testFunctions.map(func => {
+				let showReferences: Command = {
+					title: 'run test',
+					command: 'go.test.function',
```
ramya-rao-a: Instead of creating a new command, I'd prefer re-using the existing `go.test.cursor`. 
You can pass the arguments as `{ functionName: func.name}.
And update `testAtCursor` to look for `args.functionName`. If it exists use it, else use the existing logic to find the test function under the cursor

src/goRunTestCodelens.ts
```diff
@@ -0,0 +1,25 @@
...
+ *--------------------------------------------------------*/
+
+'use strict';
+
+import vscode = require('vscode');
+import { CodeLensProvider, TextDocument, CancellationToken, CodeLens, Command } from 'vscode';
+import { getTestFunctions } from './goTest';
+
+export class GoRunTestCodeLensProvider implements CodeLensProvider {
+	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
```
ramya-rao-a: How about adding a codelens at the package level as well? Every go file has a `package` statement up top. So we can add a code lens there and map it to the `go.test.package` command

goenning:
Done. Also included test current file.

screen shot 2017-04-21 at 07 33 28

src/goRunTestCodelens.ts
```diff
@@ -0,0 +1,60 @@
...
+	private getCodeLensForPackage(document: TextDocument): Thenable<CodeLens[]> {
+		let documentSymbolProvider = new GoDocumentSymbolProvider();
+		return documentSymbolProvider.provideDocumentSymbols(document, null)
+				.then(symbols => symbols.find(sym => sym.kind === vscode.SymbolKind.Package && !!sym.name))
+				.then(pkg => {
+					if (pkg) {
+						const range = pkg.location.range;
+						return [
+							new CodeLens(range, {
+								title: 'run package tests',
```
c633: 2 cents suggestion: s/run/Run. The same for `run file tests`, `run test`.

ramya-rao-a: How about `Run tests in package`, `Run tests in file` and `Run test` ?

goenning: I personally prefer lowercase + shorter text to not get much attention. It should be as unobtrusive as possible so devs can focus on code an not on these actions. It's like this on [C# codelens](https://github.com/OmniSharp/omnisharp-vscode/blob/f92b9dfe378418e218aa62b305e8f438193a3f34/src/features/dotnetTest.ts#L172)

Next enhancement could be `debug test` 😄

ramya-rao-a: > as unobtrusive as possible
well, when you put it that way it makes sense. 

> Next enhancement could be debug test
There is already an ask for it https://github.com/Microsoft/vscode-go/issues/879 :)

ramya-rao-a:
This feature is now out in the latest update (0.6.60)
Thanks @goenning !

ironcladlou:
Cool feature! Any objection to a feature level config option? I can imagine wanting to disable this feature without disabling CodeLens generally.

ramya-rao-a:
For references codelens, we added the setting, because `guru` is known to be slow on Windows and people might want to turn it off, due to multiple `guru` spinning for long times.

For the run and debug tests, there is no downside perf wise and it doesnt hinder any workflow so didnt add the setting. (Also our settings are really getting big)

Any particular reason you would want to turn it off?

ironcladlou:
@ramya-rao-a 
>Any particular reason you would want to turn it off?

Perhaps the user only uses keyboard shortcuts for this and doesn't want an editor row occupied by the hints for every function. Vertical editor space is often precious.

goenning:
I guess it's up for discussion, but I share same feelings about the amount of settings the extension already has. 

You can disable it on workspace level, which should be roughly the same as a feature level setting. The only downside I can see is if your workspace has code from multiple languages and you want to disable codelens just for Go.

ramya-rao-a:
>  Vertical editor space is often precious.

`editor.codeLens` can be set to false for no codelens at all.

ramya-rao-a:
@ironcladlou See https://github.com/Microsoft/vscode-go/issues/964#issuecomment-299345928

coder543:
Personally, I think this feature is awesome. I hope we can have more codelens integration!

PR #939, opened 2017-04-22
Goimpl implementation

itsjamie:
A start of the goimpl implementation, pretty simple feature as it exists right now, just pop open the info box, take the information and then inject the stdout at the cursor position.

I see a few ways it could be improved, but figure it was better to just get the simple implementation done.

Idea for improvement
* Check if cursor is on a typedef, if so, allow input to only contain interface
* If cursor is not on a empty line, ensure to insert selection on first empty line below cursor

### DIFF
--- a/src/goImpl.ts
+++ b/src/goImpl.ts
@@ -1,0 +1,69 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------*/
+
+'use strict';
+
+import vscode = require('vscode');
+import cp = require('child_process');
+import { getBinPath } from './util';
+import { promptForMissingTool } from './goInstallTools';
+
+interface GoImplInput {
+	receiver: string;
+	interface: string;
+}
+
+// Supports only passing interface, see TODO in implCursor to finish
+const inputRegex = /^(\w+\ \*?\w+\ )?([\w.]+)$/;
+
+export function implCursor() {
+	let cursor = vscode.window.activeTextEditor.selection;
+	return vscode.window.showInputBox({
+		placeHolder: 'f *File io.Closer',
+		prompt: 'What interface do you want to implement?'
+	}).then(implInput => {
+		if (typeof implInput === 'undefined') {
+			return;
+		}
+		const matches = implInput.match(inputRegex);
+		if (!matches) {
+			vscode.window.showInformationMessage(`Not parsable input: ${implInput}`);
+			return;
+		}
+
+		// TODO: automatically detect type name at cursor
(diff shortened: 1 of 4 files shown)


itsjamie:
Just waiting for DocuSign email regarding CLA request.

ramya-rao-a:
Hey @itsjamie, I've been meaning to look at this PR since a while, but I've been swamped this week.
Will definitely review it early next week.

Thanks for your work!

src/goImpl.ts
```diff
@@ -0,0 +1,69 @@
...
+		if (typeof implInput === 'undefined') {
+			return;
+		}
+		const matches = implInput.match(inputRegex);
+		if (!matches) {
+			vscode.window.showInformationMessage(`Not parsable input: ${implInput}`);
+			return;
+		}
+
+		// TODO: automatically detect type name at cursor
```
ramya-rao-a: In this case, if you use the cursor to determine the type, then where would you propose to place the stub. End of the file?

src/goMain.ts
```diff
@@ -117,6 +118,10 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		removeTags(args);
 	}));
 
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.impl.cursor', () => {
+		implCursor();
```
ramya-rao-a: You need to add this command in `package.json` as well, so that the command shows up in the command pallet.

Am curious, how did you test this without the command showing up in the command pallet?

itsjamie: I registered the command to a key command in keybindings.json

src/goImpl.ts
itsjamie: I was thinking of right after the type. So it would need some detection to find the first empty line after the type.

src/goMain.ts
ramya-rao-a: Ah! thought so

src/goImpl.ts
ramya-rao-a: @fatih Where does vim-go paste the stub while taking the identifier under the cursor as receiver? Am referring to https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt#L665

fatih: @ramya-rao-a currently it just places the generated stub under the cursor. It's not in place. There is an open PR that aims to fix it: https://github.com/josharian/impl/pull/16

src/goMain.ts
itsjamie: Added this to the package.json in commit b640cc3

renannprado:
waiting for this to be merged :)

src/goImpl.ts
ramya-rao-a: Thanks @fatih

PR #945, opened 2017-04-26
showCovered/showUncovered

kode4food:
Adding two configuration options:
* `go.showCovered` (defaults to true) will decorate covered code in green
* `go.showUncovered` (defaults to true) will decorate uncovered code in red

### DIFF
--- a/package.json
+++ b/package.json
@@ -446,19 +446,10 @@
           "description": "Run 'go test' on save for current package. It is not advised to set this to `true` when you have Auto Save enabled."
         },
         "go.coverOnSave": {
-          "type": "boolean",
+          "type": ["boolean", "string"],
+          "enum": [true, false, "showCoveredOnly", "showUncoveredOnly"],
           "default": false,
           "description": "Run 'go test -coverprofile' on save"
-        },
-        "go.showCovered": {
-          "type": "boolean",
-          "default": true,
-          "description": "Decorate covered code"
-        },
-        "go.showUncovered": {
-          "type": "boolean",
-          "default": true,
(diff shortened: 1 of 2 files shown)


kode4food:
The one failing build configuration seems to be completely unrelated to the PR

package.json
```diff
@@ -450,6 +450,16 @@
           "default": false,
           "description": "Run 'go test -coverprofile' on save"
         },
+        "go.showCovered": {
```
ramya-rao-a: Am a little hesitant to add new settings. The list of settings we have is really blowing up.
How about re-using the existing setting for coverage.

```
"go.coverOnSave": {
   "type": ["boolean", "string"],
   "enum": [true, "showCoveredOnly", "showUncoveredOnly"],
   "default": true,
   "description": "..."
}
```

kode4food: @ramya-rao-a Just as well! I won't be able to get around to it until Friday though

ramya-rao-a: no problem, I am planning to release an update early next week. So Friday sounds good.

kode4food: I lied, just pushed the changes

ramya-rao-a:
@kode4food Since we have the `Go: Test coverage ..` command as well, having these options inside `go.coverOnSave` didnt seem right. So I am falling back to your initial proposal of having these options outside `coveronSave`. I am combining them (show covered/uncovered/both) to a single setting though.

See https://github.com/Microsoft/vscode-go/commit/9ddd21814a905f8d3edcbf93b38d698f6379eb29

PR #949, opened 2017-04-27
Add support for guru as a docs tool

stamblerre:
Support guru as a tool that can be used for jumping to definitions.


src/goDeclaration.ts
```diff
@@ -156,6 +158,50 @@ function definitionLocation_gogetdoc(document: vscode.TextDocument, position: vs
...
+				definitionInfo.file = match[1];
+				definitionInfo.line = +match[2] - 1;
+				definitionInfo.column = +match[3] - 1;
+				return resolve(definitionInfo);
+			} catch (e) {
+				reject(e);
+			}
+		});
+		let documentText = document.getText();
+		let documentArchive = document.fileName + '\n';
```
ramya-rao-a: you can use `getFileArchive()` from `util.ts` to do this

src/goDeclaration.ts
```diff
@@ -39,8 +39,10 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 		// Assume it's > Go 1.5
 		if (toolForDocs === 'godoc' || (ver && (ver.major < 1 || (ver.major === 1 && ver.minor < 6)))) {
 			return definitionLocation_godef(document, position, offset, includeDocs);
+		} else if (toolForDocs == 'guru') {
```
ramya-rao-a: There are linting errors that is breaking the build. 
- `==` should be `===` in line 42
- There is a trailing space in line 45

src/goSignature.ts
```diff
@@ -43,7 +43,7 @@ export class GoSignatureHelpProvider implements SignatureHelpProvider {
 				let funcName = declarationText.substring(0, nameEnd);
 				sig = declarationText.substring(sigStart);
 				si = new SignatureInformation(funcName + sig, res.doc);
-			} else {
+			} else if (res.toolUsed === 'gogetdoc') {
```
ramya-rao-a: Both Signature Help and Hover Info features are currently broken.
Until `guru` has the feature to provide docs, you can call `godoc` to provide the docs.

Pull https://github.com/Microsoft/vscode-go/blob/1302fcf0f704f66c974e043258b7ed9783ba104b/src/goDeclaration.ts#L85-L105 out into a function and use it in both `definitionLocation_godef` and `definitionLocation_guru`

stamblerre: Latest changes should fix this - if guru is set as the docsTool, I just passed in a config object with "godoc" set as the docsTool for provideHover and provideSignatureHelp.

PR #952, opened 2017-04-29
Add snippet for table tests

slomek:
Type `tdt` + tab:

	func Test(t *testing.T) {
		testCases := []struct {
			desc	string
			
		}{
			{
				desc: "",
				
			},
		}
		for _, tC := range testCases {
			t.Run(tC.desc, func(t *testing.T) {
				
			})
		}
	}

### DIFF
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -160,6 +160,10 @@
 			"prefix": "bf",
 			"body": "func Benchmark$1(b *testing.B) {\n\tfor ${2:i} := 0; ${2:i} < b.N; ${2:i}++ {\n\t\t$0\n\t}\n}"
 		},
+		"table driven test": {
+		"prefix": "tdt",
+			"body": "func Test$0(t *testing.T) {\n\ttestCases := []struct {\n\t\tdesc\tstring\n\t\t$1\n\t}{\n\t\t{\n\t\t\tdesc: \"$2\",\n\t\t\t$3\n\t\t},\n\t}\n\tfor _, tC := range testCases {\n\t\tt.Run(tC.desc, func(t *testing.T) {\n\t\t\t$4\n\t\t})\n\t}\n}"
+		},
 		"init function": {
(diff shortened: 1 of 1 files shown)


snippets/go.json
```diff
@@ -160,6 +160,10 @@
 			"prefix": "bf",
 			"body": "func Benchmark$1(b *testing.B) {\n\tfor ${2:i} := 0; ${2:i} < b.N; ${2:i}++ {\n\t\t$0\n\t}\n}"
 		},
+		"table tests": {
```
ramya-rao-a: Since the `test function` and the `benchmark function` use the prefix `tf` and bf` respectively, we should probably name this snippet `table test function` and use the prefix `tbd` instead?

slomek:
I'm not sure about naming it _table test function_ though, as it just adds the inside of a function. It is intentional since we already have _test function_ as you mentioned. It is intended to be used as in the example: first, create a function, rename it as you with, then inside it call this snippet and fill its fields.

ramya-rao-a:
Sorry, my mistake. I thought this was a complete test function with the table structure inside.

`tbd` was my Mac spellchecking what I originally intended which was `tbf` which on second thoughts is not that great :)

Any reason you want to keep this as a 2 step process. Use the existing snippet first and then this one?
You can combine the 2, name it "table driven test" and use the prefix "tdt" ?

slomek:
@ramya-rao-a I applied your suggestions ;)

ramya-rao-a:
@slomek The final stop $0 is at the function name, which doesnt seem right. The first thing to change in the snippet should be the name of the function.

I'll update the snippet accordingly
See https://github.com/Microsoft/vscode-go/commit/850e42e5304206eb1d21ad0ce2131bfbcbf85d8e

ramya-rao-a:
This is out in the latest update (0.6.60)

PR #971, opened 2017-05-06
add support for testEnvFile

goenning:
The main purpose of this pull request is to enable use of .env files when running/debugging tests through VSCode no matter what triggered it (CodeLens, Tasks, Command ...)

This pull request contains:

1) New `go.testEnvFile` configuration option with same intent as envFile from vscode debug settings;
2) Any test function ran by vscode-go will load env variables from: `process.env`, `go.toolsEnvVars`, `go.testEnvFile` and `go.testEnvVars`;
3) Debug CodeLens now loads `go.testEnvFile` and `go.testEnvVars`;
4) moved .env parser to a specific function for reusability purpose;

I didn't like this bit here `testEnvFile = testEnvFile.replace(/\${workspaceRoot}/g, vscode.workspace.rootPath);`, but didn't find any other way to do so.

Regards,
Guilherme

### DIFF
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -9,6 +9,7 @@ import vscode = require('vscode');
 import path = require('path');
 import { CodeLensProvider, TextDocument, CancellationToken, CodeLens, Command } from 'vscode';
 import { getTestFunctions } from './goTest';
+import { getTestEnvVars } from './util';
 import { GoDocumentSymbolProvider } from './goOutline';
 
 export class GoRunTestCodeLensProvider implements CodeLensProvider {
@@ -17,9 +18,7 @@ export class GoRunTestCodeLensProvider implements CodeLensProvider {
 				'type': 'go',
 				'request': 'launch',
 				'mode': 'test',
-				'env': {
-					'GOPATH': process.env['GOPATH']
-				}
+				'env': getTestEnvVars()
 			};
 
 	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -17,12 +17,12 @@ export class GoRunTestCodeLensProvider implements CodeLensProvider {
 				'name': 'Launch',
 				'type': 'go',
 				'request': 'launch',
-				'mode': 'test',
-				'env': getTestEnvVars()
+				'mode': 'test'
 			};
 
 	public provideCodeLenses(document: TextDocument, token: CancellationToken): CodeLens[] | Thenable<CodeLens[]> {
-		let codeLensConfig = vscode.workspace.getConfiguration('go').get('enableCodeLens');
+		let config = vscode.workspace.getConfiguration('go');
+		let codeLensConfig = config.get('enableCodeLens');
 		let codelensEnabled = codeLensConfig ? codeLensConfig['runtest'] : false;
 		if (!codelensEnabled || !document.fileName.endsWith('_test.go')) {
 			return;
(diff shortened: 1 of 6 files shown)


goenning:
btw, I guess build didn't failed due to this PR changes.

src/goRunTestCodelens.ts
```diff
@@ -69,7 +68,12 @@ export class GoRunTestCodeLensProvider implements CodeLensProvider {
 					arguments: [ { functionName: func.name} ]
 				};
 
-				let config = Object.assign({}, this.debugConfig, { args: ['-test.run', func.name], program: path.dirname(document.fileName) });
+				const args = ['-test.run', func.name];
+				const program = path.dirname(document.fileName);
+				const env = vsConfig['testEnvVars'] || {};
```
ramya-rao-a: Why not use the `getTestEnvVars()` function from `util.ts` here as well?

src/goRunTestCodelens.ts
```diff
@@ -9,29 +9,28 @@ import vscode = require('vscode');
...
 import { GoDocumentSymbolProvider } from './goOutline';
 
 export class GoRunTestCodeLensProvider implements CodeLensProvider {
 	private readonly debugConfig: any = {
 				'name': 'Launch',
 				'type': 'go',
 				'request': 'launch',
-				'mode': 'test',
-				'env': {
-					'GOPATH': process.env['GOPATH']
```
ramya-rao-a: am guessing you are removing this because you expect this to be passed in the testEnvFile?

For someone who has set the GOPATH in the settings, but doesn't use the `testEnvVars` or the new `testEnvFile` setting, the debug test codelens will not work if this is removed.

src/util.ts
```diff
@@ -296,6 +297,21 @@ export function getToolsEnvVars(): any {
...
 
+export function getTestEnvVars(): any {
+	const config = vscode.workspace.getConfiguration('go');
+	const toolsEnv = getToolsEnvVars();
+	const testEnv = config['testEnvVars'] || {};
+
+	let fileEnv = {};
+	let testEnvFile = config['testEnvFile'];
+	if (testEnvFile) {
+		testEnvFile = testEnvFile.replace(/\${workspaceRoot}/g, vscode.workspace.rootPath);
```
ramya-rao-a: you can use the `resolvePath` method here.

src/util.ts
```diff
@@ -296,6 +297,21 @@ export function getToolsEnvVars(): any {
 	return Object.assign({}, process.env, toolsEnvVars);
 }
 
+export function getTestEnvVars(): any {
```
ramya-rao-a: `goTest.ts` would be a better place for this function as that file has all things test related.

src/util.ts
```diff
@@ -5,9 +5,10 @@
 
 import vscode = require('vscode');
 import path = require('path');
-import { getGoRuntimePath, getBinPathWithPreferredGopath, resolvePath } from './goPath';
+import { parseEnvFile, getGoRuntimePath, getBinPathWithPreferredGopath, resolvePath, stripBOM } from './goPath';
```
ramya-rao-a: stripBOM not needed here

src/goRunTestCodelens.ts
goenning: I removed because of [this line](https://github.com/goenning/vscode-go/blob/test-env-file/src/debugAdapter/goDebug.ts#L218). GoDebug is always loading everything `process.env`. Am I missing something?

src/goRunTestCodelens.ts
goenning: From the codelens point of view, It's just creating a VSCode debug configuration with `env` and `envFile` that is later processed/merged by GoDelve, like it does with any user-configured `launch.json`.

src/util.ts
goenning: Nice catch, I'll remove it.

src/util.ts
goenning: I'll move it :)

src/util.ts
goenning: Nice, didn't see this function. Will change it.

src/goRunTestCodelens.ts
ramya-rao-a: Good point

PR #997, opened 2017-05-14
Support configuring gocode.

mprobst:
- support configuring the binary location for gocode
- support passing additional default flags to gocode

This can be used to for example connect to a remote gocode server, use
a different socket type, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,6 +500,19 @@
           "default": {},
           "description": "Environment variables that will passed to the processes that run the Go tools (e.g. CGO_CFLAGS)"
         },
+        "go.gocodeTool": {
+          "type": "string",
+          "default": null,
+          "description": "The gocode tool to run, e.g. '/usr/bin/mygocode'. Defaults to look up gocode in GOPATH."
+        },
+        "go.gocodeFlags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "description": "Flags to pass to gocode (e.g. ['-address', '-sock'])"
+        },
         "go.gocodeAutoBuild": {
           "type": "boolean",
           "default": true,
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -11,6 +11,7 @@ import { dirname, basename } from 'path';
 import { getBinPath, parameters, parseFilePrelude, isPositionInString, goKeywords, getToolsEnvVars } from './util';
 import { promptForMissingTool } from './goInstallTools';
 import { listPackages, getTextEditForAddImport } from './goImport';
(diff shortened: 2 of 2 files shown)


mprobst:
Friendly ping @ramya-rao-a.

ramya-rao-a:
Hey @mprobst, have been buy this month, so didn't get much time to go over any of the PRs, but will get to them this week.

Thanks for your patience :)

mprobst:
No worries, just wanted to make sure this didn't drown in the noise or wait
for something from my side :-)

ramya-rao-a:
@mprobst 

I am a little out of my depth here, so if you don't mind and answer a few queries, that would be great.

> support configuring the binary location for gocode

What is the scenario for this? For all the tools we currently use in the extension we first try to find it in `GOPATH/bin`, then `PATH` and then `GOROOT/bin`. It doesn't feel right that `gocode` gets to have a setting to specify a location when the rest of the tools can't. 

> support passing additional default flags to gocode

There are flags other than `addr` and `sock` which when passed might result in `gocode` behaving unexpectedly. Like `-s`, `-debug`, `profile` etc. Is there any other scenario other than wanting to connect to a remote gocode server where you see the use of configuring the gocode flags?

If not, then I'd prefer a setting like `go.gocodeServerAddress` which when provided you use it for `addr` and pass `tcp` for `sock`

mprobst:
My use case is that I have a `gocode` binary that behaves differently from the one in my `$GOPATH` in use for some projects. My `gocode` calls through an ssh tunnel to a remote machine. It's maybe similar to the situation with `goimports` vs `gofmt` vs `goreturns`.

I don't actually need the flags right now, I added that for symmetry with `go.formatFlags`, and it seems useful to configure the external tool. I agree it has some potential sharp corners - but so will configuring the server address (in case the address is wrong). I'd expect a user who configures that to be able to debug it when they misconfigured their options.

ramya-rao-a:
For your case, you can make use of the the `go.toolsGopath` setting in the Workspace settings of the few projects where you want to use a different gocode.

- Move your special `gocode` binary inside a `bin` folder. Say the final location is `/Users/me/mySpecialGoCode/bin`
- In the Workspace settings for the projects where you want to use the special `gocode`, set `go.toolsGopath` to `/Users/me/mySpecialGoCode`. Note: Ensure you do this in the Workspace settings and not the User settings. The settings from the latter will get applied to all projects.

This way the extension will first look for the Go tools in the location provided in `go.toolsGopath`, then fallback to the actual GOPATH, then PATH and so on.

Does that help?

mprobst:
It does, makes sense.

PR #1006, opened 2017-05-22
Add testEnvVars from workspace config to debug environment

Emreu:
Command **"debug test"** from codelense did not populate env vars with those from workspace config. Usually this command runs in same environment as plain tests, so it is natural to expect **testEnvVars** from workspace config being used.
This pr fix behavior to expected.


ramya-rao-a:
Thanks for the fix @Emreu !

Issue #1015, opened 2017-05-31
vscode triggeres too many go build/lint/... tasks


grobie:
First of all, thanks for all your work on vscode and the go plugin, the integration is awesome.

My issue is that vscode can easily render my system unusable for minutes by triggering many parallel tasks for the same actions:

* Using gorename and changing many files at once, then selecting `Save All` seems to start multiple build / compile jobs, supposedly one for each changed file? I had to restart my laptop as no user input was possible anymore.
* Using go-metalinter on save seems to run the linters across the whole project even if only a single file has been changed. Multiple files saves in short succession will consume all CPU for minutes on a bigger project (kubernetes, prometheus).

I disabled metalinter already for that reason, here is my user config:

```json
// Place your settings in this file to overwrite the default settings
{
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
    "go.gopath": "/home/grobie/code/go",
    "go.formatTool": "goreturns",
    // "go.lintTool": "gometalinter",
    // "go.lintFlags": ["--disable-all", "--deadline=1m", "--enable=gofmt", "--enable=golint", "--enable=vet", "--enable=staticcheck", "--enable=unconvert", "--enable=unused"],
    "window.zoomLevel": 0,
    // "editor.wordWrap": false,
    "workbench.statusBar.visible": true,
    "workbench.activityBar.visible": true,
    "workbench.iconTheme": "vs-minimal",
    "workbench.welcome.enabled": false,
    "workbench.colorTheme": "Visual Studio Dark"
}
```

Here is an example of all the builds started after a gorename and save all in 20+ files
![screenshot from 2017-05-31 10-29-33](https://cloud.githubusercontent.com/assets/3432/26623159/2035633a-45ed-11e7-9872-5bc6ecef5984.png)

Is there anything I can do to limit the lintTool to the currently changed file (package)? Are there any other options I missed to improve the behavior?

ramya-rao-a:
There is nothing I know of that you (as a user) can do at the moment to reduce the triggering of build/lint/vet on multiple files when Save All is used.

We should see if there is a way to identify a Save All event vs Save event in which case the extension can be smarter and trigger build/lint/vet on a package only once even if multiple files in the package was saved.

grobie:
[...]

PR #1023, opened 2017-06-07
Run on-save tools for entire workspace

vapourismo:
Changes the default behavior from running lint/vet/build for the package of the current active file to running it for the entire workspace.

Potential downside is not having a flag which allows the behavior to be changed between `current package` and `workspace` (as with #830). Something for the future.

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -106,43 +106,6 @@ function runTool(args: string[], cwd: string, severity: string, useStdErr: boole
 	});
 }
 
-/**
- * Check if the given path is a directory.
- * @param filePath Path to the potential directory
- */
-function isDirectory(filePath: string): boolean {
-	try {
-		return fs.statSync(filePath).isDirectory();
-	} catch (e) {
-		return false;
-	}
-}
-
-/**
- * List directories (including the given dirPath) which contain files with the '_test.go' suffix.
- * This ignores hidden directories and their sub-directories.
- * @param dirPath Directory in which to start searching
- */
-function findTestDirectories(dirPath: string): string[] {
-	let results = [];
-	let files = fs.readdirSync(dirPath);
-
-	// Contains test files?
-	if (files.some(name => name.endsWith('_test.go'))) {
-		results.push(dirPath);
-	}
-
-	let subDirs = files.filter(name => {
-		return isDirectory(path.join(dirPath, name)) && !name.startsWith('.');
-	});
-
-	subDirs.forEach(name => {
-		results = results.concat(findTestDirectories(path.join(dirPath, name)));
-	});
-
-	return results;
-}
-
 export function check(filename: string, goConfig: vscode.WorkspaceConfiguration): Promise<ICheckResult[]> {
 	outputChannel.clear();
 	let runningToolsPromises = [];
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
We do need a flag to control current package vs whole workspace.

In #830 we talked about a setting `go.multiPackage`, but on thinking more on this, something like the below would be better instead of adding another setting.

```
 "go.buildOnSave": {
          "type":"string",
          "enum": [            
            "currentPackage",
            "currentWorkspace"
          ],
          "default": "currentPackage",
          "description": "Run 'go build'/'go test -c' on save on either the current package or the current workspace."
        }
```

With this, people who have set `go.buildOnSave` to `true` or `false` beforehand, will get a linting error

image

In the code, we can continue to honor either of `true`, `false` as well as the new values so that we are backward compatible.

vapourismo:
Which value should you set, if you want it to be disabled? `null` or `"nothing"`?

ramya-rao-a:
How about `off` ?

vapourismo:
Okay, I added these options as `workspace`, `package` and `off`.

There are those two failed tests in the CI which only occur for macOS. Seems to be something with `gometalinter`. They seem odd to me, since I didn't modify how the linter is invoked.

vapourismo:
Nevermind, it is just the `tip` line that is failing. And it seems unrelated to this PR.

src/goCheck.ts
```diff
@@ -142,20 +142,41 @@ export function check(filename: string, goConfig: vscode.WorkspaceConfiguration)
...
+			if (!buildWorkspace) {
+				args.push('-o', tmppath);
+			}
+		}
+
+		args.push('-tags', buildTags, ...buildFlags);
+
+		if (buildWorkspace) {
+			buildWorkDir = vscode.workspace.rootPath;
+			args.push('./...');
```
ramya-rao-a: When the current file that is open is a test file, this will end up running `go test -i ./...` which will run the tests in the workspace instead of trying to compile them

For compiling test files, we need to use `go test -c` and `-c` cannot be used on multi-package.

@rakyll @campoy @mattetti Do you know of a way to compile test files under a directory?

vapourismo:
How about this? Original behavior with `package` should be the same, I think.

src/goCheck.ts
```diff
@@ -106,6 +106,43 @@ function runTool(args: string[], cwd: string, severity: string, useStdErr: boole
...
+		return false;
+	}
+}
+
+/**
+ * List directories (including the given dirPath) which contain files with the '_test.go' suffix.
+ * This ignores hidden directories and their sub-directories.
+ * @param dirPath Directory in which to start searching
+ */
+function findTestDirectories(dirPath: string): string[] {
```
ramya-rao-a: I appreciate the work here. Consider this
- User with a very large workspace with deeply nested directories
- User with auto-save on

The directories check here is synchronous.
We later wait for all promises (build, lint, vet) to finish before showing diagnostics to the user. 

So you can imagine for each save, this might delay results more than needed.
Note: We have #1015 tracking to find a way to optimize this.

Also, I'd much rather prefer a better way to compile test files from Go itself.

Until then, we can skip trying to compile the test files in the whole workspace.

vapourismo: I've had the exact same thought, but I've come to the conclusion that it doesn't matter. If you set it to build the entire workspace, you should expect that to happen. In case somebody uses a big workspace, they can set it to build only the current package in the workspace-local settings.

Additionally, even though the directory traversal uses synchronous functions, it is way faster than the lint/vet/etc jobs. It appears that the extension accumulates the results of all "check"-Promises anyway, so delaying some for a couple milliseconds doesn't really hurt.

I understand and agree that this is not a particularly smart solution, but it should work for most use cases.

> Also, I'd much rather prefer a better way to compile test files from Go itself.

Not sure, if that is going to happen. The only smart alternative is to run `go test -run ! ./...`, which will check the test files, but won't run any  since they don't match the pattern. Could you work with that?

ramya-rao-a: > go test -run ! ./...

Lets give this a try. I was initially not in favor of this as the output is different (lots of "no test files" followed by a "build failed" on the pkg which has error) and we will have to write a different regex to do the problem matching. 

But taking a closer look, there is also a line in there which matches our expectation and our current regex should work.

screen shot 2017-06-08 at 12 12 22 pm

vapourismo:
`go test -run ! ./..` works, sort of. It also check the packages, so there are duplicates from `go build`. 

Interestingly, it also checks packages that do not contain tests. So theoretically `go build` and `go test` can be merged, in the check function.

vapourismo:
Latest commit prevents duplicates.

PR #1032, opened 2017-06-14
Show full package import path for tools install result.

### LINKED ISSUE #1029: Installing go tools must print the import path
```
Installing 1 tool
  impl

Installing impl SUCCEEDED

All tools successfully installed. You're ready to Go :).
```

This doesn't help me. I maintain a separate `install_gotools.sh` script that I run on all my workstations to install go development tools (incl. the ones vscode-go uses). I think this operation should print the full package import path while it's doing `go get`, not just the package name.

bidrohi:
fixes #1029


ramya-rao-a:
Thanks @bidrohi

PR #1034, opened 2017-06-15
Add support for megacheck as the lintTool

fortytw2:
This PR adds support for @dominikh's all-in-one linter tool. Generally it's much quicker than `gometalinter` and uses fewer resources, as it shares work between all checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ If you want to run only specific linters (some linters are slow), you can modify
   "go.lintFlags": ["--disable-all", "--enable=errcheck"],
 ```
 
+Alternatively, you can use [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck) which 
+may have significantly better performance than `go-tools`, while only supporting a subset of the tools.
+
 Finally, the result of those linters will show right in the code (locations with suggestions will be underlined),
 as well as in the output pane.
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ If you want to run only specific linters (some linters are slow), you can modify
 ```
 
 Alternatively, you can use [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck) which 
-may have significantly better performance than `go-tools`, while only supporting a subset of the tools.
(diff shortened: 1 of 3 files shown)


fortytw2:
👋 anyone get a chance to look at this yet?

README.md
```diff
@@ -80,6 +80,9 @@ If you want to run only specific linters (some linters are slow), you can modify
   "go.lintFlags": ["--disable-all", "--enable=errcheck"],
 ```
 
+Alternatively, you can use [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck) which 
+may have significantly better performance than `go-tools`, while only supporting a subset of the tools.
```
ramya-rao-a: what `go-tools` are you referring to here?

package.json
```diff
@@ -391,7 +391,8 @@
           "description": "Specifies Lint tool name.",
           "enum": [
             "golint",
-            "gometalinter"
+            "gometalinter",
+            "megacheck"
```
ramya-rao-a: I tried testing this by declaring but not using a variable, expecting the linting error "declared but not used".

golint and gometalinter send such warnings via stdout which is what the extension uses.

but megacheck seems to be sending the warnings via stderr. which is why they don't show up in the Problems view.

Can you confirm?

README.md
fortytw2: I definitely meant `gometalinter`, whoops

package.json
fortytw2: Hmm, I definitely tested that this was working before sending the PR in, seemed ok for me. Have a new computer setup since then though, so I'll get my env set back up and verify (hopefully later this evening EST).

fortytw2: image

Finally got around to testing, (hello from Gophercon!), everything looks like it works ok for me without changes.

ramya-rao-a: I still don't see it...

lint

ramya-rao-a: Ah! I was looking for a linting warning that staticcheck, gosimple and unused do not give.

The below works

screen shot 2017-07-16 at 2 03 02 pm

fortytw2: Awesome! 😄

PR #1059, opened 2017-06-29
code coverage: Added a gutter decorator and bugfix

alexandrevez:
- Added a gutter decorator for code coverage
- Added the configuration option to choose between "gutter" and "highlight"
- Fixed a bug where coverage files are not processed because `go-code-cover` file starts each line with an underscore


alexandrevez:
Can't reproduce the build failure. Let me know if you need something

package.json
```diff
@@ -494,6 +495,15 @@
           "default": "showBothCoveredAndUncoveredCode",
           "description": "Use these options to control whether only covered or only uncovered code or both should be highlighted after running test coverage"
         },
+        "go.coverageDecorator": {
+          "type": "string",
+          "enum": [
+            "highlight",
+            "gutter"
+          ],
+          "default": "hightligh",
```
ramya-rao-a: missing t?

src/goGutter.ts
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+import vscode = require('vscode');
+
+export let coveredGutter;
+export let uncoveredGutter;
+
+export function initGutterDecorators(ctx: vscode.ExtensionContext) {
```
ramya-rao-a: any reason to have this in a separate function and call from `goMain.ts` instead of doing the same as covered/uncovered Highlight?

src/goCover.ts
```diff
@@ -140,6 +148,13 @@ export function getCoverage(coverProfilePath: string, showErrOutput: boolean = f
 				let fileRange = data.match(/([^:]+)\:([\d]+)\.([\d]+)\,([\d]+)\.([\d]+)\s([\d]+)\s([\d]+)/);
 				if (!fileRange) return;
 
+				// I really don't know how to fix this elsewhere. I get an underscores in front o thef path (on macOS)
+				//  in go-code-cover file. This fixes it.
+				// _/Users/me/.../file.go => /Users/me/.../file.go
+				if (fileRange[1].charAt(0) === '_') {
```
ramya-rao-a: I don't have this problem in my Mac.

I did see `_` in test output once long ago when I had opened the GOPATH in VS Code instead of just my Go project. Is that the same case with you?

alexandrevez: Not quite the same, but similar. I am using gb to manage my build : https://getgb.io/ and I open the whole gb project.

I think this is not the best way to fix the problem, but understanding the source of the underscores would help. Do you know where it is coming from (and why)?

src/goGutter.ts
alexandrevez: Fixed in next commit. You're right, it is cleaner to init the gutter in goCover.ts

The thing about calling from goMain.ts is that  I need an instance of `vscode.ExtensionContext` in order to access the absolute path of the images folder (gutter colors are actually SVGs rendered in the gutter) :

`gutterIconPath: ctx.asAbsolutePath('images/gutter-green.svg')`
or
`gutterIconPath: ctx.asAbsolutePath('images/gutter-red.svg')`

I did not find another way to access the `vscode.ExtensionContext` from goCover.ts (or any other file other than goMain.ts)

src/goCover.ts
ramya-rao-a: You might not have GOPATH set. Can you run the command `Go: Current GOPATH` and see what is the result?

See https://stackoverflow.com/questions/31664129/executing-go-tests-underscore-in-place-of-correct-path

ramya-rao-a:
@alexandrevez You might not have GOPATH set. Can you run the command `Go: Current GOPATH` and see what is the result?

See https://stackoverflow.com/questions/31664129/executing-go-tests-underscore-in-place-of-correct-path

ramya-rao-a:
Also there is a linting issue that is breaking the build

ramya-rao-a:
Thanks @alexandrevez

alexandrevez:
NP @ramya-rao-a ; thanks to you!

Have a nice week.

ramya-rao-a:
You too and Happy Coding!

ramya-rao-a:
Hey @alexandrevez 

What's your twitter handle? I want to thank you in https://twitter.com/mbanzon/status/891046317111345153

alexandrevez:
Hey,

My handle is @alexandrevez

https://twitter.com/alexandrevez

Thanks!

-------
Alexandre Vézina

ramya-rao-a:
People are loving this feature, thanks to you!

Issue #1037, opened 2017-06-16
formatOnSave re-saves several times


ghaabor:
So I've started using the Go extension yesterday and noticed a weird issue. When formatOnSave is on, I push Ctrl-S and the file is saved several times and the undo stack gets crowded with characters getting selected and deselected. It always selects the same characters that were before the cursor when I pressed Ctrl-S. I tried it with all three formatters and they all do the same. 

I think there could be some bigger issue behind this because turning off formatOnSave stops this behavior but the characters in the file still get selected but only once (same here, the character before the cursor).

I use VSCode with Ruby as well and I haven't noticed anything similar. I use the latest Code (1.13.1 I think) and the latest extension. I'm on Ubuntu 17.04.

ramya-rao-a:
That sounds like some other extension is interfering with the save action. We have seen this before.

What extensions do you have? Can you try disabling all extensions, then enable just the Go extension and see if this still persists?

![image](https://user-images.githubusercontent.com/16890566/27299306-c41d2fee-54e0-11e7-8421-01ffba13dbf4.png)

wader:
Hello, i think i have the same issue.
macOS 10.12.5 (16F73)
Code 1.13.1 (1.13.1)
vscode-go 0.6.62

Get same issue (two file writes) when all extensions disabled except vscode-go. But I noticed that if i disable formatOnSave and just use "Format document" i still get one file write, could that be a lead or is that expected behaviour?

wader:
Oh just noticed https://github.com/Microsoft/vscode-go/blob/bca4dd5f31f32ac49da79580c07b4000f06287a3/src/goFormat.ts#L69 is the save needed because the formatting tools can only read from file?

ramya-rao-a:
When you say 2 file writes do you mean the file state indicator next to file name goes from dirty - save - dirty - save?

The line you are pointing at would only cause dirty - save - save

wader:
Ok the go formatting tools support reading from stdin but vscode-go is currently not using it.

So could the issue be that when `onDidSaveTextDocument` is fired the file has already been saved and then the formatting tools are executed and then the document is saved again?

wader:
You mean the dot left of the filename? when i make a change the dot appears. Then when i save it flickers fast and disappears 

Im using https://github.com/cortesi/modd to trigger builds and if i run it in debug mode (`--debug`) i see two writes.
```
18:42:35: notify.Write: "/path/to/main.go"
18:42:35: Delta:
[...]

Issue #630, opened 2016-11-19
goimports fails to consider global variables in package


briantkennedy:
I just updated to 0.6.49 and format on save is now broken for goimports.  After looking at recent changes, I believe that https://github.com/Microsoft/vscode-go/commit/b143a0d5cde4c6579a23d12f0c277370084b0975 triggered the breakage.

The difference is that the text for the file is passed via stdin instead of providing the on-disk filename.  The consequence of passing from stdin is that the goimports process is not able to determine what global variables exist within the package.  For example, if the global variable "log" is declared foo.go, while saving bar.go which calls "log.Info" or some other function which the "log" package declares, the command will operate on local data and assume you want to import the "log" package rather than using the package global variable which results in the tool incorrectly adding the "log" import.

The fix for this is to do one of the following:
1. Continue to pass the file contents via stdin and invoke as "goimports -d -srcdir [ filename ]"  (see goimports help for the -srcdir flag semantics when a filename is specified)
2. Do not pass the file contents via stdin and invoke "goimports -d [ filename ]"

Unfortunately, there doesn't appear to be a way to configure the current filename as a parameter in "go.FormatFlags" otherwise that would be an easy workaround.

ramya-rao-a:
The reason why I changed from passing the filename to passing the filecontents via stdin was that previously formatting an unsaved file, forcefully saved the file.

#613 is an issue that `goimports` didnt find the imports for vendor packages and so removed them.
The root cause was the change mentioned above.

I put a fix in where I pass the `cwd` to the process running the formatting tool, so that it could find the vendor packages.

I'll see if that fixes your case. If not will look into your suggestions.

Regardless, I'll push an update on Monday that will contain a fix to your issue

ramya-rao-a:
@briantkennedy I am able to repro your case.

Here is the funny thing.

I reverted the change I made regarding passing file contents via stdin instead of providing the filename to the formatting tool.

The revert fixes your issue but only when you use `goreturns`. Using `goimports` still has issues.

Can you confirm this by running both `goreturns` and `goimports` in the command line for your file?
I tried it and `goreturns` does not add the import for `log` and `goimports` does add the import.

ramya-rao-a:
I have reverted the said change, and the latest update (0.6.50) has the fix.

ramya-rao-a:
@briantkennedy I am working on bringing back the "pass file contents via stdin to formatter" feature.
[...]

Issue #613, opened 2016-11-11
Format removes imports of vendored packages in use


sindrepm:
I just upgraded to the latest version and discovered a bug. When I have imports like the following:

```go
import (
   "github.com/xanzy/go-gitlab"
   "github.com/emicklei/go-restful"
)
```

Whenever I save the file, those import lines are removed even if they are referenced in the code. This has worked fine up until the last upgrade so not sure what has happened. From what I have found, it only applies to packages which contains a hyphen as shown above (e.g. "go-gitlab" and "go-restful").

I have been able to reference these packages just by the "gitlab" or "restful" names before, but now it seems that the goimports tool doesn't like that? If I alias the imports with "gitlab" or "restful" it works and the lines are not removed.

It is quite an annoying bug so hopefully this can be fixed.

AlekSi:
> it only applies to packages which contains a hyphen as shown above

For me it's `gopkg.in/Shopify/sarama.v1`. I think this is due to package name is different from import path's last element.

It's really annoying.

AlekSi:
It works with `"go.formatTool": "gofmt"`. So my guess it is something between `goreturns`, `GOPATH` and `vendor/`.

ramya-rao-a:
Can you run `goreturns` and/or `goimports` from the command line independent of VS Code and see if you still have the issue? This is to determine if the problem is with the formatting tool or with how VS Code uses the formatting too.

sindrepm:
I have tried running `goreturns`, `goimports` and `gofmt` directly on the .go file, and it works fine. The import lines are not removed. However, saving from the editor it immediately removes those imports.

ramya-rao-a:
Can you share your settings?

ramya-rao-a:
I tried the below code and I get build errors(`syntax error: unexpected -, expecting expression`), but no imports get removed on save
```
package main

import (
	"fmt"
	"github.com/emicklei/go-restful"
)

func main() {

	fmt.Println("Hello")
	go-restful.BodyParameter("hello", "bye")
}
```

AlekSi:
My settings don't have anything Go-related except `"go.gopath"` with a single path.

Most, if not all, tools for Go (including `gofmt`, `goimports` and `goreturns`) work by parsing code into AST, modifying AST and printing it back. If Go code has syntax errors, it can't be parsed, and the tool just exits with an error message. That's why import is not removed by tool – tool did nothing.

[...]

PR #1069, opened 2017-07-05
Format via stdin instead of document file path

wader:
Fixes double save issues as now we don't have to save
the document to do formatting.

Also fixes save on "Format Document" (now no save is done).

Related to #1037

### DIFF
--- a/src/goFormat.ts
+++ b/src/goFormat.ts
@@ -16,6 +16,7 @@ export class Formatter {
 	public formatDocument(document: vscode.TextDocument): Thenable<vscode.TextEdit[]> {
 		return new Promise((resolve, reject) => {
 			let filename = document.fileName;
+			let documentText = document.getText();
 			let formatTool = vscode.workspace.getConfiguration('go')['formatTool'] || 'goreturns';
 			let formatCommandBinPath = getBinPath(formatTool);
 			let formatFlags = vscode.workspace.getConfiguration('go')['formatFlags'] || [];
@@ -29,7 +30,7 @@ export class Formatter {
 			}
 			let t0 = Date.now();
 			let env = getToolsEnvVars();
-			cp.execFile(formatCommandBinPath, [...formatFlags, filename], {env}, (err, stdout, stderr) => {
+			let process = cp.execFile(formatCommandBinPath, [...formatFlags], {env}, (err, stdout, stderr) => {
 				try {
 					if (err && (<any>err).code === 'ENOENT') {
 						promptForMissingTool(formatTool);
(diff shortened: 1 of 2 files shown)


wader:
Seems to work fine in my limited tests. Feedback very welcome!

One possible issue with this change is that `onWillSaveTextDocument` is now used which seems to have a time limit on how long the `waitUntil` will wait for the edit promise.

> An event that is emitted when a text document will be saved to disk.
>
> Note 1: Subscribers can delay saving by registering asynchronous work. For the sake of data integrity the editor might save without firing this event. For instance when shutting down with dirty files.
>
> Note 2: Subscribers are called sequentially and they can delay saving by registering asynchronous work. Protection against misbehaving listeners is implemented as such:
>* there is an overall time budget that all listeners share and if that is exhausted no further listener is called
>* listeners that take a long time or produce errors frequently will not be called anymore
>
> The current thresholds are 1.5 seconds as overall time budget and a listener can misbehave 3 times before being ignored.

wader:
I must say that i'm very impressed by the the visual code extension development experience and vscode-go code quality, was very easy to get going!

wader:
I noticed one thing while reading the code. In `isDiffToolAvailable` it checks if the `diff` binary is in `PATH` but while reading the jsdiff code can't see that it is actually used. Maybe i'm missing something?

wader:
Maybe with the change `{cwd: vscode.workspace.rootPath}` is needed?

ramya-rao-a:
@wader 

Thanks for the PR. Am back from the vacation and had some time to dig around my previous work with both using `onWillSaveTextDocument` and `stdin`

On the usage of onWillSaveTextDocument:

Telemetry shows that about 1/5th of the Go extension users have had formatting times > 1.5 secs.
If we make this switch, then formatting would stop working for such users. This was the reason that I hadn't made the change to use onWillSaveTextDocument myself when this API was released.

Either, the users have to live with that or we need to be able to provide a fallback

I'll look into what options we have.

On the usage of stdin:

The last time I did this, there were 2 bugs that surfaced due to which I had to revert the change
- #613 Format removes imports of vendored packages in use
- #630 goreturns fails to consider global variables in package

I believe these will occur again with your changes as well?

ramya-rao-a:
> I noticed one thing while reading the code. In isDiffToolAvailable it checks if the diff binary is in PATH but while reading the jsdiff code can't see that it is actually used.

We use `isDiffToolAvailable` to decide whether or not to send the `-d` flag to the formatting tool.

ramya-rao-a:
Also once you have the `stdin` approach working, we don't really need to use the `onWillSaveTextDocument` directly. We can deprecate `go.formatOnSave` and rely on `editor.formatOnSave` (which internally uses the `onWillSaveTextDocument` event) instead.

wader:
@ramya-rao-a Thanks for the feedback

About diff i just found that gofmt actually uses external diff https://github.com/golang/go/blob/master/src/cmd/gofmt/gofmt.go#L257 i just assumed the diff was done in go somehow 😄 

Hopefully I will have some time to look into this again this week. Yeah i also suspect that my change will reintroduce those issues again, will try to to see if it can be fixed.

ramya-rao-a:
Closing this in favor of https://github.com/Microsoft/vscode-go/commit/e95eed681077442a42979f91402d642118798daf

Issue #1060, opened 2017-06-29
Build Workspace except vendor folder


RoarkeRandall:
Hello, currently the setting `"go.buildOnSave"` only allows the 3 options: `workspace`, `package`, and `off`. If you are using glide or the vendoring directory & you use the `workspace` option, then it will build the entire vendoring folder. The directories to build are not a flag so I'm not sure how to handle this scenario. 

For glide you would run `go build $(glide novendor)`, is there a way to replicate this behavior on save?

ramya-rao-a:
Fixed with #1074

ramya-rao-a:
The fix is out in the latest update (0.6.63) to the Go extension

Issue #846, opened 2017-03-07
vscode-go doesn't work well with symlinked directroy


bruceauyeung:
reproduce steps:
1. `mkdir $HOME/heapster-build/src/k8s.io/ -p`
2. `mkdir /tmp/heapster/ -p && cd /tmp/heapster/ && git clone https://github.com/kubernetes/heapster.git`
3. `cd $HOME/heapster-build/src/k8s.io/ && ln -s mkdir /tmp/heapster/`
4. `export GOPATH=$HOME/heapster-build && code $HOME/heapster-build`
5. open `$HOME/heapster-build/src/k8s.io/heapster/metrics/heapster.go` and vscode-go will complain packages such as `k8s.io/apimachinery/pkg/fields` can not be found, but actually they are in the vendor directory.
6. `cd $HOME/heapster-build/src/k8s.io/heapster/metrics` , and then run `go build heapster.go auth.go handlers.go` , this can build successfully.

if i copy `/tmp/heapster/ to `$HOME/heapster-build/src/k8s.io` instead of  making  a symlink , vscode-go works well

vscode:1.10.1
vscode-go:0.6.55
os:Ubuntu 16.04.1 LTS

bruceauyeung:
@ramya-rao-a

ramya-rao-a:
@bruceauyeung I went through your steps in my Mac, except I had to do `ln -s /tmp/heapster/ .` instead of `ln -s mkdir /tmp/heapster`

I can see the errors like you mentioned.

But I also see the same errors when I try `go build` as per your step in the terminal outside of VS Code. They do not build successfully.

Can anybody else repro this? i.e it builds successfully in the terminal?

ramya-rao-a:
I can repro! (finally)

Had to tweak some of your repro steps though.

- Don't create the `heapster` in `/tmp`, because `git clone` will take care of that
- when creating the symlink, `ln -s /tmp/heapster/ .` instead of `ln -s mkdir /tmp/heapster/`

I can see the errors in VS Code, but building in the terminal is successful.

Looks like the process that is spawned to run `go build` by the extension is not aware of the vendor folder, not sure why.

Anybody having any ideas is appreciated.

ramya-rao-a:
Can you try `go list all | grep "glog"` ? Or grep for any of the vendor packages. It doesn't show up for me.

ramya-rao-a:
Turns out command line and newly spawned process resolve `.` differently in case of symlinks.
[...]

PR #1074, opened 2017-07-10
Refactor build workspace on save feature

ramya-rao-a:
The PR refactors the work done around build on save feature in #1023 to fix the below issues

- The issue found in #846 is back with the work done in #1023 
- The `-i` flag which results in faster builds doesn't get used when building workspace
- The vendor packages get built as well when building workspace See #1060


ramya-rao-a:
@vapourismo Since you worked on #1023, can you take a look at this PR? You might find something I missed.

vapourismo:
I can't reliably review it right now, sorry.

PR #1102, opened 2017-07-21
Exclude 'vendor' directory when test all workspace

uudashr:
This is to avoid test vendor directory when using vendor tools such as Glide

### DIFF
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -23,7 +23,7 @@ export function goListAll(): Promise<Map<string, string>> {
 		return Promise.resolve(allPkgs);
 	}
 	return new Promise<Map<string, string>>((resolve, reject) => {
-		cp.execFile(goRuntimePath, ['list', '-f', '{{.Name}};{{.ImportPath}}', 'all'], (err, stdout, stderr) => {
+		cp.execFile(goRuntimePath, ['list', '-f', '{{.Name}};{{.ImportPath}}', 'all'], {env: {}}, (err, stdout, stderr) => {
 			if (err) return reject();
 			stdout.split('\n').forEach(pkgDetail => {
 				if (!pkgDetail || !pkgDetail.trim() || pkgDetail.indexOf(';') === -1) return;
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -230,31 +230,37 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 			return Promise.resolve();
 		}
 
-		if (testconfig.functions) {
-			args.push('-run');
-			args.push(util.format('^%s$', testconfig.functions.join('|')));
-		} else if (testconfig.includeSubDirectories) {
-			args.push('./...');
-		}
-
-		outputChannel.appendLine(['Running tool:', goRuntimePath, ...args].join(' '));
-		outputChannel.appendLine('');
-
-		let proc = cp.spawn(goRuntimePath, args, { env: testEnvVars, cwd: testconfig.dir });
-		proc.stdout.on('data', chunk => {
-			let testOutput = expandFilePathInOutput(chunk.toString(), testconfig.dir);
-			outputChannel.append(testOutput);
-
-		});
-		proc.stderr.on('data', chunk => outputChannel.append(chunk.toString()));
-		proc.on('close', code => {
-			if (code) {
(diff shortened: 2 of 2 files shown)


uudashr:
Done cc @ramya-rao-a

ramya-rao-a:
Thanks @uudashr!

I just added a commit to re-use some code that I pulled out from `goCheck.ts` earlier for finding the non vendor packages

uudashr:
On `goPackages.ts`
```typescript
export function getNonVendorPackages(folderPath: string): Promise<string[]> {
	let goRuntimePath = getGoRuntimePath();

	if (!goRuntimePath) {
		vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
		return Promise.resolve(null);
	}
	return new Promise<string[]>((resolve, reject) => {
		const childProcess = cp.spawn(goRuntimePath, ['list', './...'], { cwd: folderPath });
		let pkgs = [];
		childProcess.stdout.on('data', (stdout) => {
			pkgs = pkgs.concat(stdout.toString().split('\n').filter(pkgPath => pkgPath && pkgPath.indexOf('/vendor/') === -1));
		});

		childProcess.on('close', (status) => {
			return resolve(pkgs);
		});
	});
}
```

The `childProcess.stdout.on('data', (stdout) => {})` can be invoked more than once, and the output can be split like
```
github.com/KurioApp/adops/vendor/github.com/mattes/migrate/testing
github.com/KurioApp/adops/vendor/github.com/mattn/go-colorable
github.com/KurioApp/adops/vendor/github.com/mattn/go-isatty
github.com/KurioApp/adops/vendor/github.com/pmezard/go-difflib/difflib
github.com/Kur
---split---
ioApp/adops/vendor/github.com/shopspring/decimal
github.com/KurioApp/adops/vendor/github.com/streadway/amqp
github.com/KurioApp/adops/vendor/github.com/stretchr/objx
github.com/KurioApp/adops/vendor/github.com/stretchr/testify
github.com/KurioApp/adops/vendor/github.com/stretchr/testify/assert
```
So you have to concat all the output, then split by new line character.
@ramya-rao-a do you want to me to make PR for the fix?

And another one, when I run the test (on my local machine and travis) I always experiencing
`
  1) Go Extension Tests Test Completion on unimported packages:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
`
It caused by `goListAll` method on `goPackages.ts`.. it took more than 5000ms to execute
```typescript
// original source
cp.spawn(goRuntimePath, ['list', '-f', '{{.Name}};{{.ImportPath}}', 'all'])
```
But if I added some options, it executed very fast
```typescript
cp.spawn(goRuntimePath, ['list', '-f', '{{.Name}};{{.ImportPath}}', 'all'], {env: {}})
```

ramya-rao-a:
> So you have to concat all the output, then split by new line character.

Nice catch, yes please send a PR

About the tests running faster with the empty env... that's weird.
I have no idea why that would be so :)
You can add that empty env , just add a comment above it as well, so that I don't end up removing it :)

ramya-rao-a:
@uudashr Sending empty env resets the GOPATH if any was set by VS Code. So I have updated the empty env obj to `getToolEnvVars()` which is what we use in other places.

Can you check if your speed issue still persists after my change?

Also @marwan-at-work pointed out that in 1.9 `./...` will no longer match vendor directory.
So we will have to test for version before trying to run `go list` when building/testing workspaces.
No rush.

uudashr:
@ramya-rao-a Timeout
```
1) Go Extension Tests Test Completion on unimported packages:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
You are not experiencing this on your local machine or travis?

Some in the env make the execution slower, not sure what and why.
Quick workaround: we might not need the environment variable. Possible variable to pass in is `GOROOT` or `GOPATH`

ramya-rao-a:
I used to see it before, but not anymore.

I have pulled the promise out of the individual test to the beginning and moved the completion tests to the end. This will not solve the problem of local failures in case of huge GOPATH, but will definitely solve the issue for travis.

I have some plans to re-write the tests and the caching the packages. Will revisit this then.

PR #1103, opened 2017-07-22
Stream 'go list' cmd for large Browse Packages imports

marwan-at-work:
1. In certain cases, `go list all` was throwing "buffer too large" so it needed to be streamed.
2. No need to highlight an import path to get `Browse Packages` to show the files of that import path.

### DIFF
--- a/src/goBrowsePackage.ts
+++ b/src/goBrowsePackage.ts
@@ -16,21 +16,17 @@ export function browsePackages() {
 	if (!goRuntimePath) {
 		return;
 	}
-	let selection = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.selection : undefined;
 	let selectedText = '';
-	if (selection && !selection.isEmpty) {
-		selectedText = vscode.window.activeTextEditor.document.getText(selection).replace(/"/g, "");
-	} else {
-		// if selectedText is not highlighted, then get the whole line the cursor is currently on. 
-		selectedText = vscode.window.activeTextEditor.document.lineAt(selection.active.line).text.trim();
-
-		// if selectedText is a group import and not a single import line.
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
Thanks @marwan-at-work 

I added a commit for the following cases
- When no editor is active, do not try to guess pkg path
- Cases when there is an alias for the import path

marwan-at-work:
@ramya-rao-a yay!

src/goBrowsePackage.ts
```diff
@@ -16,8 +16,18 @@ export function browsePackages() {
 	if (!goRuntimePath) {
 		return;
 	}
-	let selection = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.selection : undefined;
-	let selectedText = (selection && !selection.isEmpty) ? vscode.window.activeTextEditor.document.getText(selection) : '';
+	let selectedText = '';
+	if (vscode.window.activeTextEditor) {
```
marwan-at-work: Maybe more readable if we did the following? 

```
	const { activeTextEditor } = vscode.window;
	if (activeTextEditor) {
		let selection = activeTextEditor.selection;
		if (!selection.isEmpty) {
			// get selected text
			selectedText = activeTextEditor.document.getText(selection);
		} else {
			// if selection is empty, then get the whole line the cursor is currently on.
			selectedText = activeTextEditor.document.lineAt(selection.active.line).text;
		}
		selectedText = getImportPath(selectedText);
	}
```

ramya-rao-a: yes definitely, I'll sneak that in the next time I touch this file

marwan-at-work: igu :) https://github.com/Microsoft/vscode-go/pull/1105

PR #1113, opened 2017-07-25
Add helloweb snippet

thomasdarimont:
`helloweb` generates small web app with an http endpoint
which returns a greeting and the current time.

This is can be used to quickly generate a minimal web app
for demo purposes.

The generated code looks like:
```go
package main

import (
	"fmt"
	"net/http"
	"time"
)

func greet(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "Hello World! %s", time.Now())
}

func main() {
	http.HandleFunc("/", greet)
	http.ListenAndServe(":8080", nil)
}
```


ramya-rao-a:
This snippet is now out in the latest update to the Go extension 0.6.64

PR #1123, opened 2017-07-28
Add check for files outside $GOPATH

### LINKED ISSUE #1122: Coverage highlighting does not work when files are outside of $GOPATH
[applyCoverage()](https://github.com/Microsoft/vscode-go/blob/4f33a078a8fd65dd9be763b213554501972641bb/src/goCover.ts#L117)  does not work with files outside the $GOPATH.

`coverageFiles` paths come from `go test`, which returns non-$GOPATH paths like `_/main.go`

arbourd:
Fixes #1122


ramya-rao-a:
Thanks @arbourd!

Issue #1120, opened 2017-07-28
Testing tool doesn't show log output


mileusna:
I love "run test" option in test files.

However, after VS Code update on 1.14.2 on MacOS executing tests inside Code doesn't show up any log output. So if I have ```log.Println(err)``` it won't be printed, however, ```fmt.Println(err)``` will be printed.

ramya-rao-a:
I am able to see the log.Println output

<img width="591" alt="screen shot 2017-07-28 at 2 31 33 pm" src="https://user-images.githubusercontent.com/16890566/28737012-769bb256-73a1-11e7-83ab-99b7c3db381f.png">

Can you share sample code?

mileusna:
Sure, anything. Same testing works OK in terminal window.

It just started to happen after the VS Code upgrade.
![screen shot 2017-07-29 at 01 16 13](https://user-images.githubusercontent.com/4697935/28739494-a62eede4-73fb-11e7-8631-cf9f5f613a69.png)

ramya-rao-a:
Ah! Looks like the `log` part is going to stderr and the `fmt` one goes to stdout

<img width="616" alt="screen shot 2017-07-28 at 4 34 09 pm" src="https://user-images.githubusercontent.com/16890566/28739766-b2e338a4-73b2-11e7-8640-de522ef83336.png">

And in a recent commit stderr is not shown unless there is a test failure https://github.com/Microsoft/vscode-go/commit/94dadaa42af7e59cee6f97eb660de1a14ad45409

@uudashr Can you take a look at this?

mileusna:
Sorry, but I don't see log.* output, failure or no failure. It has nothing to do with the test's outcome.
![screen shot 2017-07-29 at 05 37 28](https://user-images.githubusercontent.com/4697935/28741828-18ebd756-7420-11e7-8ec5-bd6a8b7e69c9.png)

ramya-rao-a:
@mileusna Its right there.... look at the line which has the file path... In the end it says "Failed, but no Hello"

ramya-rao-a:
I have pushed a fix. But even with the fix the `log` will appear after the rest of the output. This is because like I said earlier, log is going to stderr. And while running the `go test` the extension can only fetch stdout and stderr separately.

uudashr:
@ramya-rao-a your changes on the master?

mileusna:
[...]

PR #1124, opened 2017-07-29
Return the output immediately both for stdout and stderr

uudashr:
This supposed to solve issue #1120


uudashr:
done @ramya-rao-a

PR #1136, opened 2017-08-08
goBrowsePackages: toString bug fix + do not wait for goListAll

marwan-at-work:
This PR fixes two things:

1. An error in `goListAll()` where we call `chunks.toString()`. The method puts `,`'s between every chunk which causes some lines to not be correct imports. I think this is why the auto-import on tab didn't work for all imports (at least for me)

2. Refactored `goBrowsePackages` so that we don't have to wait for `goListAll` to finish. This should speed up browsing packages a lot at initial time, especially for users with slower computers and very large GOPATH tree.

In `goBrowsePackages` we still need a feature that depends on `goListAll()`, which is to show a list of *all* the packages in your GOPATH to select from. So that logic is still there.

### DIFF
--- a/src/goBrowsePackage.ts
+++ b/src/goBrowsePackage.ts
@@ -26,10 +26,63 @@ export function browsePackages() {
 		selectedText = getImportPath(selectedText);
 	}
 
-	if (isGoListComplete()) {
-		return showPackages(selectedText);
+	showPackageFiles(selectedText);
+}
+
+function showPackageFiles(pkg: string)  {
+	const goRuntimePath = getGoRuntimePath();
+	if (!goRuntimePath) {
+		return vscode.window.showErrorMessage('Could not locate Go path. Make sure you have Go installed');
 	}
 
+	cp.execFile(goRuntimePath, ['list', '-f', '{{.Dir}}:{{.GoFiles}}:{{.TestGoFiles}}:{{.XTestGoFiles}}', pkg], (err, stdout, stderr) => {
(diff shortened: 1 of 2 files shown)


src/goBrowsePackage.ts
```diff
@@ -26,64 +26,73 @@ export function browsePackages() {
 		selectedText = getImportPath(selectedText);
 	}
 
-	if (isGoListComplete()) {
-		return showPackages(selectedText);
+	showPackageFiles(selectedText);
+}
+
+function showPackageFiles(pkg: string)  {
```
ramya-rao-a: if `pkg` is empty string which can happen when the command is run without any editor open, then we will be needlessly creating a process to run `go list`

src/goBrowsePackage.ts
```diff
@@ -26,64 +26,73 @@ export function browsePackages() {
...
+			.then(pkgFromDropdown => {
+				if (!pkgFromDropdown) return;
+				showPackageFiles(pkgFromDropdown);
 			});
-		});
 	});
 }
 
+function showTryAgainLater() {
+	// `go list all` has not completed. Wait for a second which is an acceptable duration of delay.
```
ramya-rao-a: if at the end of this second, `go list` completes, we won't be showing anything. Was that intended?

marwan-at-work: This is so that we don't get stuck in an infinite loop. And since we say try after sometime, it shouldn't be a surprise.

PR #1156, opened 2017-08-22
merge test flags when there are extra args.

### LINKED ISSUE #1155: "Test Package" ignores buildFlags and testFlags when coverOnTestPackage is true
title

OneOfOne:
Fixes #1155

### DIFF
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -65,7 +65,7 @@ export function getTestEnvVars(config: vscode.WorkspaceConfiguration): any {
 
 export function getTestFlags(goConfig: vscode.WorkspaceConfiguration, args: any): string[] {
 	let testFlags = goConfig['testFlags'] ? goConfig['testFlags'] : goConfig['buildFlags'];
-	return (args && args.hasOwnProperty('flags') && Array.isArray(args['flags'])) ? args['flags'] : testFlags;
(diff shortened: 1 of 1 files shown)


src/testUtils.ts
```diff
@@ -65,7 +65,7 @@ export function getTestEnvVars(config: vscode.WorkspaceConfiguration): any {
 
 export function getTestFlags(goConfig: vscode.WorkspaceConfiguration, args: any): string[] {
 	let testFlags = goConfig['testFlags'] ? goConfig['testFlags'] : goConfig['buildFlags'];
-	return (args && args.hasOwnProperty('flags') && Array.isArray(args['flags'])) ? args['flags'] : testFlags;
+	return testFlags.concat((args && args.hasOwnProperty('flags') && Array.isArray(args['flags'])) ? args['flags'] : []);
```
ramya-rao-a: There can be duplicates here, so we should check for that instead of just concat

ramya-rao-a: On second thoughts, I want `getTestFlags` to not use build/test flags settings when args are present.

The initial idea was that you can have keybindings to the test commands and can pass the flags you need via args.

If the args get combined with the test/build flags, then there would be no way to override them

I fixed #1155 with this instead https://github.com/Microsoft/vscode-go/commit/aca9f3105b1ae0a654041bbee758330498d88fc8

Give it a try and let me know if the tests and coverage run using the build/test flags as expected now

PR #1162, opened 2017-08-24
Fix the non vendor packages output handling

uudashr:
Found bugs... fixed now


uudashr:
cc @ramya-rao-a

ramya-rao-a:
Thanks @uudashr!

PR #1164, opened 2017-08-25
pass <package> to dlv test/debug if applicable

zhouhaibing089:
we should prefer passing package name to `dlv test` or `dlv debug`. this fixes the issue mentioned at https://github.com/Microsoft/vscode-go/issues/846#issuecomment-323589274.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -230,14 +230,19 @@ class Delve {
 
 			let env = Object.assign({}, process.env, fileEnv, launchArgsEnv);
 
+			let dirname = isProgramDirectory ? program : path.dirname(program);
 			if (!fileEnv['GOPATH'] && !launchArgsEnv['GOPATH'] && (mode === 'debug' || mode === 'test')) {
 				// If user hasnt specified GOPATH & file/package to debug is not under env['GOPATH'], then infer it from the file/package path
 				// Not applicable to exec mode in which case `program` need not point to source code under GOPATH
-				let programNotUnderGopath = !env['GOPATH'] || !getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], isProgramDirectory ? program : path.dirname(program));
+				let currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+				let programNotUnderGopath = !env['GOPATH'] || !currentGOWorkspace;
 				if (programNotUnderGopath ) {
-					env['GOPATH'] = getInferredGopath(isProgramDirectory ? program : path.dirname(program)) || env['GOPATH'];
+					env['GOPATH'] = getInferredGopath(dirname) || env['GOPATH'];
 				}
 			}
+
+			let currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+			let importPath = currentGOWorkspace ? dirname.substr(currentGOWorkspace.length + 1) : '.';
 
 			verbose(`Using GOPATH: ${env['GOPATH']}`);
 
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@zhouhaibing089 Your issue in https://github.com/Microsoft/vscode-go/issues/846#issuecomment-323589274 is about running tests where as your PR here is in the debugging area. How will it help in fixing the issue of running tests?

zhouhaibing089:
@ramya-rao-a the change in `src/testUtils.ts` is about the fix on running test.

src/testUtils.ts
```diff
@@ -123,6 +123,11 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 			return Promise.resolve();
 		}
 
+		// append the package name to args if applicable
```
ramya-rao-a: - this might not work in case of multiple GOPATHs, because then `process.env['GOPATH']` can have `:` or `;` separated paths. Follow https://github.com/Microsoft/vscode-go/blob/0.6.63/src/goCheck.ts#L180-L181 instead
- we are moving away from using GOPATH from `process.env`. Instead use `getCurrentGoPath()`
- When the command `Go: Test All Packages In Workspace` is run, this `go test` command will fail with "can't load package.. no buildable Go source files in ..." if the workspace itself is just a folder containing packages and not a package itself

zhouhaibing089: PTAL again, I am not sure how the third item can be solved(and seems it is more than the effort here)

ramya-rao-a: `testconfig.includeSubDirectories` is `true` only for the "test all pks in workspace" scenario.
So you can add a check to ensure `testconfig.includeSubDirectories` is `false` before adding the importPath

src/debugAdapter/goDebug.ts
```diff
@@ -230,15 +230,20 @@ class Delve {
...
+				let currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+				let programNotUnderGopath = !env['GOPATH'] || !currentGOWorkspace;
 				if (programNotUnderGopath ) {
-					env['GOPATH'] = getInferredGopath(isProgramDirectory ? program : path.dirname(program)) || env['GOPATH'];
+					env['GOPATH'] = getInferredGopath(dirname) || env['GOPATH'];
 				}
 			}
 
+			let currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+			let importPath = currentGOWorkspace ? dirname.substr(currentGOWorkspace.length + 1) : '.';
```
ramya-rao-a: shouldnt we do this only if `mode` is `debug` or `test`?
Other values for `mode` are `remote` and `exec`

ramya-rao-a:
@zhouhaibing089 One issue with passing the package to `go test` is that the `log.Println` or `fmt.Println` from the tests are no longer getting piped to stdout

@rakyll @francesc Would you know why `go test` would print out the `log.Println` and `fmt.Println` statements but `go test import-path-to-my-package` doesn't?

zhouhaibing089:
`go test -v import-path-to-my-package` will do.

zhouhaibing089:
If this is a regression issue, I propose that we add a configuration like: `test.verbose`, if it is true, we add the flag `-v`, otherwise suppress the output.

rakyll:
go test and go test <pkg> should be behave similarly, not outputting to stdout if -v is not set. Can you file an [issue](https://golang.org/issues/new) if that's not the case?

ramya-rao-a:
Looks like its a platform issue

In Windows `go test -v import-path-to-my-package` prints the text from `log.Println` statements
But in Mac it doesn't. Both are using Go 1.9

Logged https://github.com/golang/go/issues/22113

ramya-rao-a:
Nope not a platform issue. 
It was all me and my duplicate set up of the same project in the default GOPATH and my other GOPATH.

I was making the changes in one and `go test package-path` was testing the other.

All is well with the world folks, sorry for the churn

PR #1165, opened 2017-08-25
Improvements for "go to implementation"

hummerd:
1. Not fail if guruOutput.to or guruOutput.to_method is undefined
2. If we looked for method implementation, go to method implementations only

### DIFF
--- a/src/goImplementations.ts
+++ b/src/goImplementations.ts
@@ -22,6 +22,7 @@ interface GuruImplementsRef {
 interface GuruImplementsOutput {
 	type: GuruImplementsRef;
 	to: GuruImplementsRef[];
+	to_method: GuruImplementsRef[];
 	from: GuruImplementsRef[];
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Thanks @hummerd!

hummerd:
I am glad I can help!

ramya-rao-a:
@hummerd I'll be releasing an update on the coming Tuesday or Wednesday and your changes will be part of it.

Do you mind helping me out with some testing before that? All you have to do is
- Run `code --install-extension Go-latest.vsix` from the location where the above is installed
- Reload vs code and use it like you would any other day and let me know if there are any issues.

hummerd:
Ok, I've installed Go-latest.vsix. If anything goes wrong I will report about it.

ramya-rao-a:
Thanks!

PR #1176, opened 2017-08-30
Snippets for t.Log() and t.Logf()

kmikiy:
`t.Log(...) and t.Logf(...)` are also commonly used. I propose adding:

```json
{
	"t.Log": {
		"prefix": "tp",
		"body": "t.Log(\"$1\")"
	},
	"t.Logf variable content": {
		"prefix": "tv",
		"body": "t.Logf(\"${1:var}: %#+v\\\\n\", ${1:var})"
	}
}
```

### DIFF
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -100,6 +100,14 @@
 			"prefix": "lv",
 			"body": "log.Printf(\"${1:var}: %#+v\\\\n\", ${1:var})"
 		},
+		"t.Log": {
+			"prefix": "tp",
+			"body": "t.Log(\"$1\")"
+		},
+		"t.Logf variable content": {
+			"prefix": "tv",
+			"body": "t.Logf(\"${1:var}: %#+v\\\\n\", ${1:var})"
+		},
 		"make(...)": {
 			"prefix": "make",
 			"body": "make(${1:type}, ${2:0})"
--- a/snippets/go.json
+++ b/snippets/go.json
@@ -101,11 +101,15 @@
 			"body": "log.Printf(\"${1:var}: %#+v\\\\n\", ${1:var})"
 		},
 		"t.Log": {
-			"prefix": "tp",
+			"prefix": "tl",
 			"body": "t.Log(\"$1\")"
 		},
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Hey @kmikiy 

Sorry it took me so long to get to this PR.

Why not use `tl` and `tlf` as prefixes instead? Arent they more intutive for `tLog` and `tLogf` ?

kmikiy:
I chose `tp` and `tv` initially because it was analogous to the already existing `lp` And `lv`.
```json
...
		"log.Println": {
			"prefix": "lp",
			"body": "log.Println(\"$1\")"
		},
		"log.Printf": {
			"prefix": "lf",
			"body": "log.Printf(\"$1\", ${2:var})"
		},
		"log variable content": {
			"prefix": "lv",
			"body": "log.Printf(\"${1:var}: %#+v\\\\n\", ${1:var})"
		},
...
```

Although, `tf` won’t work because it would conflict with the existing `tf` which is for test functions.

I will gladly change the prefix if you think `tl`and `tlf` is mor fitting.

ramya-rao-a:
`lp` for `log.Println` makes sense its the first letter for the pkg name and the function name
`tp` for `t.Log` wouldn't because there is no `p` in anywhere in the pkg/function name
So its a definite no for `tp`

That leaves `tl` for `t.Log`
And to keep the  prefix for `t.Logf` analogous to `t.Log`, `tlf`is better

kmikiy:
How about
`tl` for `t.log`
`tlf` for `t.logf`
and 
`tlv` for logging variable content with `t.logf`
?

e.g:
```
		"t.Log": {
			"prefix": "tl",
			"body": "t.Log(\"$1\")"
		},
		"t.Logf": {
			"prefix": "tlf",
			"body": "t.Logf(\"$1\", ${2:var})"
		},
		"t.Logf variable content": {
			"prefix": "tlv",
			"body": "t.Logf(\"${1:var}: %#+v\\\\n\", ${1:var})"
		},
```

ramya-rao-a:
sure

Issue #1178, opened 2017-08-30
Debugger fails when stepping into standard library with "Unable to open 'server.go'"


rpfister51:
While debugging the sample application (`github.com/lukehoban/webapp-go/tree/debugging`)  (I’ve attached a partial screenshot.)  The message is “Unable to open ‘server.go’: File not found (c:\Users\rpfister\usr\local\go\src\net\http\server.go).

![unable to find server go](https://user-images.githubusercontent.com/12035948/29892194-b8a98286-8d9b-11e7-9bc5-2df4dbf5dbe8.png)

This seems to be a mishmash of fragments from the local and container paths.

The local file is here: `C:\Go\src\net\http\server.go`
The container file is here: `/usr/local/go/src/net/http/server.go`

But it looked here: `c:\Users\rpfister\usr\local\go\src\net\http\server.go`, which is home on the local machine (`C:\Users\rpfister`) added to the remote path (`\local\go\src\net\http\server.go`) to create a path that doesn't exist anywhere.

Additional info:
The local host’s GOROOT is `C:\Go`.
The local host's GOPATH is `C:\Users\rpfister\go`.
The container’s GOROOT is `/go`.
The container's GOPATH is undefined.

#### To reproduce

Clone the debugging branch of https://github.com/lukehoban/webapp-go/tree/debugging

Change line 1 of Dockerfile from "FROM golang:1.6" to "FROM golang:1.7" (The docker build fails using golang:1.6.)

Change `"host": "192.168.99.100"` to `"host: "localhost"` in line 29 of .vscode/launch.json (assuming you are running on the same machine).

Follow instructions in the comments in the launch.json; i.e., run the following two commands from the command prompt.

```
$ docker build -t webapp-go .   
$ docker run -d --name webapp-go --privileged -p 8080:8080 -p 2345:2345 webapp-go
```

Set breakpoint to line 37 in main.go

Launch the "Remote debug in Docker" debug configuration.

Step over repeatedly until the error is encountered.

##### Delve
```
PS C:\Users\rpfister> dlv version
Delve Debugger
Version: 1.0.0-rc.1
Build: $Id: b6077a6a35c48a31716aad2ad3fdf77bbdf70abd $
```

##### Golang

```
PS C:\Users\rpfister> go version
go version go1.8.3 windows/amd64
```

##### Docker

```
PS C:\Users\rpfister> docker version
Client:
 Version:      17.06.1-ce
 API version:  1.30
 Go version:   go1.8.3
[...]

PR #1179, opened 2017-08-31
Map remote goroot to local goroot

ramya-rao-a:
Fix for #1178

See https://github.com/Microsoft/vscode-go/issues/1178#issuecomment-326179112

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -511,10 +511,10 @@ class GoDebugSession extends DebugSession {
 		// Fix for https://github.com/Microsoft/vscode-go/issues/1178
 		// When the pathToConvert is under GOROOT, replace the remote GOROOT with local GOROOT
 		if (!pathToConvert.startsWith(this.delve.remotePath)) {
-			let index = pathToConvert.indexOf('src');
+			let index = pathToConvert.indexOf(`${this.remotePathSeparator}src${this.remotePathSeparator}`);
 			let goroot = process.env['GOROOT'];
-			if (goroot) {
-				return path.join(goroot, ...pathToConvert.substr(index).split(this.remotePathSeparator));
+			if (goroot && index > 0) {
+				return path.join(goroot, pathToConvert.substr(index));
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -503,11 +503,21 @@ class GoDebugSession extends DebugSession {
...
 		if (this.delve.remotePath.length === 0) {
-			return this.convertDebuggerPathToClient(path);
+			return this.convertDebuggerPathToClient(pathToConvert);
 		}
-		return path.replace(this.delve.remotePath, this.delve.program).split(this.remotePathSeparator).join(this.localPathSeparator);
+
+		// Fix for https://github.com/Microsoft/vscode-go/issues/1178
+		// When the pathToConvert is under GOROOT, replace the remote GOROOT with local GOROOT
+		if (!pathToConvert.startsWith(this.delve.remotePath)) {
+			let index = pathToConvert.indexOf('src');
```
roblourens: This is supposed to be `/src/` in the goroot, right? It could match a different src, like `/Users/src/goroot/src/...`

src/debugAdapter/goDebug.ts
```diff
@@ -503,11 +503,21 @@ class GoDebugSession extends DebugSession {
...
+			let goroot = process.env['GOROOT'];
+			if (goroot) {
+				return path.join(goroot, ...pathToConvert.substr(index).split(this.remotePathSeparator));
+			}
+		}
+		return pathToConvert.replace(this.delve.remotePath, this.delve.program).split(this.remotePathSeparator).join(this.localPathSeparator);
```
roblourens: You use `path.join` above and `.join(this.localPathSeparator);` here, should it be the same?

ramya-rao-a: path.join uses the path separator for the local machine so it would be the same `localPathSeparator`
additionally it also takes care of trailing `/` at the end of goroot if there is any.

So if I do a split on the goroot, remove trailing `/` if it exists, then I can use `.join(this.localPathSeparator)`

src/debugAdapter/goDebug.ts
ramya-rao-a: Updated to match `/src/`

Issue #647, opened 2016-11-24
Incorrect import by auto-complete when pkgname != last part of pkg path


AlekSi:
That's a continuation of #613. Get code from https://github.com/AlekSi/vscode-bug and then edit main.go as shown:

![2016-11-24_16-11-38](https://cloud.githubusercontent.com/assets/11512/20599753/e87a5236-b260-11e6-8fca-847d2cf44c6f.png)

Two things are wrong after hitting Enter:
1. It adds `sarama.v1` to code. As I explained in #613, the package name is `sarama`, not `sarama.v1`. Package name != last part of import path.
2. It adds `vendor/gopkg.in/Shopify/sarama.v1` which is incorrect import path. `vendor/` is [treated differently](https://golang.org/cmd/go/#hdr-Vendor_Directories) by Go toolchain. goreturns does the right thing:
```diff
$ goreturns -d src/bugger/cmd/bugger/main.go
diff src/bugger/cmd/bugger/main.go gofmt/src/bugger/cmd/bugger/main.go
--- /var/folders/fl/c3xvc0tj1js__3jnrnbkjll40000gn/T/gofmt002523159	2016-11-24 16:22:09.000000000 +0300
+++ /var/folders/fl/c3xvc0tj1js__3jnrnbkjll40000gn/T/gofmt440158346	2016-11-24 16:22:09.000000000 +0300
@@ -1,5 +1,7 @@
 package main

+import sarama "gopkg.in/Shopify/sarama.v1"
+
 func main() {
 	sarama.V0_10_0_0
 }
```

ramya-rao-a:
@AlekSi There are 2 separate things at play here.

- Auto-complete completes `sara` to `sarama.v1` instead of `sarama`. This will happen to all packages from gopkg.in due to the versioning supported there. 
- `vendor` included in the import path. This will happen only when the `vendor` folder is the immediate child of $GOPATH/src.

For the first issue, I have a PR out: #659

For the second, I am curious. Is it very common to have the vendor folder right under #GOPATH/src ?
Most of the projects I have seen, have the vendor folder under the Go project.

AlekSi:
> Auto-complete completes sara to sarama.v1 instead of sarama. This will happen to all packages from gopkg.in due to the versioning supported there.

It has nothing to do with `v1` and gopkg.in. For example, try with `go-xerial-snappy`:

[...]

PR #1191, opened 2017-09-04
Show completion for newly created package.

uudashr:
Previously the `goListAll()` cache the packages, resulting completion for newly created package not shown because the outdated cache and it quite irritating.

To takeout cache, we need to invoke `go list all` on demand. But the execution time is unacceptable, it will try to load packages from GOROOT, GOPATH and also user home directory. That is why `cp.spawn('go', [], { env: process.env })` took so long, and by not providing the env will execute faster.

Better implementation package listing tools `https://github.com/haya14busa/gopkgs`. It search the packages from golang `src`. So I replace the `go list all` using `gopkgs`.

`go list all` took 4.87s
`gopkgs` took 0.08s

- No more cache on `goListAll()` and it fast.
- Browse package also can lookup non-installed package (based on src, not the binary)
- The extension will check if the gopkgs is not the expected one and prompt as missing tool

### DIFF
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -26,8 +26,8 @@ export function listPackages(excludeImportedPkgs: boolean = false): Thenable<str
 
 	return Promise.all([pkgsPromise, importsPromise]).then(([pkgMap, importedPkgs]) => {
 		let pkgs = Array.from(pkgMap.keys());
-		pkgs = pkgs.filter(pkg => importedPkgs.indexOf(pkg) === -1)
-		return Array.from(new Set(pkgs)).sort()
+		pkgs = pkgs.filter(pkg => importedPkgs.indexOf(pkg) === -1);
+		return Array.from(new Set(pkgs)).sort();
 	});
 }
 
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -20,7 +20,7 @@ export function listPackages(excludeImportedPkgs: boolean = false): Thenable<str
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
 	let vendorSupportPromise = isVendorSupported();
 	let goPkgsPromise = new Promise<string[]>((resolve, reject) => {
-		let cmd = cp.spawn(getBinPath('gopkgs'), ['-short=false'], { env: getToolsEnvVars() });
+		let cmd = cp.spawn(getBinPath('gopkgs'), ['-f', '{{.ImportPath}}'], { env: getToolsEnvVars() });
 		let chunks = [];
 		let err;
 		cmd.stdout.on('data', (d) => chunks.push(d));
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -11,8 +11,6 @@ import { parseFilePrelude, isVendorSupported, getBinPath, getCurrentGoPath, getT
 import { documentSymbols } from './goOutline';
 import { promptForMissingTool } from './goInstallTools';
 import path = require('path');
-import { getRelativePackagePath } from './goPackages';
-import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
 import { getImportablePackages } from './goPackages';
 
 const missingToolMsg = 'Missing tool: ';
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -20,7 +20,7 @@ export function listPackages(excludeImportedPkgs: boolean = false): Thenable<str
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
 	let vendorSupportPromise = isVendorSupported();
 	let goPkgsPromise = new Promise<string[]>((resolve, reject) => {
-		cp.execFile(getBinPath('gopkgs'), [], {env: getToolsEnvVars()}, (err, stdout, stderr) => {
+		cp.execFile(getBinPath('gopkgs'), ['-short=false'], {env: getToolsEnvVars()}, (err, stdout, stderr) => {
 			if (err && (<any>err).code === 'ENOENT') {
 				return reject(missingToolMsg + 'gopkgs');
 			}
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -19,15 +19,13 @@ const missingToolMsg = 'Missing tool: ';
 
 export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
-
-	let currentFileDirPath = path.dirname(vscode.window.activeTextEditor.document.fileName);
-	let currentWorkspace = getCurrentGoWorkspaceFromGOPATH(getCurrentGoPath(), currentFileDirPath);
 	let pkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName);
 
 	return Promise.all([pkgsPromise, importsPromise]).then(([pkgMap, importedPkgs]) => {
-		let pkgs = Array.from(pkgMap.keys());
-		pkgs = pkgs.filter(pkg => importedPkgs.indexOf(pkg) === -1);
-		return Array.from(new Set(pkgs)).sort();
+		importedPkgs.forEach(pkg => {
+			pkgMap.delete(pkg);
+		});
+		return Array.from(pkgMap.keys()).sort();
 	});
 }
 
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -16,9 +16,7 @@ import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
 
 const missingToolMsg = 'Missing tool: ';
 
-let allPkgsCache: string[];
-
-function goPkgsNoCache(): Promise<string[]> {
+function goPkgs(): Promise<string[]> {
 	return new Promise<string[]>((resolve, reject) => {
 		let cmd = cp.spawn(getBinPath('gopkgs'), ['-f', '{{.ImportPath}}'], { env: getToolsEnvVars() });
 		let chunks = [];
@@ -33,19 +31,6 @@ function goPkgsNoCache(): Promise<string[]> {
 			let lines = chunks.join('').split('\n').filter((pkg) => pkg);
 			return resolve(lines);
 		});
-	});
-}
-
-function goPkgs(): Promise<string[]> {
-	if (allPkgsCache) {
-		goPkgsNoCache().then((pkgs: string[]) => {
-			allPkgsCache = pkgs;
-		});
-		return Promise.resolve(allPkgsCache);
-	}
-
-	return goPkgsNoCache().then((pkgs: string[]) => {
-		return allPkgsCache = pkgs;
 	});
 }
 
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -16,10 +16,10 @@ import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
 
 const missingToolMsg = 'Missing tool: ';
 
-export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
-	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
-	let vendorSupportPromise = isVendorSupported();
-	let goPkgsPromise = new Promise<string[]>((resolve, reject) => {
+let allPkgsCache: string[];
+
+function goPkgsNoCache(): Promise<string[]> {
+	return new Promise<string[]>((resolve, reject) => {
 		let cmd = cp.spawn(getBinPath('gopkgs'), ['-f', '{{.ImportPath}}'], { env: getToolsEnvVars() });
 		let chunks = [];
 		let err;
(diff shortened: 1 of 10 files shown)


uudashr:
@ramya-rao-a check please

src/goPackages.ts
```diff
@@ -19,37 +12,31 @@ export function isGoListComplete(): boolean {
...
 		cmd.on('close', (status) => {
-			chunks.join('').split('\n').forEach(pkgDetail => {
-				if (!pkgDetail || !pkgDetail.trim() || pkgDetail.indexOf(';') === -1) return;
-				let [pkgName, pkgPath] = pkgDetail.trim().split(';');
-				allPkgs.set(pkgPath, pkgName);
+			// TODO do we need to cache?
+			let pkgs = new Map<string, string>();
+			chunks.join('').split('\n').forEach((pkgPath) => {
+				if (!pkgPath) return;
+				const lastIndex = pkgPath.lastIndexOf('/');
```
ramya-rao-a: The main reason I used `go list` instead of `gopkgs` initially was to get the package name and not rely on the import path of the package to determine the package name like you are doing here.

Because package names need not necessarily be the last part of the package path. See https://github.com/Microsoft/vscode-go/issues/647

And we use the package names to provide completion for unimported packages.

uudashr: I see... make sense, because it completion, not import package

ramya-rao-a: Yes, and later on I added the `Go: Browse Packages` feature because then I had all the packages in hand, so it seemed like a good feature to add

uudashr: We can strip off the version part from "pkgname.v1" , it works when people always using standard way on naming package related to the package path, but that doesn't guarantee

ramya-rao-a: That's exactly what I did in https://github.com/Microsoft/vscode-go/commit/cb0d2bc29a0c9ee1929c1a89781ef8c740447d92 :)

uudashr: [cb0d2bc](https://github.com/Microsoft/vscode-go/commit/cb0d2bc29a0c9ee1929c1a89781ef8c740447d92) no longer in master?

Hare are some example, can't guarantee consistent pattern:
- speech;google.golang.org/api/speech/v1beta1
- spectrum;google.golang.org/api/spectrum/v1explorer
- speech;google.golang.org/api/speech/v1
- prediction;google.golang.org/api/prediction/v1.2
- tilde;gopkg.in/mattes/go-expand-tilde.v1

So using `go list -f '{{.Name}};{{.ImportPath}}' all` is the proper way, it just need to make it faster.

Case
1. Autocomplete un-imported will suggest only packages on cache, no way to suggest new package since it need invalidation. Need to listen some events to invalidates, haven't figure out which one.
2. Go: Browse Packages use `https://github.com/tpng/gopkgs` which looks in `$GOPATH/pkg`, new package will not exists in `$GOPATH/pkg`, so browse package will not help. Except we change it, use `https://github.com/haya14busa/gopkgs` which search in   `$GOPATH/src`. This is workaround, but still completion still stale.

Btw, why not change to `https://github.com/haya14busa/gopkgs` for Browse Package? It seems good improvement, solve one problem.

ramya-rao-a: Yes, I replaced https://github.com/Microsoft/vscode-go/commit/cb0d2bc29a0c9ee1929c1a89781ef8c740447d92 with using `go list -f '{{.Name}};{{.ImportPath}}' all`

I am fine with auto complete using cache as it does today. 

I didnt use `https://github.com/tpng/gopkgs` for the `Browse Packages` feature as it will not have the "main" packages as in the ones that are executable

I am fine with using `https://github.com/haya14busa/gopkgs` for just the Browse Packages feature as well

uudashr:
@ramya-rao-a latest changes should be able to show correct package name

uudashr:
Travis build failed on every linux machine
```
Xlib:  extension "RANDR" missing on display ":99.0".
{ errorCode: 'load',
  moduleId: 'keytar',
  neededBy: [ 'vs/platform/credentials/node/credentialsService' ],
  detail: 
   Error: libsecret-1.so.0: cannot open shared object file: No such file or directory
       at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:173:20)
       at Object.Module._extensions..node (module.js:598:18)
       at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:173:20)
       at Module.load (module.js:488:32)
       at tryModuleLoad (module.js:447:12)
       at Function.Module._load (module.js:439:3)
       at Module.require (module.js:498:17)
       at n (/home/travis/gopath/src/github.com/Microsoft/vscode-go/.vscode-test/VSCode-linux-x64/resources/app/out/vs/loader.js:4:10954)
       at Object.<anonymous> (/home/travis/gopath/src/github.com/Microsoft/vscode-go/.vscode-test/VSCode-linux-x64/resources/app/node_modules/keytar/lib/keytar.js:1:169)
       at Object.<anonymous> (/home/travis/gopath/src/github.com/Microsoft/vscode-go/.vscode-test/VSCode-linux-x64/resources/app/node_modules/keytar/lib/keytar.js:55:3) }
```

@ramya-rao-a do you know anything about this?

ramya-rao-a:
@chrmarti https://github.com/Microsoft/vscode-go/pull/1191#issuecomment-327986633 looks like its related to the keytar work you have been doing

@uudashr Thanks for all the work on this PR. I'll take a look at it over the weekend

chrmarti:
@uudashr `libsecret-1` seems to be missing, you can add that to your travis config.

uudashr:
@chrmarti thanks... works as expected

src/goSuggest.ts
```diff
@@ -235,6 +230,9 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 			});
 		});
 
+		return Promise.all([setPkgsList, setGocodeProps]).then(() => {
```
ramya-rao-a: With the `-format` option I am guessing that `gopkgs` is a bit slower than when you made the comparison between `go list all` and `gopkgs` in the description of this PR.

What are your current numbers?

Every new word typed triggers completion. So you can imagine that `gopkgs` will be run for every word typed. 
Some users have a whole lot of packages that `go list all` took over 10 seconds to finish. How does `gopkgs -format` fare?

I still believe we should use cache and have the cache updated at a regular interval instead of running `gopkgs` for every word.

What do you think?

src/goPackages.ts
```diff
@@ -2,54 +2,30 @@ import vscode = require('vscode');
...
-
-		const cmd = cp.spawn(goRuntimePath, ['list', '-f', '{{.Name}};{{.ImportPath}}', 'all'], { env: env, stdio: ['pipe', 'pipe', 'ignore'] });
+	return new Promise<Map<string, string>>((resolve, reject) => {
+		const cmd = cp.spawn(getBinPath('gopkgs'), ['-format', '{{.Name}};{{.ImportPath}}'], { env: getToolsEnvVars() });
 		const chunks = [];
 		cmd.stdout.on('data', (d) => {
 			chunks.push(d);
 		});
 
 		cmd.on('close', (status) => {
```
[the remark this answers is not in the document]
uudashr: Will check foe the missing tool when execution, but cannot detect failure on flag, because the old gopkgs doesn't care with any flags or arguments.

The only way to check it use different gopkgs version is by read the help provided by the binary.

src/goSuggest.ts
uudashr: `gpkgs -format` faster than `go list all`

For 3382 packages packages:
```
go list all  4.62s user 3.27s system 124% cpu 6.360 total
gopkgs  0.65s user 0.70s system 264% cpu 0.508 total
```
6.36s vs 0.5s, big difference.
This gopkgs using fastwalk, faster than golang standard `fs.Walk()`.

Not every word typed, it only invoked on only when completion triggered. Typing "strings" 7 words typed will only cost single `gopkgs` invocation. So this is quite cheap and fast enough.

Cache will be much better of course.
If we cache for 10 or 20 second, user probably can create new file in new folder with new package faster than cache expiry time. So new package will not provided by completion. The workaround crossed in my mind is we need to some kind of file system watcher and check if there is changed file, then invalidate the cache or run `gopkgs` immediately to refresh the cache, etc. My suggestion is we can do this cache things on another PR.

src/goPackages.ts
```diff
@@ -113,25 +90,31 @@ export function getRelativePackagePath(currentFileDirPath: string, currentWorksp
 }
 
 /**
- * Returns import paths for all non vendor packages under given folder
+ * Returns import paths for all packages under given folder (vendor will be excluded)
  */
-export function getNonVendorPackages(folderPath: string): Promise<string[]> {
+export function getPackages(folderPath: string): Promise<string[]> {
```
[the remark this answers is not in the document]
uudashr: Done

src/goPackages.ts
```diff
@@ -2,54 +2,30 @@ import vscode = require('vscode');
...
-let goListAllCompleted: boolean = false;
-let goListAllPromise: Promise<Map<string, string>>;
-
-export function isGoListComplete(): boolean {
-	return goListAllCompleted;
-}
+import { isVendorSupported, getCurrentGoPath, getToolsEnvVars, getGoVersion, getBinPath, SemVersion } from './util';
 
 /**
  * Runs go list all
```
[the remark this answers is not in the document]
uudashr: Done

src/goImport.ts
```diff
@@ -16,27 +16,31 @@ import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
...
-		cp.execFile(getBinPath('gopkgs'), [], {env: getToolsEnvVars()}, (err, stdout, stderr) => {
+function goPkgs(): Promise<string[]> {
+	return new Promise<string[]>((resolve, reject) => {
+		let cmd = cp.spawn(getBinPath('gopkgs'), ['-format', '{{.ImportPath}}'], { env: getToolsEnvVars() });
+		let chunks = [];
+		let err;
+		cmd.stdout.on('data', (d) => chunks.push(d));
+		cmd.on('error', (e) => err = e);
+		cmd.on('close', () => {
 			if (err && (<any>err).code === 'ENOENT') {
```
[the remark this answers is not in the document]
uudashr: Good idea

uudashr:
@ramya-rao-a please check

src/goSuggest.ts
uudashr: Changed my mind... let's try to use 5s cache expiry

src/goPackages.ts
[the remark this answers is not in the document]
ramya-rao-a: That's unfortunate

src/goSuggest.ts
ramya-rao-a: > Not every word typed, it only invoked on only when completion triggered.

Completion is triggered for every word typed. I think you misunderstood my statement as every "letter" typed

src/goInstallTools.ts
```diff
@@ -282,9 +282,38 @@ export function offerToInstallTools() {
...
+			reject(err);
+		});
+	});
+}
+
 function getMissingTools(goVersion: SemVersion): Promise<string[]> {
 	let keys = Object.keys(getTools(goVersion));
 	return Promise.all<string>(keys.map(tool => new Promise<string>((resolve, reject) => {
+		if (tool === 'gopkgs') {
+			gopkgsMissing().then((missing) => {
```
ramya-rao-a: This line will be reached only when there is no `gopkgs` tool in the user's machine.
Whereas you are expecting to reach here to make a decision on whether the `gopkgs` present in the user machine is the old one or the new one

ramya-rao-a: What do you think of this?
https://github.com/ramya-rao-a/vscode-go/commit/2119ec1849f18a0e2f9d5eae348a0d7c1e9cdd3f

src/goSuggest.ts
uudashr: Ahh... yup..

src/goInstallTools.ts
uudashr: Ahhh... I see your point. Agree with that. We can remove the `gopkgsMissing()` then

ramya-rao-a: Ok, I have pushed my change to your branch

ramya-rao-a:
@uudashr The below assert is failing when vendor is not supported. Any idea why?

https://github.com/uudashr/vscode-go/blob/faster-go-list-all/test/go.test.ts#L590

uudashr:
@ramya-rao-a Not sure, still investigating...
For sure because now `listPackages()` use `getImportablePackages()`.
Logic from old `listPackages()` not compatible with `getImportablePackages()`
And it happens only on Go 1.5 and 1.6

ramya-rao-a:
@uudashr Tests are fixed now

ramya-rao-a:
There are a few things I still want to discuss
1. We don't actually need the gopkgs to complete to start the first call to `gocode`. I propose this: https://github.com/ramya-rao-a/vscode-go/commit/da5a3cdcef2da144eb418c4b4c3d8b76cc625d36
2. If gopkgs for any reason runs more than 5 seconds, then every word typed after 5 seconds will end up creating a new process. We should safeguard against this. Something like "if a current gopkgs process is running, don't start a new one"
3. Completions are usually expected to show up under a second. How do we ensure this when gopkgs takes more than a second?

uudashr:
You are welcome... thanks for fixing the test

1. Cool
2. Have something in my mind.. something like create subscription to the currently running process if any
3. Something like `Promise.race([goAllPackagesNoCache(), timeout(1000).then(() => cache)])`

uudashr:
@ramya-rao-a for point number 2, roughly will be going to be like https://gist.github.com/uudashr/8fbacf975cb8140b962fc7fd3122ff22

Issue #1186, opened 2017-09-02
Suggestions for "log" don't show up before it's imported


roblourens:
I tried several unimported packages, and autocomplete works for them, but it doesn't work just for `log` for some reason.

e.g. `fmt.` -> Shows `Errorf`, `Printf`, etc.

`log.` -> nothing...

Once I import `log`, its suggestions show up.

ramya-rao-a:
<img width="474" alt="screen shot 2017-09-04 at 10 37 39 pm" src="https://user-images.githubusercontent.com/16890566/30046942-2d912650-91c3-11e7-8efb-35bda50f1441.png">

roblourens:
![image](https://user-images.githubusercontent.com/323878/30067363-6f0cbaa8-920f-11e7-8bdc-0157c43a8ae5.png)

sgyang:
So do I. (code 1.15.1, vscode-go 0.6.64)
It may be related to #1152. I set "go.autocompleteUnimportedPackages" to true but it didn't work.
![go_fmt](https://user-images.githubusercontent.com/1253518/30093220-81311c5a-92f6-11e7-8e5d-a913c1842df5.png)
![go_log](https://user-images.githubusercontent.com/1253518/30093228-87109826-92f6-11e7-9dfb-415b29fe3334.png)

sgyang:
OK I found version 0.6.62 worked properly. 0.6.63 and 0.6.64 didn't work.

FYI: I did:

- Remove current vscode-go
- Disable auto-update feature of extensions (`"extensions.autoUpdate": false`) 
- Download previous version from https://lukehoban.gallery.vsassets.io/_apis/public/gallery/publisher/lukehoban/extension/go/0.6.62/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage
- Rename it to go-0.6.62.vsix
- Install this by `code --install-extension go-0.6.62.vsix`

And then the issue is fixed.

ramya-rao-a:
@sgyang Do you also the issue only for the `log` pkg? And `fmt` works?

sgyang:
@ramya-rao-a The issue is related to all unimported packages. This `log` and `fmt` thing is just an example.
Before importing `log`, hitting `log` does not show suggestions, but after importing `log`, hitting `log` fires auto-completion. Previously, auto-completion works properly without importing packages (and I love this feature).
[...]

PR #1197, opened 2017-09-06
Ignore stderr outputs of go list all to fix #1152

### LINKED ISSUE #1152: "go.autocompleteUnimportedPackages" doesn't work after update to VSCode 1.15
My editor can not using "autocomplete Unimported Packages" any more, after I update my vscode to 1.15, I'm not sure if it is the reason.

Previously I can import package like 'strings' just type "strings.S" and auto complete list will show with "strings.Splite()" and after select, the package 'strings' will be imported.

But now, if I didn't import the package "strings", the tips won't show.

It's very pleasure if any one can help.

My OS: macOS Sierra 10.12.6
VSCode: 1.15.1 (1.15.1) , just update this morning, but the problem is not just occur now, it has been occur for about 2 weeks.

If any other info is needed, please let me know ! Thanks a lot

sgyang:
If `go list all` emits warning messages on stderr, it seems stuck forever, and then auto completion (before importing) feature does not work. So just ignore the outputs of stderr.
It fixes #1152. And it's maybe related to #1186.

### DIFF
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -32,7 +32,7 @@ export function goListAll(): Promise<Map<string, string>> {
 		// Use `{env: {}}` to make the execution faster. Include GOPATH to account if custom work space exists.
 		const env: any = getToolsEnvVars();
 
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Thanks @sgyang!

PR #1208, opened 2017-09-12
Ensure only single gopkgs running

uudashr:
Avoid multiple gopkgs running


uudashr:
@ramya-rao-a maybe like this

ramya-rao-a:
Nice work @uudashr!

PR #1209, opened 2017-09-12
Ensure completion to respond faster

uudashr:
I put some guard for 1sec timeout for `getImportablePackages()`

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -8,7 +8,7 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import { dirname, basename } from 'path';
-import { getBinPath, parameters, parseFilePrelude, isPositionInString, goKeywords, getToolsEnvVars } from './util';
+import { getBinPath, parameters, parseFilePrelude, isPositionInString, goKeywords, getToolsEnvVars, timeout } from './util';
 import { promptForMissingTool } from './goInstallTools';
 import { getTextEditForAddImport } from './goImport';
 import { getImportablePackages } from './goPackages';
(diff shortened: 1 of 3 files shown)


uudashr:
@ramya-rao-a please check

src/goSuggest.ts
```diff
@@ -214,7 +215,12 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
 	// TODO: Shouldn't lib-path also be set?
 	private ensureGoCodeConfigured(): Thenable<void> {
-		let setPkgsList = getImportablePackages(vscode.window.activeTextEditor.document.fileName).then(pkgMap => {
+		let importablePkgsPromise: Promise<Map<string, string>> = getImportablePackages(vscode.window.activeTextEditor.document.fileName);
+		if (this.pkgsList.size) {
+			importablePkgsPromise = Promise.race([timeout(1000).then(() => this.pkgsList), importablePkgsPromise]);
+		}
+
+		let setPkgsList = importablePkgsPromise.then(pkgMap => {
 			this.pkgsList = pkgMap;
```
ramya-rao-a: How about this?
```
let importablePkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName).then(pkgMap => this.pkgList = pkgMap);

let setpkgslist = Promise.race([timeout(1000).then(() => this.pkgsList), importablePkgsPromise]);
```

Case when gopkgs runs under a second: 
Completion will always use the latest result

Case when gopkgs runs over a second: 
Completion will use cache. If its the first completion call, then no pkgs will be available which is fine, we can attribute this to cold start
When gopkgs eventually finishes, it will update the result, so that the next completion gets to use a more recent cache

ramya-rao-a: The reason I am suggesting the above is that with your current design 
- a user can still get stuck waiting for the very first call to completion
- a user can get stale data if gopkgs always takes over 1 second, and the next call from completion happens after 5 sec (the cache expiry time)

uudashr: Ahh... I see

ramya-rao-a:
@uudashr The `Test listPackages` test timesout and so fails for the tip consistently in travis ( I re-ran a couple of times).

I dont have Go tip, you have any ideas about this?

uudashr:
Tip version on Mac looks good on my machine. Re-run the travis build and all good.

PR #1219, opened 2017-09-16
📚 Fix misspelling of "editor context menu" setting

hankfanchiu:
[`go.editorContextMenuCommands`](https://github.com/Microsoft/vscode-go/blob/master/package.json#L691) is misspelled as `go. editortorContextMenuCommands` in the README.


ramya-rao-a:
Thanks @hankfanchiu!

PR #1222, opened 2017-09-18
Add command to fetch import on current line

juicemia:
This command allows users to just fetch a dependency from their editor. Under the hood, all it does is run `go get -u` for the import path on the current line.

**NOTE:** This is not a replacement for dependency management tools like dep.


juicemia:
I see that Travis is failing but I don't know why. @ramya-rao-a can you see why? I have very little experience with Travis.

ramya-rao-a:
@juicemia Its just the linter failing
image

Run `npm run lint` on your branch and you will be able to see the same errors as well

package.json
```diff
@@ -169,6 +169,11 @@
         "command": "go.browse.packages",
         "title": "Go: Browse Packages",
         "description": "Browse packages and Go files inside the packages."
+      },
+      {
+        "command": "go.get.import",
+        "title": "Go: Get Import on Curent Line",
```
ramya-rao-a: How about `Go: Import Package`?  `Go: Install Package` ? `Go: Get Package` ? `Go: Install Imported Package`
(naming is seriously the hardest thing to do)

And the command  would be more intuitive if it had the word "package" like `go.get.package` or `go.import.package`

Also update the description. It should be clear that we run a `go get -u` on the package

src/goGetImport.ts
```diff
@@ -0,0 +1,37 @@
...
+	const selectedText = editor.document.lineAt(selection.active.line).text;
+
+	const importPath = getImportPath(selectedText);
+
+	const goRuntimePath = getGoRuntimePath();
+	if (!goRuntimePath) {
+		return vscode.window.showErrorMessage('Could not locate Go binaries. Make sure you have Go installed');
+	}
+
+	fs.exists(path.join(getCurrentGoPath(), 'src', importPath), (exists) => {
```
ramya-rao-a: we can skip this check and run `go get -u` regardless
If the package already exists then it will be updated.
Thoughts?

src/goGetImport.ts
```diff
@@ -0,0 +1,37 @@
...
+			return;
+		}
+
+		cp.execFile(goRuntimePath, ['get', '-u', importPath], (err, stdout, stderr) => {
+			if (stderr !== '') {
+				vscode.window.showErrorMessage(stderr);
+				return;
+			}
+
+			vscode.window.showInformationMessage(`Successfully fetched package ${importPath}`);
```
ramya-rao-a: What do you think of showing the progress/success/failure messages in the output pane just like the current command `Go: Install/Update Tools`

src/goGetImport.ts
```diff
@@ -0,0 +1,37 @@
...
+import fs = require('fs');
+import { getGoRuntimePath } from './goPath';
+import { getCurrentGoPath, getImportPath } from './util';
+
+export function goGetImport() {
+	const editor = vscode.window.activeTextEditor;
+	const selection = editor.selection;
+	const selectedText = editor.document.lineAt(selection.active.line).text;
+
+	const importPath = getImportPath(selectedText);
```
ramya-rao-a: If user runs this command on a line thats not an import statement, shouldnt we exit earlier? We know `go get` would fail then.

package.json
juicemia: I think `Go: Get Package` is the best one out of all of them, because it follows the semantics of using `go get`. I'll change that now.

src/goGetImport.ts
juicemia: I think it's a good idea, if we make sure to let the user know that their package will be updated. They might not want that for whatever reason.

src/goGetImport.ts
juicemia: I think that's a great idea.

src/goGetImport.ts
juicemia: Yup absolutely. Forgot that check.

src/goGetImport.ts
juicemia: One thing I just thought of: `go get -u` doesn't output anything by default unless it fails. For this to be useful, we would have to run `go get -u -v`. We could also show the output pane only if the command fails. Thoughts?

test/go.test.ts
```diff
@@ -738,4 +739,19 @@ It returns the number of bytes written and any write error encountered.
 			assert.ok(false, `error in OpenTextDocument ${err}`);
 		}).then(() => done(), done);
 	});
+
+	test('getImportPath()', () => {
```
ramya-rao-a: may I just say that I absolutely love it that you have added tests 🥇

package.json
ramya-rao-a: Agreed

src/goGetPackage.ts
```diff
@@ -0,0 +1,41 @@
...
+		vscode.window.showErrorMessage('No import path to get');
+		return;
+	}
+
+	const goRuntimePath = getGoRuntimePath();
+	if (!goRuntimePath) {
+		return vscode.window.showErrorMessage('Could not locate Go binaries. Make sure you have Go installed');
+	}
+
+	fs.exists(path.join(getCurrentGoPath(), 'src', importPath), (exists) => {
```
ramya-rao-a: getCurrentGoPath() can return `:` or `;` separated paths as well. This would fail then

src/goGetImport.ts
ramya-rao-a: Agreed

ramya-rao-a:
@juicemia I went ahead and rebased your branch from master as there were some conflicts. You might have to delete your local branch and checkout from the remote again.

Sorry for the trouble, I usually work alone, so didnt think of the troubles rebasing might cause others :(

src/goGetPackage.ts
ramya-rao-a: Instead you can run `go list` and pass the importpath to see if the package exists.
Or skip this check altogether.

I really feel that if a user has explicitly run this `Go: Get Package` command then they would know that this would get the package using `go get`

juicemia:
It's okay. I'll figure it out, no worries.

ramya-rao-a:
Thanks @juicemia!

PR #1232, opened 2017-09-20
Fix gopkgs tools

uudashr:
Fix the gopkgs install source path


uudashr:
This is what going to be happen:
1. Invalid tools installed
2. Then autocompletion cannot works
3. Prompt for update shown, but after that autocomplete still not working.
cc @ramya-rao-a

ramya-rao-a:
The old gopkgs got in when I merged the multi-root changes, Sorry abt that!

PR #1244, opened 2017-09-24
add auto save to tests on test run or debug

### LINKED ISSUE #1225: Can clicking "run tests" automatically save the file?
If I edit a test and click "run test", there is a dialog:

    File has unsaved changes. Save and try again.

It seems like it would be better to automatically save the file?

chaseadamsio:
Fixes #1225 by adding `editor.document.save()` if the document state is dirty. Also adds a `go.saveTestFileOnRunOrDebug` (defaults to `true`) to disable auto save in the case that someone wants to disable this feature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -565,12 +565,6 @@
           "description": "Flags to pass to `go test`. If null, then buildFlags will be used.",
           "scope": "resource"
         },
-        "go.saveTestFileOnRunOrDebug": {
-          "type": "boolean",
-          "default": true,
-          "description": "Automatically test files when 'Run Test' or 'Debug Test' is run.",
-          "scope": "resource"
-        },
         "go.toolsEnvVars": {
           "type": "object",
           "default": {},
--- a/package.json
+++ b/package.json
@@ -565,6 +565,12 @@
           "description": "Flags to pass to `go test`. If null, then buildFlags will be used.",
           "scope": "resource"
         },
+        "go.saveTestFileOnRunOrDebug": {
+          "type": "boolean",
+          "default": true,
+          "description": "Automatically test files when 'Run Test' or 'Debug Test' is run.",
+          "scope": "resource"
+        },
         "go.toolsEnvVars": {
           "type": "object",
           "default": {},
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -151,6 +151,7 @@ export function testCurrentFile(goConfig: vscode.WorkspaceConfiguration, args: s
 		return;
 	}
 
+	editor.document.save().then(() => {
 	return getTestFunctions(editor.document).then(testFunctions => {
 		const testConfig = {
 			goConfig: goConfig,
@@ -162,6 +163,7 @@ export function testCurrentFile(goConfig: vscode.WorkspaceConfiguration, args: s
 		lastTestConfig = testConfig;
 
 		return goTest(testConfig);
+		});
 	}).then(null, err => {
 		console.error(err);
 		return Promise.resolve(false);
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -565,6 +565,12 @@
           "description": "Flags to pass to `go test`. If null, then buildFlags will be used.",
           "scope": "resource"
         },
+        "go.saveTestFileOnRunOrDebug": {
```
ramya-rao-a: `go test` can only work on files on disk. Therefore, it is reasonable to say that the current file needs to be saved for any test in the current file to be run.

I don't see any reason why one would want to disable this setting.
Unless we have a scenario for it, I would say we don't need this setting.

src/goTest.ts
```diff
@@ -32,8 +32,12 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, args: any)
 		return;
 	}
 	if (editor.document.isDirty) {
-		vscode.window.showInformationMessage('File has unsaved changes. Save and try again.');
-		return;
+		if (goConfig['saveTestFileOnRunOrDebug'] === true) {
+			editor.document.save();
```
ramya-rao-a: I believe, `editor.document.save()` returns a promise. We should wait for the promise to complete before trying to run the tests.

src/goTest.ts
```diff
@@ -32,8 +32,12 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, args: any)
 		return;
 	}
 	if (editor.document.isDirty) {
```
ramya-rao-a: A little history...

The `run/debug test` codelens and the `Go: Test function at cursor` command required us to parse the file, find the name of the test method to pass the same to `go test`. This is why I had added the check on whether the file is dirty or not in the first place

Testing the package/workspace doesn't need us to do this

On the other hand testing current file also required us to parse the file and find the test methods.
So we should add the same "save" operation in the `testCurrentFile` method in this file

package.json
chaseadamsio: 👍

src/goTest.ts
chaseadamsio: Thanks for the background! 

Is it a correct understanding that all four methods then should still perform the save operation and now that we're performing it in `testAtCursor` the check for `editor.document.isDirty` is also no longer required?

src/goTest.ts
chaseadamsio: 👍 thanks! Will take a stab (it's been a _long_ time since I've used promises and this is my first `typescript` PR, so thanks for your patience) and re-submit.

src/goTest.ts
ramya-rao-a: Which 4 methods? Cursor, File, Package and Workspace?

Cursor and File should have the file save. Package and Workspace dont, as you can run those even without any file open.

Yes, you can skip the check on dirty file and perform the save directly for cursor and file.

src/goTest.ts
```diff
@@ -153,17 +151,19 @@ export function testCurrentFile(goConfig: vscode.WorkspaceConfiguration, args: s
...
-	return getTestFunctions(editor.document).then(testFunctions => {
-		const testConfig = {
-			goConfig: goConfig,
-			dir: path.dirname(editor.document.fileName),
-			flags: getTestFlags(goConfig, args),
-			functions: testFunctions.map(func => { return func.name; })
-		};
-		// Remember this config as the last executed test.
-		lastTestConfig = testConfig;
+	editor.document.save().then(() => {
```
ramya-rao-a: The tests are failing due to the missing `return` here.
The `testCurrentFile` method should return a promise

PR #1248, opened 2017-09-27
Fix buildFlags type in 'debug test' Codelens

ka2n:
Make the type of buildArgs executed by Codelens consistent with LaunchRequestArguments.
This fix the problem that `go.buildTags` is ignored by `debug test`.

### DIFF
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -81,8 +81,7 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
 				const envFile = vsConfig['testEnvFile'];
 				let buildFlags = getTestFlags(vsConfig, null);
 				if (vsConfig['buildTags'] && buildFlags.indexOf('-tags') === -1) {
-					buildFlags.push('-tags');
-					buildFlags.push(`"${vsConfig['buildTags']}"`);
+					buildFlags.push(`-tags '${vsConfig['buildTags']}'`);
 				}
 
 				let config = Object.assign({}, this.debugConfig, { args, program, env, envFile, buildFlags });
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -81,10 +81,11 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
 				const envFile = vsConfig['testEnvFile'];
 				let buildFlags = getTestFlags(vsConfig, null);
 				if (vsConfig['buildTags'] && buildFlags.indexOf('-tags') === -1) {
-					buildFlags.push(`-tags '${vsConfig['buildTags']}'`);
+					buildFlags.push('-tags');
+					buildFlags.push(`${vsConfig['buildTags']}`);
 				}
 
-				let config = Object.assign({}, this.debugConfig, { args, program, env, envFile, buildFlags });
+				let config = Object.assign({}, this.debugConfig, { args, program, env, envFile, buildFlags: buildFlags.map(x => `'${x}'`).join(' ') });
 				let debugTestCmd: Command = {
 					title: 'debug test',
 					command: 'vscode.startDebug',


src/goRunTestCodelens.ts
```diff
@@ -85,7 +85,7 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
 					buildFlags.push(`"${vsConfig['buildTags']}"`);
```
ramya-rao-a: Since this is related to buildtags specifically and not buildflags in general, shouldnt we do the join that you did on line 85 instead?

ka2n: Referring to [usage](https://github.com/Microsoft/vscode-go/blob/dd927509a3db3833dc02103e1cf6290074f56a4b/src/debugAdapter/goDebug.ts#L297) and [interface declaration](https://github.com/Microsoft/vscode-go/blob/dd927509a3db3833dc02103e1cf6290074f56a4b/src/debugAdapter/goDebug.ts#L138), it seems to assume a `string?`, but is this reference place wrong(or not just Delve class)?

ramya-rao-a:
Can you give me a sample code where I can test this?
Sorry I don't do a lot of Go coding myself and haven't used build tags much.

ka2n:
**Problem:**
I use build tags for excluding some tests when running on CI server.

`some_test.go`
```go
// +build noci

package some

import "testing"

func TestFoo(t *testing.T) {
 // A test always fail on CI sever
}
```

Only when I add `-tags noci` to `go test` parameters, the test will run.

```shell
$ go test .
?       github.com/ka2n/example/some [no test files]
$ go test -tags noci
ok     github.com/ka2n/example/some 0.007s
```

In VSCode, I want to add `-tags noci` every test run.

`<project_dir>/.vscode/settings.json`
```json
{
  "go.buildTags": "noci other_tag"
}
```

1. Run `run test` in codelens of `TestFoo`, it outputs,

```shell
Running tool: /usr/local/Cellar/go/1.9/libexec/bin/go test -timeout 30s -tags noci other_tag -run ^TestFoo$

PASS
ok  	github.com/ka2n/example/some	0.006s
Success: Tests passed.
```

2. Run `debug test` in codelens of `TestFoo`, error GUI are shown,
it says `Failed to continue: Check the debug console for details.`

To check final argument to launch delv,  log [`JSON.stringify(dlvArgs)` in this place](https://github.com/Microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts#L308).

It output: 
`["test","--headless=true","--listen=127.0.0.1:4887","--build-flags=-tags,\"noci other_tag\"","--","-test.run","TestFoo"]`

`-tags` are malformed from `["-tags", "\"noci other_tag\""]`(object) to `-tags,\"noci other_tag\"`(string).

To fix this problem I added [this](https://github.com/Microsoft/vscode-go/pull/1248/commits/efc1d3d4f1a7456ddcc8dcda5014591685454323) commit.


----

**Second problem** is something different situation, but I noted here for someone's reference.

I want to add `-ldflags`, and specify tags,
`<project_dir>/.vscode/settings.json`
```json
    "go.buildTags:" "noci other_tag",
    "go.buildFlags": [
        "-ldflags",
        "-X mypackage.myvalue=./"
    ]
```

run `debug test` in codelens of`TestFoo`
got error output in DEBUG CONSOLE
```
could not launch process: stat /Users/ka2n/src/github.com/msetsu/cointown/some/debug.test: no such file or directory
Process exiting with code: 1\
```

In this situation, `JSON.stringify(dlvArgs)` is `["test","--headless=true","--listen=127.0.0.1:24136","--build-flags=-ldflags,-X mypackage.myvalue=./,-tags,\"integration noci\"","--","-test.run","TestFoo"]`

I've try to solve this problem, got partial success, but it make `go.buildFlags` incompatible with normal `run test`.

ramya-rao-a:
I spent an hour trying to reproduce your first problem, but couldnt repro. Turns out this is not an issue in Windows. Then I tried on my Mac and was able to repro the issue instantly.

So, please confirm, the problem seems to be that delve requires buildFlags to be a single string which is space separated and `go test` requires it to be an array (because we use `cp.execFile`)

In that case, your initial solution of converting the entire `buildFlags` to string should help in both issues correct?

ka2n:
@ramya-rao-a 
Thank you for taking a time for this PR and I'm sorry I have not include platform information.

At first, my initial solution was not correct to solve entire problem.
I just write a patch to make `goDebug.ts: launchArgs.buildFlags` type and actual value match.

So, my latest commit is focus on just `go.buildTags`, I have to change the PR title :|.

---

For both issues correct
In the situation setting are below,

```json
{
  "go.buildFlags": [
    "-tags",
    "noci other_tag",
    "-ldFlags",
    "-X mypackage.myvalue=./"
  ]
}
```

- `go test` argument should: `["test", "-tags", "noci other", "-ldFlags", "-X mypackage.myvalue=./" ]`
- `dlv debug` argument should: `["--buildFlags", "-tags 'noci other_tag' -ldFlags '-X mypackage.myvalue=./'"]` (argument values are quoted)

But I can't find a way to filter argument value and quote it for Delv arguments.
And, now I think this problem is out of scope of this PR.

ramya-rao-a:
@ka2n 

Hey, my mistake, I am able to see the issue in Windows as well.
I had an old `debug.test` binary lying around and the debugger was using that so couldn't find the error before.
I cleaned my repo and now I am able to repro in Windows as well.

I made a commit to your branch for quoting both the flags and the flag values and that seem to work well for all cases
So your settings would be
```
"go.buildFlags": [
    "-ldflags",
    "-X mypackage.myvalue=./"
],
"go.buildTags: "noci"
```
Both "run tests" and "debug tests" codelens should now work.

Can you give it a try?

ka2n:
@ramya-rao-a 

I've tested with macOS(`10.13 (17A365)`), VSCode(`Version 1.17.0-insider (1.17.0-insider)`), vscode-go(https://github.com/Microsoft/vscode-go/commit/47c8b54fd93899e0bd5049c750e9cf2c3f051067) and works fine!

Thank you!

ramya-rao-a:
Awesome, thanks for working on this @ka2n 
I'll release an update later today

PR #1252, opened 2017-09-29
Adding option to not add types on function autocomplete. Fixes #1241

### LINKED ISSUE #1241: Add option for autocomplete to not add the types on function/method calls
Currently, when autocompleting a function or method the entire method signature, including types, is used. For example, in `strings.Split`, it currently autocompletes as `strings.Split(s string, sep string)` rather than `strings.Split(s, sep)`. In a lot of my code, I have variable names that are the same as argument names (i.e. my string is named `s`), and it would be convenient to only have to change the ones that are different.

alexkohler:
Allows a user to specify either `useCodeSnippetsOnFunctionSuggest` or `useCodeSnippetsOnFunctionSuggestWithoutType` for function auto completion. The former being set to true will autocomplete functions with the parameter signature including the variable type (as previously supported), and the latter will autocomplete functions with the parameter signature excluding the variable type. (Fixes #1241)


ramya-rao-a:
Thanks @alexkohler!

PR #1255, opened 2017-09-30
Fixed tests "Error checking" and "Gometalinter error checking"

ndrewnee:
If user set "go.lintOnSave" to false in User settings, then tests "Error checking" and "Gometalinter error checking" are failed


ramya-rao-a:
Thanks @ ndrewnee!

Issue #1256, opened 2017-10-01
Completion show internal package


uudashr:
internal package are not allowed, except it is from the same project. So should not suggested by autocompletion.

<img width="576" alt="screen shot 2017-10-01 at 1 37 40 pm" src="https://user-images.githubusercontent.com/166053/31053495-b876df02-a6c8-11e7-84b5-2d3f3c710930.png">
<img width="313" alt="screen shot 2017-10-01 at 1 37 59 pm" src="https://user-images.githubusercontent.com/166053/31053496-bcdc398e-a6c8-11e7-917a-1547c4a70927.png">
<img width="523" alt="screen shot 2017-10-01 at 2 31 19 pm" src="https://user-images.githubusercontent.com/166053/31053497-bf356426-a6c8-11e7-9285-de59a64e0d86.png">

PR #1257, opened 2017-10-01
Add internal package rules

uudashr:
Enforce internal package rules on autocompletion
Fixes for #1256

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -487,14 +487,11 @@ function didLangServerConfigChange(useLangServer: boolean, langServerFlags: stri
 	return false;
 }
 
-function loadPackages(): Promise<void> {
-	return new Promise<void>((resolve, reject) => {
-		vscode.window.withProgress<void>({location: vscode.ProgressLocation.Window, title: 'Loading packages'}, p => {
-			p.report({message: 'Load Packages'});
-			return getAllPackages().then(() => {
-				resolve();
-				return;
-			});
+function loadPackages(): Thenable<void> {
+	return vscode.window.withProgress<void>({location: vscode.ProgressLocation.Window, title: 'Loading packages'}, p => {
+		p.report({message: 'Load Packages'});
+		return getAllPackages().then(() => {
+			return;
 		});
 	});
 }
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -28,6 +28,7 @@ import { testAtCursor, testCurrentPackage, testCurrentFile, testPrevious, testWo
 import { showTestOutput } from './testUtils';
 import * as goGenerateTests from './goGenerateTests';
 import { addImport } from './goImport';
+import { getAllPackages } from './goPackages';
 import { installAllTools, checkLanguageServer } from './goInstallTools';
 import { isGoPathSet, getBinPath, sendTelemetryEvent, getExtensionCommands, getGoVersion, getCurrentGoPath, getToolsGopath } from './util';
 import { LanguageClient } from 'vscode-languageclient';
@@ -106,7 +107,10 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		}
 
 		if (vscode.window.activeTextEditor && isGoPathSet()) {
-			runBuilds(vscode.window.activeTextEditor.document, vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor.document.uri));
+			// preload packages so the cache are ready to use
+			loadPackages().then(() => {
+				runBuilds(vscode.window.activeTextEditor.document, vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor.document.uri));
+			});
 		}
 	});
 
@@ -481,4 +485,16 @@ function didLangServerConfigChange(useLangServer: boolean, langServerFlags: stri
 		}
 	}
 	return false;
+}
+
+function loadPackages(): Promise<void> {
+	return new Promise<void>((resolve, reject) => {
+		vscode.window.withProgress<void>({location: vscode.ProgressLocation.Window, title: 'Loading packages'}, p => {
+			p.report({message: 'Load Packages'});
+			return getAllPackages().then(() => {
+				resolve();
+				return;
+			});
+		});
+	});
 }
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -108,9 +108,8 @@ export function activate(ctx: vscode.ExtensionContext): void {
 
 		if (vscode.window.activeTextEditor && isGoPathSet()) {
 			// preload packages so the cache are ready to use
-			loadPackages().then(() => {
-				runBuilds(vscode.window.activeTextEditor.document, vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor.document.uri));
-			});
+			loadPackages();
+			runBuilds(vscode.window.activeTextEditor.document, vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor.document.uri));
 		}
 	});
 
@@ -487,11 +486,6 @@ function didLangServerConfigChange(useLangServer: boolean, langServerFlags: stri
 	return false;
 }
 
-function loadPackages(): Thenable<void> {
-	return vscode.window.withProgress<void>({location: vscode.ProgressLocation.Window, title: 'Loading packages'}, p => {
-		p.report({message: 'Load Packages'});
-		return getAllPackages().then(() => {
-			return;
-		});
-	});
+function loadPackages() {
+	getAllPackages();
 }
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -206,11 +206,6 @@ export function getNonVendorPackages(folderPath: string): Promise<string[]> {
 // see: https://golang.org/doc/go1.4#internalpackages
 // see: https://golang.org/s/go14internal
 export function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string) {
-	if (toDirPath === path.join(currentWorkspace, pkgPath)) {
-		// cannot import it's own package
-		return false;
-	}
-
 	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
 	if (internalPkgFound) {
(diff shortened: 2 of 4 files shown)


uudashr:
@ramya-rao-a what is the minimum Go version supported by `vscode-go`?
Go 1.5..?

ramya-rao-a:
@uudashr Yes we support Go 1.5

src/util.ts
```diff
@@ -481,3 +481,27 @@ export function guessPackageNameFromFile(filePath): Promise<string> {
 		});
 	});
 }
+
+// This will check whether it's regular package or internal package
+// Regular package will always allowed
+// Internal package only allowed if the package doing the import is within the tree rooted at the parent of "internal" directory
+// see: https://golang.org/doc/go1.4#internalpackages
+// see: https://golang.org/s/go14internal
+export function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string, goVersion: SemVersion) {
```
ramya-rao-a: Shouldn't this logic be inside `getImportablePackages()` method?

uudashr:
@ramya-rao-a does it support less than 1.4? internal package applied since 1.4. If vscode-go support 1.5 above then I'll remove go version checking

src/util.ts
```diff
@@ -481,3 +481,27 @@ export function guessPackageNameFromFile(filePath): Promise<string> {
...
+	if (toDirPath === path.join(currentWorkspace, pkgPath)) {
+		// cannot import it's own package
+		return false;
+	}
+
+	// internal package restriction applied since Go 1.4
+	if (goVersion.major < 1 || goVersion.major === 1 && goVersion.minor < 4) {
```
ramya-rao-a: We can skip the version check. We officially support Go 1.5 onwards. Let's keep it simple

src/util.ts
uudashr: I wish... but currently the completion using `this.pkgsList` cache in `GoCompletionItemProvider`.

Import internal package behavior works differently to each file (resident in a package) that try to import. So the set of packages would not be the same to each importing package. See https://github.com/Microsoft/vscode-go/pull/1257/files#diff-52de61e365f975d5f753902051296b23R289

src/util.ts
uudashr: Cool

src/util.ts
uudashr: Somehow not waiting for the `getImportablePackage` give illusion the completion was fast.. it just the completion would not give the full set of the packages for the first time. Somehow I like it because it fast. So if we stick with this, then the filtering need to be reside on the `GoCompletionItemProvider`.

ramya-rao-a: > Import internal package behavior works differently to each file (resident in a package) that try to import.

Its different for each package, but not for files in the same package correct?

The logic we have in place in `getImportablePackages` to return the relative paths for vendor pkgs also works differently to each package. Which is why it takes the `filePath` as a parameter.

If filtering for internals is inside the `GoCompletionProvider`, then the filtering for vendors should also be inside `GoCompletionProvider`

But I intentionally chose not to do so for vendors to keep things simple.

The cost of keeping the filtering out of `GoCompletionProvider` for users would be that the first time they edit a file inside a package, they might get suggestions from `vendor` or `internal` folders that they shouldn't get.

But the cache will catch up, and any further edits in the package will get self corrected.

ramya-rao-a:
> Import internal package behavior works differently to each file (resident in a package) that try to import.

Its different for each package, but not for files in the same package correct?

The logic we have in place in `getImportablePackages` to return the relative paths for vendor pkgs also works differently to each package. Which is why it takes the `filePath` as a parameter.

If filtering for internals is inside the `GoCompletionProvider`, then the filtering for vendors should also be inside `GoCompletionProvider`

But I intentionally chose not to do so for vendors to keep things simple.

The cost of keeping the filtering out of `GoCompletionProvider` for users would be that the first time they edit a file inside a package, they might get suggestions from `vendor` or `internal` folders that they shouldn't get.

But the cache will catch up, and any further edits in the package will get self corrected.

uudashr:
Correct.. same package (or same sub package)

I think `getImportablePackages` will return same result for a single project, haven't check more detail about this but `vendor` directory will be always the same, they will reside on the project directory. `vendor` package outside the project directory will be ignored. CMIIW. So valid for vendor. But agree we can move this to `GoCompletionProvider`

Agree with keep things simple.
I'm not very comfortable with existing filtering of `this.pkgsList` either, because filtering need to be done twice... see `getMatchingPackages` and `getPackagePathFromLine`. It fast but inefficient. Need to refactor later on.

So seems that `getImportablePackages` is no longer describe by it names, because not so importable until we consider the internal package.

Posible cases:
1. When user first time launch the vscode and start using completion, the suggestion will perfect, people assume it's first time load.
2. When user first time create new package, and it's not shown on completion yet, people can assume it's not loaded yet because of first time the package created.
4. But people who using a lot of packages.. and they switch between them, they will missed the internal package oftenly. The package was right there, not a new package.

My Case: for **non-library** project... I use a lot packages (sub packages) and don't want to share my code to other project, so I've plan to use internal package seems good solution to isolate the code. So switching package and look for internal package will be a huge cache missed.

But you may be right, need to test drive this. Maybe user won't realize the incomplete packages.

uudashr:
Or we can move the logic `isAllowToImport` to `getImportablePackages` and move the cache also on `getImportablePackages` so no more `this.pkgsList`.

uudashr:
@ramya-rao-a just understand with what you are saying.
I move the filtering responsibility to getImportablePackages and still using cache.
And the cache will catch up with latest updates.

uudashr:
@ramya-rao-a I was thinking to to load packages first time the vscode launch just to load the cache. Show on the bottom bar something like "Indexing packages...". User doesn't have to wait.. but they will realize if the completions hasn't show all the packages then it because the indexing is in progress.

And is it possible for the autocompletion... replace or update the completion items while it shown? Seems not possible.

src/goPackages.ts
```diff
@@ -97,11 +97,21 @@ export function getAllPackages(): Promise<Map<string, string>> {
...
+ * @param useCache. Force to use cache
  * @returns Map<string, string> mapping between package import path and package name
  */
-export function getImportablePackages(filePath: string): Promise<Map<string, string>> {
-	return Promise.all([isVendorSupported(), getAllPackages()]).then(([vendorSupported, pkgs]) => {
+export function getImportablePackages(filePath: string, useCache: boolean = false): Promise<Map<string, string>> {
+	let getAllPackagesPromise: Promise<Map<string, string>>;
+	if (useCache) {
+		// forced to use cache
+		getAllPackagesPromise = Promise.race([getAllPackages(), allPkgsCache]);
```
ramya-rao-a: When `allPkgsCache` is not set yet, then the Promise here waits for `getAllPackages()` to finish?

src/goMain.ts
```diff
@@ -106,7 +107,10 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		}
 
 		if (vscode.window.activeTextEditor && isGoPathSet()) {
-			runBuilds(vscode.window.activeTextEditor.document, vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor.document.uri));
+			// preload packages so the cache are ready to use
+			loadPackages().then(() => {
```
ramya-rao-a: why tie the preload to the initial running of builds? We can run the `getAllPackages()` independent of this correct?

src/goSuggest.ts
```diff
@@ -227,10 +227,10 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				});
 			});
 		});
-		return setGocodeProps;
-		// return Promise.all([setPkgsList, setGocodeProps]).then(() => {
-		// 	return;
-		// });
+		// return setGocodeProps;
+		return Promise.all([setPkgsList, setGocodeProps]).then(() => {
```
ramya-rao-a: So this would mean if there is any delay to due to `gopkgs` it will only affect the first auto-completion result. Right?

ramya-rao-a:
> And is it possible for the autocompletion... replace or update the completion items while it shown? Seems not possible.

No that's not possible

src/goPackages.ts
uudashr: It use `Promise.race`
`getAllPackages` has cache which is `allPkgsCache` and has it's own caching logic.
when `getAllPackages` use it's cache then it will selected by `Promise.race`
when not using cache then it will fallback to `allPkgsCache` which can be empty or outdated cache that is why I'm using `Promise.race([..., allPkgsCache])`

uudashr: it always return 'fast'
```js
function timeout(millis) {
  return new Promise((resolve, reject) => {
    setTimeout(resolve, millis);
  });
}

function fastResponse() {
  return new Promise((resolve, reject) => {
    resolve('fast');
  });
}

function slowResponse() {
  return timeout(500).then(() => 'slow');
}

const p = Promise.race([slowResponse(), fastResponse(), 'instant']);
p.then((v) => console.log(v));
```

src/goMain.ts
uudashr: Actually doesn't have to tie the initial running build. But it just UI/UX consideration (maybe while I'm not an expert in UX).

`runBuild()` will show "Tests Running".
While `loadPackages()` will show "Load Packages" and disappear when it done.
screen shot 2017-10-04 at 9 11 52 am
1. No separator... and it seems too crowd
2. When "Load Packages" disapear.. it _snaps_ the "Test Running" position to the left.

What do you think?

src/goSuggest.ts
uudashr: Yes... and `goImportablePackages(useCache=true)` so it should always use cache. Fast, doesn't have wait.
First time invocation can result to empty packages. But "preload packages" on startup will avoid the empty packages.

src/goMain.ts
ramya-rao-a: The tests running only shows up if `go.testOnSave` is set to `true`
For now, I dont think we need this messaging at all.

Lets wait for some user feedback.
I also have added the telemetry to get the time taken for the `gopkgs`

We can re-visit this next month by which time I will have the data.

uudashr:
"Load Packages" message on bottom bar removed

ramya-rao-a:
great, thanks @uudashr :)

PR #1281, opened 2017-10-14
Add snippet for Example functions

slomek:
This allows VSCode users to create testable Example functions using a short snippet. It's represented by `ef` since we already have `tf` for tests and `bf` for benchmarks.

More on Example functions: https://blog.golang.org/examples.


ramya-rao-a:
Thanks @slomek !

Issue #789, opened 2017-02-12
Package docs from godoc in VS Code


elithrar:
Being able to open the current symbol in godoc—within VS Code—can be a huge productivity boost, as you don't need to leave VS Code and can browse related package documentation quickly.

This is something vim-go does very well: https://github.com/fatih/vim-go#example-mappings

> Open the relevant Godoc for the word under the cursor with `<leader>gd` or open it vertically with `<leader>gv`
> `au FileType go nmap <leader>gd <Plug>(go-doc)`
> `au FileType go nmap <leader>gv <Plug>(go-doc-vertical)`

Being able to do this, and configure keybindings for the behaviour, would be very useful.

ramya-rao-a:
That would be cool to do, PRs are welcome :)

ramya-rao-a:
Related ask in https://github.com/Microsoft/vscode-go/issues/1237from @apiarian:

It would be great to have access to the output of `godoc` for a package (and possibly a specific identifier) within the editor without having to open up a browser to godoc.org or similar.

Should be possible using https://code.visualstudio.com/docs/extensionAPI/vscode-api#TextDocumentContentProvider . Some care will need to be taken to handle vendored code correctly. `godoc` seems to default to looking in the `$GOPATH` or `$GOROOT` so we'll probably have to explicitly check `$PROJECTROOT/vendor` first.

I'd like to take this issue on myself, but it'll be my first foray into vscode extensions out of my backend-development cave, so if somebody is keen to see this quicker, please let me know. I'd certainly be willing to hand the branch over or collaborate on the work.

navossoc:
I can't see the example on the first post, the link seems broken.

I was thinking in creating something like this:
https://goplay.space/

> Smart help lookup: double-click on e.g. package keyword or Println function name in source code, and you will see the relevant help topic. Try it!

It doesn't seems that hard to implement it on VS Code as a plugin, but I don't have any knowledge on how VS extensions work.

Some components that may help:
- [vscode-uri](https://www.npmjs.com/package/vscode-uri) - The URI implementation used by VS Code and its extensions.
- [request-light](https://www.npmjs.com/package/request-light) - A light weight Node.js request library with proxy support

Some sample extensions that may help too:
- https://github.com/Microsoft/vscode-extension-samples/blob/master/contentprovider-sample/README.md
- https://github.com/Microsoft/vscode-extension-samples/blob/master/previewhtml-sample/README.md

apiarian:
[...]

PR #1282, opened 2017-10-15
Go documentation window

apiarian:
Open a window with package documentation for #789

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,6 +233,11 @@
         "command": "go.install.package",
         "title": "Go: Install Current Package",
         "description": "Install the current package."
+      },
+      {
+        "command": "go.doc",
+        "title": "Go: Doc",
+        "description": "Get go documentation."
       }
     ],
     "debuggers": [
--- a/package.json
+++ b/package.json
@@ -1007,6 +1007,13 @@
           "group": "Go group 2"
         }
       ]
-    }
+    },
+    "keybindings": [
+      {
+        "command": "go.get.doc",
+        "key": "ctrl+alt+d",
+        "when": "editorTextFocus && resourceLangId == go"
+      }
+    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -926,6 +926,11 @@
           "default": [],
           "description": "Folder names (not paths) to ignore while using Go to Symbol in Workspace feature",
           "scope": "resource"
+        },
+        "go.docsHtml": {
+          "type": "boolean",
+          "default": false,
+          "description": "Show rendered HTML godoc instead of plain text"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -888,6 +888,11 @@
               "type": "boolean",
               "default": true,
               "description": "If true, adds command to upload the current file or selection to the Go Playground"
+            },
+            "goDoc": {
+              "type": "boolean",
+              "default": true,
+              "description": "If true, adds command to get go documentation to the editor context menu"
             }
           },
           "default": {
@@ -902,7 +907,8 @@
             "generateTestForPackage": false,
             "addImport": true,
             "testCoverage": true,
-            "playground": true
+            "playground": true,
+            "goDoc": true
           },
           "description": "Experimental Feature: Enable/Disable entries from the context menu in the editor.",
           "scope": "resource"
@@ -983,6 +989,11 @@
         {
           "when": "editorTextFocus && config.go.editorContextMenuCommands.testCoverage && resourceLangId == go",
           "command": "go.test.coverage",
+          "group": "Go group 1"
+        },
+        {
+          "when": "editorTextFocus && config.go.editorContextMenuCommands.goDoc && resourceLangId == go",
+          "command": "go.doc",
           "group": "Go group 1"
         },
         {
--- a/package.json
+++ b/package.json
@@ -238,6 +238,11 @@
         "command": "go.doc",
         "title": "Go: Doc",
         "description": "Get go documentation."
+      },
+      {
+        "command": "go.get.doc",
+        "title": "Go: Get Documentation",
+        "description": "Get docuemtnation for package using godoc."
       }
     ],
     "debuggers": [
@@ -993,7 +998,7 @@
         },
         {
           "when": "editorTextFocus && config.go.editorContextMenuCommands.goDoc && resourceLangId == go",
-          "command": "go.doc",
+          "command": "go.get.doc",
           "group": "Go group 1"
         },
         {
--- a/src/goDocumentation.ts
+++ b/src/goDocumentation.ts
@@ -5,14 +5,13 @@ import cp = require('child_process');
 import path = require('path');
 import { CancellationToken, TextDocumentContentProvider, Uri } from 'vscode';
 import { getBinPath, getToolsEnvVars, byteOffsetAt, getFileArchive } from './util';
-
-const missingToolMsg = 'Missing tool: ';
+import { promptForMissingTool } from './goInstallTools';
 
 export function getDocumentation(): void {
 	let gogetdoc = getBinPath('gogetdoc');
 	if (!path.isAbsolute(gogetdoc)) {
-		vscode.window.showErrorMessage(missingToolMsg + 'gogetdoc');
-		return;
+		promptForMissingTool('gogetdoc');
+		_getDocumentationForImport();
 	}
 
 	let ate = vscode.window.activeTextEditor;
@@ -28,22 +27,22 @@ export function getDocumentation(): void {
 
 				_getDocumentationForImport(imprt);
 			} catch (e) {
-				_getDocumentationForImport('');
+				_getDocumentationForImport();
 			}
 		});
 		p.stdin.end(getFileArchive(document));
 	} else {
-		_getDocumentationForImport('');
+		_getDocumentationForImport();
 	}
 }
 
-function _getDocumentationForImport(imprt: string): void {
+function _getDocumentationForImport(imprt?: string): void {
 	vscode.window.showInputBox({
 		prompt: 'Please enter a package name',
 		value: imprt,
-		placeHolder: imprt === '' ? 'no package detected' : '',
+		placeHolder: imprt ? 'no package detected' : '',
 	}).then(pkgInput => {
-		if (typeof pkgInput === 'undefined' || pkgInput === '') {
+		if (!pkgInput) {
 			return;
 		}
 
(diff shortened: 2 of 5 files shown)


apiarian:
This is my first PR for vscode-go and my first foray into typescript ever and javascript in many years. I'd very much appreciate any comments or notes on this first draft. The code is minimally functional, though there's some things I'd like to add, like context menu selection.

apiarian:
OK, so this is now pretty much feature complete as far as I can tell. I might add a bit of godoc output massaging to make the import bit look nicer for vendored packages, but you can get docs for a package that you can name, and it will try to detect some package under the cursor courtesy of `gogetdoc`.

Hopefully there aren't any terrible or subtle bugs in here =/ 

What's the code review / merging process here?

apiarian:
Did I somehow break the `listPackages` test? Or is it just flaky?

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+import path = require('path');
+import { CancellationToken, TextDocumentContentProvider, Uri } from 'vscode';
+import { getBinPath, getToolsEnvVars, byteOffsetAt, getFileArchive } from './util';
+
+const missingToolMsg = 'Missing tool: ';
+
+export function getDocumentation(): void {
+	let gogetdoc = getBinPath('gogetdoc');
+	if (!path.isAbsolute(gogetdoc)) {
+		vscode.window.showErrorMessage(missingToolMsg + 'gogetdoc');
```
ramya-rao-a: I am guessing this was picked up from `goDeclaration.ts` file.
When a tool is missing, we usually call the `promptForMissingTool` method that will convey the message and prompt the user to install the tool. 
We will need to do the same here instead of just showing the error message

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+				_getDocumentationForImport('');
+			}
+		});
+		p.stdin.end(getFileArchive(document));
+	} else {
+		_getDocumentationForImport('');
+	}
+}
+
+function _getDocumentationForImport(imprt: string): void {
```
ramya-rao-a: you can have `imprt` as an optional parameter. That way you call `_getDocumentationForImport()` instead of `_getDocumentationForImport('')` and later replace the check `imprt === '' ? 'no package detected' : ''` with `imprt ? 'no package detected' : ''`

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+	vscode.window.showInputBox({
```
ramya-rao-a: why show the inputbox at all if an import path was provided?

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+		prompt: 'Please enter a package name',
+		value: imprt,
+		placeHolder: imprt === '' ? 'no package detected' : '',
+	}).then(pkgInput => {
+		if (typeof pkgInput === 'undefined' || pkgInput === '') {
```
ramya-rao-a: you can replace this with a single check `if (!pkgInput)`

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+			return;
+		}
+
+		let uri = vscode.Uri.parse('godocumentation://');
+		uri = uri.with({
+			path: pkgInput,
+		});
+
+		vscode.workspace.openTextDocument(
```
ramya-rao-a: you can directly call `showTextDocument` here

`vscode.window.showTextDocument(uri, { viewColumn: vscode.window.activeTextEditor.viewColumn + 1, preserveFocus: true });`

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export class GoDocumentationContentProvider implements TextDocumentContentProvider {
+	public provideTextDocumentContent(uri: Uri, token: CancellationToken): Thenable<string> {
+		let godoc = getBinPath('godoc');
+
+		let pkg = uri.path;
+
+		// godoc doesn't handle vendor directories correctly. So we first try
+		// the local workspace vendor directory to see if there's a vendored
+		// version of the package. If not, we'll return whatever godoc would
```
ramya-rao-a: if we didnt strip out "vendor" from the gogetdoc output, then we can skip this part, correct?

src/goDocumentation.ts
```diff
@@ -0,0 +1,114 @@
...
+			});
+		}).then(vendorOutput => {
+			if (vendorOutput !== '') {
+				return vendorOutput;
+			}
+
+			return new Promise<string>((resolve, reject) => {
+				cp.execFile(godoc, [pkg], {env: getToolsEnvVars()}, (err, stdout, stderr) => {
+					let output = stdout + stderr;
+					let match = /cannot find package/.exec(output);
```
ramya-rao-a: why not look for "cannot find package" only in stderr?

package.json
```diff
@@ -174,6 +174,11 @@
         "command": "go.get.package",
         "title": "Go: Get Package",
         "description": "Run `go get -v` on the package on the current line."
+      },
+      {
+        "command": "go.doc",
+        "title": "Go: Doc",
+        "description": "Get go documentation."
       }
```
ramya-rao-a: How about `go.get.doc` , `Go: Get Documentation` and `Get documentation using godoc for package`  instead?

src/goDocumentation.ts
apiarian: This way if the cursor happens to be on some `gogetdoc`-able thing accidentally, the user has the opportunity to enter a package to view. There's one command to get docs. I suppose it might be possible to split it so that a context menu click or a keyboard shortcut invocation would trigger the "try to find the thing my cursor is on and only show me the input if you can't find anything" flow, while a more general `command-shift-p` invocation  would always show the user the box. The current "one command to rule them all" way seemed less complicated. Is there a way to detect how the command is being invoked (context menu, vs. shortcut, vs. command palette) ?

src/goDocumentation.ts
apiarian: True, but it seems that wiring that into this part of the code would be somewhat non-trivial. I suppose we could add a query parameter to the URI, something like `?vendored=1` and then look for it here. But then what do you do for packages that were entered manually instead of using the `gogetdoc` flow? The `godoc` invocation is fairly cheap, so why not just try it?

apiarian:
@ramya-rao-a I think I've addressed most of your comments. Not quite sure how to proceed with the invocation-source-detection stuff, though.

src/goDocumentation.ts
ramya-rao-a: The callback for `vscode.commands.registerCommand` takes in an array of arguments.
This would be null/undefined when the command is run from the command pallet
If the command was run from context menu, then this would have the file details corresponding to the file in the editor of the context menu.

You can also use a `quickPick` control and pre-fill it with all valid packages when no valid pkg can be inferred from gogetdoc. See https://github.com/Microsoft/vscode-go/blob/0.6.66/src/goBrowsePackage.ts#L75

src/goDocumentation.ts
ramya-rao-a: Weird.. I just tried the Go Doc command on a symbol from a vendored package.
The import path returned by gogetdoc did not have "vendor" in it.
So the attempt to strip out "vendor" became a moot point.

apiarian: That may mean that your `gogetdoc` is not detecting the right version of the package? Could you try with this set of repos https://github.com/apiarian/vscode-go-test-docs-main ? That's what I've been  using for testing. Maybe it's too small a toy?

apiarian: So here's what I see for that repo and the latest `gogetdoc` installed:

```
~/.vscode/gopath/bin/gogetdoc -json -pos main.go:#421 | jq
{
  "name": "vendoroverride",
  "import": "github.com/apiarian/vscode-go-test-docs-main/vendor/github.com/apiarian/vscode-go-test-docs-vendoroverride",
  "pkg": "vendoroverride",
  "decl": "package vendoroverride",
  "doc": "Package vendoroverride in the gopath defines foo and bar with some docs.\nThese have been overridden in the vendor directory of the main package. We\ncould probably do the same with branches or something, but this seems a bit\nmore controllable.\n",
  "pos": ""
}
```

src/goDocumentation.ts
apiarian: added this functionality too

src/goDocumentation.ts
```diff
@@ -152,12 +170,49 @@ function _packageNotFound(stderr: string): boolean {
...
+	let staticPath = godocHtmlStaticPath();
+
+	output = output.replace(/document\.ANALYSIS_DATA = ;/, '');
+	output = output.replace(/document\.CALLGRAPH = ;/, '');
+
+	output = `
+		<!DOCTYPE html>
+		<html>
+		<head>
+			<!--temporarily commented until we figure out what the licence situation is
```
apiarian: @ramya-rao-a so this bit is pending the answer to https://github.com/Microsoft/vscode-go/issues/789#issuecomment-338488571

apiarian: Otherwise this will take some CSS + JS wizardry to make the output pretty.

apiarian:
@ramya-rao-a while fixing a bug with the multi-root support (we're always looking in the zeroth root vendor directory, not the current root vendor directory) I came across another bug: the documentation view is sticky. If we ask for docs on package `foo` and then change the on-disk documentation for `foo`, asking for `foo` docs again shows the old docs. If we ask for `bar` docs before asking for `foo` docs again, the `foo` docs do get updated. "Change the on-disk documentation" may mean changing the documentation strings or it may mean moving to a different workspace root with a different vendor directory (or no vendor directory at all).

I suppose we could add some sort of `time.Now()` based cache-busting bit in the URI that we request for the `provideTextDocumentContent` method, but is there a more "correct" way to do it? Some way to signal to vscode that we don't want it to show cached data ever?

src/goDocumentation.ts
```diff
@@ -97,11 +97,12 @@ function _devendorPkgs(pkgs: string[]): string[] {
 function _showDocumentationForPackage(pkg: string): void {
 	let uri = vscode.Uri.parse('godocumentation://').with({
 		path: pkg,
+		query: 'now=' + new Date().getTime().toString(),
```
apiarian: @ramya-rao-a is there a better way to bust vscode's text document content provider cache?

ramya-rao-a:
Closing as per https://github.com/Microsoft/vscode-go/issues/789#issuecomment-409316549

PR #1297, opened 2017-10-21
Add `go.toolCommands` which specifies alternate command locations

ikedam:
[Appengine Go](https://cloud.google.com/appengine/docs/standard/go/) SDK provides `goapp` command which wraps `go` command.

I need to create a symbolic link `go` to `goapp` to develop Appengine Go applications with vscode-go.
But I don't want to do that as it masks the original `go` command. I want to use both `goapp` in the Appengine Go SDK and `go` of of the Appengine Go SDK.

This request enables to configure alternate command for `go` like:
```
    "go.toolCommands": {
        "go": "goapp.exe"
    },
```

You can also provide absolute path:
```
    "go.toolCommands": {
        "go": "/home/ikedam/go_wrap.sh"
    },
```

You can also configure alternate command for other tool:
```
    "go.toolCommands": {
        "godoc": "C:\\Users\\ikedam\\godoc_with_options.bat"
    },
```


src/goPath.ts
```diff
@@ -11,16 +11,21 @@
...
 import path = require('path');
 import os = require('os');
+import vscode = require('vscode');
 
 let binPathCache: { [bin: string]: string; } = {};
 let runtimePathCache: string = '';
 
 export function getBinPathFromEnvVar(toolName: string, envVarValue: string, appendBinToPath: boolean): string {
-	toolName = correctBinname(toolName);
+	let binname = correctBinname(toolName);
```
ramya-rao-a: `correctBinName` gets run twice resulting in strings like `golint.exe.exe` in Windows

src/goPath.ts
```diff
@@ -11,16 +11,21 @@
 import fs = require('fs');
 import path = require('path');
 import os = require('os');
+import vscode = require('vscode');
```
ramya-rao-a: This file is used by the debug adapter via the `goDebug.ts` file. The debug adapter runs in a different process than the extension host and so will not have access to `vscode`

ramya-rao-a:
To test this feature, do the following:

- Download https://github.com/ikedam/vscode-go/blob/feature/toolCommands/Go-latest.vsix
- Reload VS Code

jlindsey:
Downloaded and installed this from the .vsix file.

It seems to be doing what it says on the tin: When I add the following to my settings.json:
```
{
    "go.toolCommands": {
        "go": "wgo"
    }
}
```
Then looking in the Output tab I see the tool being replaced properly in my project:
```
/Users/jl2243/projects/terraform/tools/nagios/rpc/src/tf_nagios_api>Finished running tool: /Users/jl2243/projects/go/bin/wgo build -i -o /var/folders/3_/8n0lrvq17qb3rkdwd4dfgr_80000gp/T/go-code-check tf_nagios_api
```

However: when I intentionally break my code by adding an unused import, I do not get any error report. In the terminal, running this manually I get:
```
/Users/jl2243/projects/go/bin/wgo build -i -o /var/folders/3_/8n0lrvq17qb3rkdwd4dfgr_80000gp/T/go-code-check tf_nagios_api
# tf_nagios_api
src/tf_nagios_api/service.go:6:2: imported and not used: "encoding/json"
```
But no output in VSCode or the Problems tab. I don't know if this is a separate issue or not.

ikedam:
@ramya-rao-a
I reproduced the issue that you pointed in my environment. Sorry.
And your fix worked perfect, thanks!

ikedam:
@jlindsey Would you tell me what is `wgo`?

In my environment, error reports are shown as expected by using `goapp` instead of `go`.

jlindsey:
@ikedam @ramya-rao-a After doing some more tests on my end, I believe my results were caused by old configs in my global User settings. After removing all the `go.*`settings from there and running more tests on a few codebases, everything behaves as expected.

LGTM now. 👍

ramya-rao-a:
Thanks for verifying @ikedam and @jlindsey 

On second thoughts, I am wondering if there are really any benefits of having this customization feature for other tools (not `go`, but other tools like `godoc`, `gorename`, `goformat` etc)

This feature provides 2 benefits. One is the ability to use a totally different tool, the other is to use the same tool from a different location. 

Each of these tools take in very specific input parameters. What are the chances of having another tool that takes in exact same parameters? The benefit of different location is available even today. If you have the tool under the PATH variable, then it will get picked up.

I propose to have this customization feature just for `go`.

Thoughts?

mbenkmann:
Correct me if I'm wrong, but right now the feature seems to be completely generic. It's a mapping that just works for all tools. Would limiting it to "go" not require adding extra checks just to take away features? 

As for the usefulness for other tools, one issue I've run into several times already is that to figure out why something wasn't working (most recently why gogetdoc wasn't finding the docs for a symbol) I introduced a wrapper script that would log the parameters the tool was called with and its output. This feature would make this much easier and cleaner. Adding a script with the same name as a standard tool to a directory in the PATH is a clutch. With this feature I could do such testing in workspace settings so it wouldn't interfere with anything. 

And then there is the case of my goformat tool. Before it was added to the valid list of values for go.formatTool, I had to rename it to gofmt to avoid the error. Again, this mapping feature would have provided a cleaner solution.

Right now I'm toying with the idea of developing an alternative to gogetdoc that works without compiling, so it would be faster. This mapping feature would make testing this tool easier because I could override gogetdoc in the workspace settings as opposed to fiddling with my PATH.

To put it simply: There are a lot of fringe uses for a generic mapping feature like this. If it doesn't make the code worse, I would keep the ability to remap all tools.

ikedam:
@ramya-rao-a 
+1 and -1 for having this feature only for `go`:

+1 as
* I'm happy even with switching only `go` for my current usecase.
    * I don't have actual usecases for switching other tools.
* `vscode-go` handles the path for `go` and those for other tools in different ways (`getGoRuntimePath()` and `getBinPathWithPreferredGopath()`), and it can easily cause bugs to switching paths in the same way, just like I did. It makes sense to apply this feature only to `go` for now, and then introduce a feature for other tools in another day.

-1 as
* I believe the feature for other tools is still useful as @mbenkmann pointed:
    * Useful for per-workspace settings.
    * Wrapping a tool to change the tool behavior (e.g. add a option switch when launching).

So how about going in this way?:

1. Close this request.
2. Create a new request to generalize and unite `getGoRuntimePath()` and `getBinPathWithPreferredGopath()`.
    * I plan `getBinPath(toolName, searchPaths, defaultPaths)`.
3. Create a new request remaking this request.

I think I can work for that in June.

ikedam:
I worry that we lose the chance to generalize `getGoRuntimePath()` and `getBinPathWithPreferredGopath()` forever if we apply this feature only to `go`.

ramya-rao-a:
I'll have to go back in history to see why the two were implemented differently to begin with :) So that we dont regress anything

ramya-rao-a:
This is merged in master.  Please give it a try by following the below steps and share any feedback you have.

- Reload VS Code
- Use the setting `go.alternateTools` for your needs of mapping the tools as needed

jlindsey:
@ramya-rao-a Downloaded and installed successfully. Switched my config above for the new setting name. Everything still works fine and as expected.

Thanks for your hard work on this!

ramya-rao-a:
This feature is now available in the latest update to the Go extension (0.6.83)

ikedam:
@ramya-rao-a I tried 0.6.83-beta-1 (and 0.6.83, sorry for my late response).
It worked perfectly. I can use `goapp` in my project.
Thanks for all the work!

ramya-rao-a:
Thanks to you too @ikedam!

Issue #972, opened 2017-05-06
Ability to run Benchmarks from test file


scriptnull:
First of all, vscode-go is awesome. It is best workflow experience for writing go code.

One of the features, I love the most is ability to run a test with a single click on the "run test" link that appears above the test function

![screen shot 2017-05-06 at 11 30 25 pm](https://cloud.githubusercontent.com/assets/4211715/25774786/173cb428-32b4-11e7-8689-b5547bbe75cb.png)

I would like to see the support for running benchmarks just by clicking a link above the benchmark function

![screen shot 2017-05-06 at 11 32 28 pm](https://cloud.githubusercontent.com/assets/4211715/25774806/66a8bab6-32b4-11e7-8640-e4790f268a88.png)

Currently, the benchmark functions does not have anything above them. I would like to see the support for "run benchmark" link above the benchmark functions.

ramya-rao-a:
Currently only the Test* and Example* functions are being identified as "test functions" that can be run.

See https://github.com/Microsoft/vscode-go/blob/0.6.61/src/goTest.ts#L267

Looping in @ironcladlou who wrote most of the initial test related code in the Go extension.

@ironcladlou Looking at history, it seems initially we only identified "Test*" functions. And then somewhere along the line, we added "Example*". Do you know of any reason why "Benchmark*" was excluded?

ironcladlou:
@ramya-rao-a 
>Looking at history, it seems initially we only identified "Test*" functions. And then somewhere along the line, we added "Example*". Do you know of any reason why "Benchmark*" was excluded?

At the time, I just didn't think about them because I didn't use benchmarks. So, no special reason. 😁  I absolutely think the plugin should support them!

Thinking about them now:

1. Benchmark functions do seem distinct from tests in practice and I'd expect benchmarks to not get executed automatically when running "test". I'd be interested to discuss this if there's any disagreement.
2. Although I don't have a ton of experience with Go's benchmarking support, I suspect the benchmarks often rely on user input (e.g the `-benchtime`, and `-count` flags). I'd be interested to hear anecdotes on this point.
3. I wonder how fancy one could get with the new [progress UI support](https://code.visualstudio.com/updates/v1_12#_extension-authoring)...

If point 2 turns out to be true, is there any good way to collect user input for the benchmark executions within the VScode UI? I clearly haven't had much time to keep up with plugin or VScode development lately, so I'm not sure what capabilities exist.

[...]

PR #1303, opened 2017-10-25
run benchmark at cursor (#972)

buyology:
this is a stab at #972 along the lines of the discussion to have `run benchmark` available at the cursor level

#### note:
- [x] runs with the `-benchmem` flag by default
- [x] use `-run=^$` to exclude all tests from being run

#### look and feel

run-benchmark

(nb: this is my first foray into `typescript`, and my first locs in javascript in many years so please bear with me)

cc @scriptnull @ironcladlou @campoy @rakyll


src/testUtils.ts
```diff
@@ -196,11 +233,18 @@ function expandFilePathInOutput(output: string, cwd: string): string {
...
+
+/**
  * Get the test target arguments.
  *
  * @param testconfig Configuration for the Go extension.
  */
-function targetArgs(testconfig: TestConfig): Thenable<Array<string>> {
+function testTargetArgs(testconfig: TestConfig): Thenable<Array<string>> {
 	if (testconfig.functions) {
 		return Promise.resolve(['-run', util.format('^%s$', testconfig.functions.join('|'))]);
```
ramya-rao-a: Instead of having a separate `benchmarkTargetArgs`, you can just use `-bench` here instead of `-run` if `testconfig.isBenchmark` is `true`

src/testUtils.ts
```diff
@@ -78,29 +82,49 @@ export function getTestFlags(goConfig: vscode.WorkspaceConfiguration, args: any)
...
+
+/**
  * Returns whether a given function name has a test prefix.
  * Test functions have "Test" or "Example" as a prefix.
  *
  * @param the function name.
  * @return whether the name has a test function prefix.
  */
-function hasTestFunctionPrefix(name: string): boolean {
+export function hasTestFunctionPrefix(name: string): boolean {
```
ramya-rao-a: There was no real point of making this a separate function in the first place (my apologies).

Instead of exporting this prefix function and adding another for benchmark, we can simplify things as below
- remove `hasTestFunctionPrefix` and move the prefix check into `getTestFunctions`
- create and export a new function `getBenchmarkFunctions` which will have the same code as `getTestFunctions` but with a different prefix check.

It will definitely get simpler when importing these functions in other files

src/goRunTestCodelens.ts
```diff
@@ -95,7 +95,21 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
...
+					command: 'go.benchmark.cursor',
+					arguments: [ { functionName: func.name, isBenchmark: true } ]
+				};
+
+				codelens.push(new CodeLens(func.location.range, runBenchmarkCmd));
+			});
+
+		});
+
+		return Promise.resolve(codelens);
```
ramya-rao-a: This Promise should resolve only after both tests and benchmark functions are found

ramya-rao-a:
@buyology I went ahead and made the refactorings myself, hope you dont mind.
Great work!

buyology:
@ramya-rao-a did not mind :) thanks!

PR #1309, opened 2017-10-31
better pkg completion

wangkechun:


ramya-rao-a:
Am assuming this is to ensure standard pkgs appear before the rest in the suggestions?

Issue #1265, opened 2017-10-05
Close processes before spawning new ones


ramya-rao-a:
Features that depend on Go tools spawn a new process for every new request made.
When the tools get slow (for any which reason) this ends up generating multiple processes for the same tool.

This is more prominent for tools that back features like hover and auto-completion.

This issue is to track the possibility of closing older processes for these features when new ones are spawned.

ramya-rao-a:
This has been done. The only remaining feature that does not do this yet is the auto-completions which is tracked in #1272

PR #1326, opened 2017-11-09
Cancel currently running processes when spawning new ones

ramya-rao-a:
Related to #1265

The `DefinitionProvider`, `SignatureProvider` and `HoverProvider` all use the same tools behind the scenes. 
These tools are configured using the `go.docsTool` setting and can be `gogetdoc`, `godef` + `godoc` or `guru`

This PR includes the below features
- When cancellation is requested by VSCode via the `cancellationToken`, kill the currently running process
- Before creating a new process, kill the existing one. But these should not criss-cross among the 3 features, eg: A request for hover should not kill the process that was created for signature help request
- Major refactorings to accomodate the above

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -24,216 +24,230 @@ export interface GoDefinitionInformation {
 	toolUsed: string;
 }
 
-export function definitionLocation(document: vscode.TextDocument, position: vscode.Position, goConfig: vscode.WorkspaceConfiguration, includeDocs = true): Promise<GoDefinitionInformation> {
-	let wordRange = document.getWordRangeAtPosition(position);
-	let lineText = document.lineAt(position.line).text;
-	let word = wordRange ? document.getText(wordRange) : '';
-	if (!wordRange || lineText.startsWith('//') || isPositionInString(document, position) || word.match(/^\d+.?\d+$/) || goKeywords.indexOf(word) > 0) {
-		return Promise.resolve(null);
-	}
-	if (!goConfig) {
-		goConfig = vscode.workspace.getConfiguration('go', document.uri);
-	}
-	let toolForDocs = goConfig['docsTool'] || 'godoc';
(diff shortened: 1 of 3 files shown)


src/goDeclaration.ts
```diff
@@ -24,216 +24,230 @@ export interface GoDefinitionInformation {
...
+						}
+						return resolve(definitionInformation);
+					});
+					this.currentRunningProcess.on('close', () => { this.currentRunningProcess = null; });
 				});
-			} catch (e) {
-				reject(e);
-			}
+				this.currentRunningProcess.stdin.end(tool === 'godef' ? document.getText() : getFileArchive(document));
+				this.currentRunningProcess.on('close', () => { this.currentRunningProcess = null; });
```
roblourens: I think the nulling out should be done in `cancelCurrentRunningProcess()`. I think there could be a race condition here. If this happens

- cancelCurrentRunningProcess() kills the process
- 'close' event is queued but not handled yet
- currentRunningProcess is immediately assigned to a newly spawned process
- 'close' handler runs and nulls out the prop, but the process is actually still running.

ramya-rao-a: Done

ramya-rao-a: Done

ramya-rao-a:
@roblourens On second thoughts all this feels like an overkill.

If we assume that the cancellationToken does what its suppose to do, then we dont need to try and cancel previous prcoesses before spawning new ones.
We only need to cancel them when the token fires the cancellation event.

In that case, something like this is more than enough: https://github.com/Microsoft/vscode-go/commit/b5bbff719119ee1cd6ed86c750ff97499537ab80

What do you think?

roblourens:
Yeah that makes sense

ramya-rao-a:
Done with https://github.com/Microsoft/vscode-go/commit/63f904604b2c40ad3df53b1ef7e616feb9e1039b

PR #1354, opened 2017-11-19
Delay connecting to dlv until it's listening

### LINKED ISSUE #473: debugging hangs with no error message
Could be related to #240 but since a long time has passed since then, could also be a new issue.

When I press the "play" button a progress indication appears but nothing happens.

I'm working with:
- Visual Studio Code Version 1.5.1 (1.5.1)
- Go extension 0.6.43
- go version go1.6.2 darwin/amd64
- Delve Debugger Version: 0.11.0-alpha Build: v0.11.0

I can confirm running `dlv debug`, then `continue` on my console works.

I can not say it's a regression, since this is the first time I try it in this machine.

If I type anything in the debug console, an error message appears:

```
TypeError: Cannot read property 'currentGoroutine' of null\n    at GoDebugSession.evaluateRequest (/Users/gimix/.vscode/extensions/lukehoban.Go-0.6.43/out/src/debugAdapter/goDebug.js:595:45)\n    at GoDebugSession.DebugSession.dispatchRequest (/Users/gimix/.vscode/extensions/lukehoban.Go-0.6.43/node_modules/vscode-debugadapter/lib/debugSession.js:354:22)\n    at GoDebugSession.ProtocolServer._handleData (/Users/gimix/.vscode/extensions/lukehoban.Go-0.6.43/node_modules/vscode-debugadapter/lib/protocol.js:78:38)\n    at Socket.<anonymous> (/Users/gimix/.vscode/extensions/lukehoban.Go-0.6.43/node_modules/vscode-debugadapter/lib/protocol.js:23:60)\n    at emitOne (events.js:90:13)\n    at Socket.emit (events.js:182:7)\n    at readableAddChunk (_stream_readable.js:153:18)\n    at Socket.Readable.push (_stream_readable.js:111:10)\n    at Pipe.onread (net.js:531:20)
```

DMarby:
#473

In a project with external dependencies, debugging hang indefinitely for me as `dlv` produces output while compiling the go code, before it starts the api server.
This caused VS Code to attempt to connect to `dlv` the api server was listening.

In order to avoid this, and to avoid arbitrary delays in the code, this change makes it so that we wait for `dlv` to declare that it's ready to accept connections before attempting to connect.

I've removed the calls to connectClient from stderr completely, as there is to my knowledge only two possible situations, both which are covered:
- Dlv starts up fine, and emits the "Api server listening" message
- Dlv crashes/exits, which gets caught by the `error` and `close` listeners.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -327,10 +327,6 @@ class Delve {
 			this.debugProcess.stderr.on('data', chunk => {
 				let str = chunk.toString();
 				if (this.onstderr) { this.onstderr(str); }
-				if (!serverRunning) {
-					serverRunning = true;
-					connectClient(port, host);
-				}
 			});
 			this.debugProcess.stdout.on('data', chunk => {
 				let str = chunk.toString();


DMarby:
Unsure as to why the Travis build is failing for one version on Mac on a completely unrelated test, perhaps someone with access could trigger a restart of that job?

ramya-rao-a:
> This caused VS Code to attempt to connect to dlv the api server was listening.

Do you mean VS Code was trying to connect to delve **before** the api server was listening?

src/debugAdapter/goDebug.ts
```diff
@@ -313,29 +313,21 @@ class Delve {
...
-				if (!serverRunning) {
-					serverRunning = true;
-					connectClient(port, host);
-				}
 			});
 			this.debugProcess.stdout.on('data', chunk => {
 				let str = chunk.toString();
 				if (this.onstdout) { this.onstdout(str); }
-				if (!serverRunning) {
+				if (!serverRunning && str.startsWith('API server listening at: ')) {
```
ramya-rao-a: @derekparker Is there any other recommended way for editors to figure out when to connect the client to `dlv` other than waiting for the message `API server listening at:` on the stdout?

DMarby:
> Do you mean VS Code was trying to connect to delve before the api server was listening?

Yes, that is what was happening.

ramya-rao-a:
@DMarby I am a little uncomfortable with checking for the string "API server listening at:". 

The assumption here is that when delve returns with `API server listening at:`, it is ready for the client to make the connection.

Does delve ever return with any other kind of data before returning `API server listening at:`?
If not, then the existing code should also work, correct?
If yes, there are 2 things that could go wrong 
- The message itself may change in the future versions of delve
- The data received in `this.debugProcess.stdout.on('data', chunk => {})` is not guaranteed to be the complete string you expect. It could be just `API server` and `listening at` may come in the next chunk.

Also, @lukehoban do you remember why the delay was added before making the connection to delve?
See https://github.com/Microsoft/vscode-go/commit/69f781534b766f871a6246eaa61c879176696b56
Would waiting for  "API server listening at:"  solve the issue you were trying to solve?

src/debugAdapter/goDebug.ts
derekparker: The editor should know the host / port it is trying to connect to, would it be possible to simply retry (potentially w/ backoff, or an upper-bound retry time) the connection until the server is running and accepting connections?

DMarby:
That makes sense! As far as I can tell, nothing is outputted on stdout before the API server listening, so I've removed the string comparison.

A better long-term solution would most likely be to attempt connection multiple times like @derekparker suggested, but I think this is a good change until that can be done.

src/debugAdapter/goDebug.ts
ramya-rao-a: Yes, that's possible and has come up as one of the solutions when I was discussing with @roblourens 
Thanks @derekparker

ramya-rao-a:
@DMarby Below is the net change in this PR
- Remove attempting to make a connection when dlv sends something on the stderr 
- Remove the 200ms wait right after the first data chunk is received in stdout before making the connection

I agree with the first point.
But how is removing the 200ms wait going to help #473?

DMarby:
The issue in #473 occurs when we attempt to connect to dlv before it's ready.
As the changes in this PR ensures that dlv is ready before we attempt to connect, the 200ms delay shouldn't be needed anymore, as far as I can tell.

ramya-rao-a:
> when there's custom build flags with verbose causing stdout output during build,

In that case, our assumption that " nothing is outputted on stdout before the API server listening," is wrong and we will have to add the check on "API server listening" back, correct?

DMarby:
After looking into it further (and testing by adding "-v" as buildFlags to my debug configuration), that gets outputted on stderr rather than stdout, so we shouldn't need the api server listening check.

ramya-rao-a:
In that case, I would keep the delay as well until we hear back from @lukehoban about the reason the delay was put in the first place.
The removal of the connection attempt during stderr should fix #473

DMarby:
Fair enough, I've gone ahead and reverted the change that removed the delay.

ramya-rao-a:
Awesome, thanks for working on this @DMarby !

PR #1365, opened 2017-11-24
Add remote debugging debug configuration

mat007:
It appears the template for generating a remote debugging configuration is missing.
Somewhat related to https://github.com/Microsoft/vscode-go/pull/254


ramya-rao-a:
Thanks @mat007!

PR #1368, opened 2017-11-26
Complete method receivers of types in namespace #168

grooveygr:


src/goSuggest.ts
```diff
@@ -183,6 +185,18 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 								item.insertText = new vscode.SnippetString(suggest.name + '(' + paramSnippets.join(', ') + ')');
 							}
 
+							if (wordAtPosition && wordAtPosition.start.character === 0 &&
+								suggest.class === 'type' && !goBuiltinTypes.has(suggest.name)) {
+								let prefix = 'func (' + suggest.name[0].toLowerCase() + ' *' + suggest.name + ')';
+								let auxItem = new vscode.CompletionItem(suggest.name + ' method', vscode.CompletionItemKind.Snippet);
+								auxItem.label = suggest.name + ' (new method)';
```
ramya-rao-a: What do you think of `'func (*' + suggest.name + ')` for `label` and `Method snippet` for `detail`?

src/goSuggest.ts
```diff
@@ -183,6 +185,18 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+								auxItem.detail = prefix + '...';
+								auxItem.sortText = 'a';
+								let snippet = prefix + ' ${1:name}(${2:params}) ${3:retval} \{\n\t$0\n\}';
```
ramya-rao-a: Since `params` and `retval` can be empty, why not have them as simple tabstops instead of placeholders?

i.e `${1:methodName}(${2}) ${3}` instead of `${1:name}(${2:params}) ${3:retvall}`

grooveygr: Done.

src/goSuggest.ts
grooveygr: I experimented with you idea and found the the vscode engine does not show suggestions in some case where the `suggest.name` is not a prefix of the `label`. More specifically,  when completing an unexported type the suggestion is not shown.
For now I changed the `label` to be exactly `suggest.name` and kept your suggestion for the `label`.
Let me know what you think!

ramya-rao-a: You can get around that by adding `auxItem.filterText = suggest.name`

By default the `label` is used for displaying the item as well as filtering the item based on what is typed. If `filterText` is set, then that will be used for filtering instead

grooveygr: Nice! Learned something new.
Done.

ramya-rao-a:
@grooveygr I had a question. Is it possible to have a method on a type from a different package?

grooveygr:
@ramya-rao-a according to the golang spec (https://golang.org/ref/spec#Method_declarations) adding methods to imported types is not supported.

> The type denoted by T is called the receiver base type; it must not be a pointer or interface type and it must be defined in the same package as the method.

When you try, you're met with a compilation error, e.g. `cannot define new methods on non-local type gzip.Header`

ramya-rao-a:
Great, thanks for the clarification!

PR #1369, opened 2017-11-26
Runnable golang scratch pad

grooveygr:
Provide a way to quickly prototype and test golang concepts.
Adds two new commands:

**Go: New Scratch Pad**
Creates or resets an existing scratch pad with a minimalistic "Hello World" main package.
The scratch file is kept in a temporary directory and its persistency should not to be relied upon.

**Go: Run Scratch Pad**
Compiles and runs the scratch file, outputting its STDOUT and STDERR to the "Go" output pane.
Upon termination, the process return value is also printed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,8 @@
   "activationEvents": [
     "onLanguage:go",
     "onCommand:go.gopath",
+    "onCommand:go.newscratch",
+    "onCommand:go.runscratch",
     "onCommand:go.tools.install",
     "onDebug"
   ],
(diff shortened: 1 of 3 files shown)


ramya-rao-a:
@grooveygr Please add a description of what you are trying to achieve in this PR.

grooveygr:
Edited original PR comment with a more detailed description.

ramya-rao-a:
I do believe this is a cool feature to have, although I dont believe this extension is the right place for it. 

As VS Code and this extension has gained popularity, we have seen a gradual scope creep over the past year. Being a sole maintainer of this project, it is getting a little hard to support all incoming requests and maintaining the ever growing set of features. Therefore,one criteria I am using now in the triaging process is whether the feature request would make sense as a separate extension. This current feature request appears to be so.

Thanks for all the work. Let me know if you need any help creating the new extension if you choose to do so.

PR #1370, opened 2017-11-26
Test package/file navigation button

grooveygr:


ramya-rao-a:
@grooveygr I am guessing you are trying to get the "Go: Test Package" command in the context menu?
That is do-able currently by adding the below setting
```
"go.editorContextMenuCommands": {
  "testPackage": true
}
```

grooveygr:
Close, but not quite! This adds a button to the editor title bar (see image).
I am personally less fond of a cluttered context menus and found this to be a nice compromise.
I experimented with various icons but found that just leaving the command text provided the most convenient UX for me.

ramya-rao-a:
What is the criteria for a command to be shown this way?
What's to stop anyone else from wanting to have other commands show up as a button this way?

I would suggest the below if context menu is not your choice
- Use the command from the command pallet
- Assign a keyboard shortcut for the command
- Use the codelens "run package tests" that appear above the `package` statement in a `_test.file`

If you would still like the button on the title bar, then I would suggest publishing a separate extension to do so.

Issue #1345, opened 2017-11-16
Delete binary files created by delve after closing the debug session


ramya-rao-a:
_From @firelizzard18 on June 20, 2017 7:56_

- VSCode Version: Code 1.13.1 (379d2efb5539b09112c793d3d9a413017d736f89, 2017-06-14T18:13:05.928Z)
- OS Version: Darwin x64 16.6.0
- Extensions: 

|Extension|Author|Version|
|---|---|---|
|aurelia|AureliaEffect|0.3.4|
|vscode-npm-script|eg2|0.2.0|
|task-master|ianhoney|0.1.39|
|quicktask|lkytal|2.8.0|
|Go|lukehoban|0.6.62|
|debugger-for-chrome|msjsdiag|3.1.4|;

---

Steps to Reproduce:

1. Open a Golang project
2. Debug something
3. Stop debugging (via `Stop`, aka the red square or Shift+F5)

---

`delve` is the debugger for golang, or at least it's the one that the Go extension uses. When you debug something with delve, it creates a large binary file in the current directory. If you debug a main function (`dlv debug`), you get 'debug'. If you debug a test function (`dlv test`), you get 'debug.test'.

In normal `delve` usage, when you're done, you `quit` delve. Delve then deletes this file. Apparently VSCode gracelessly terminates (SIGKILL?) delve, which means the file sticks around.

_Copied from original issue: Microsoft/vscode#29089_

ramya-rao-a:
_From @firelizzard18 on June 20, 2017 7:59_

This may be a regression from Microsoft/vscode-go#765 which fixed Microsoft/vscode-go#438

ramya-rao-a:
_From @weinand on June 20, 2017 10:3_

Yes, VS Code terminates debug adapters "gracelessly" (and we are planning to make this more graceful in the future).
But for now the VS Code Go debugger could 'quit' delve nicely when receiving a `DisconnectRequest` from VS Code. This is always the last request sent from VS Code to the debug adapter before killing the debug adapter. And VS Code kills the VS Code Go debugger only after it has received the response from the `DisconnectRequest`. So the debug adapter can delay the killing enough so that delve could do its cleanup.

NexWeb:
There is a better cleaner solution: https://github.com/NexWeb/vscode-go/blob/1cdf2ce2a133623922acd8199411c9f67ca44931/src/debugAdapter/goDebug.ts#L382

chuchiring:
[...]

PR #1379, opened 2017-11-28
Update terminateProcess.sh to delete debug binary.

NexWeb:
RE: #1345 As simple as this fix is, it is working for me. On OS X, normally the debug binary does not get deleted, ever. With this modification, the binary debug file gets deleted as long as there are no pending breakpoints or you have continued past them. Which is appropriate, because if there are existing breakpoints and debug information, Delve can can continue to use the binary until your done.

### DIFF
--- a/scripts/terminateProcess.sh
+++ b/scripts/terminateProcess.sh
@@ -3,8 +3,8 @@
 terminateTree() {
 	for cpid in $(/usr/bin/pgrep -P $1); do
 		terminateTree $cpid
-	done
-	kill -9 $1 > /dev/null 2>&1
+		kill -9 $1 > /dev/null 2>&1
+done
 }
 
 for pid in $*; do


NexWeb:
On MAC OS, normally the debug binary does not get deleted, ever. With this modification, the binary debug file gets deleted as long as there are no existing breakpoints or you have continued past them. Which is appropriate, because if there are existing breakpoints and debug information,  delve can can continue to use the binary until your done.

ramya-rao-a:
@roblourens Can you take a look at this?

scripts/terminateProcess.sh
```diff
@@ -3,8 +3,8 @@
 terminateTree() {
 	for cpid in $(/usr/bin/pgrep -P $1); do
 		terminateTree $cpid
-	done
-	kill -9 $1 > /dev/null 2>&1
+		kill -9 $1 > /dev/null 2>&1
```
ramya-rao-a: @NexWeb Can you explain how this change deletes the binary?
cc @roblourens

ramya-rao-a:
I tried it, the debug binary still doesn't get deleted.
And looking at the code changes, there is nothing there that would delete the binary either

NexWeb:
You tried it how? It works for me, invariably. I guess you're SOL, until you figure out something else.  ;} By the way, there is a video showing the fix in action:  https://youtu.be/E_Nr2qA6bxo

ramya-rao-a:
I used your code (from both the PRs) and debugged a simple Go program. The binary was still there.

NexWeb:
On OSX?  You could try disabling the Lukehoben extension temporarily and extracting the following zip into your extension directory. 
[nexWebVSCode.zip](https://github.com/Microsoft/vscode-go/files/1537335/nexWebVSCode.zip)

roblourens:
What does it actually do? Seems like it's just running the same command to kill the process, but possibly more than once?

NexWeb:
Did you try using the modified extension on OSX?

roblourens:
I wouldn't merge code that I don't understand, even if it works.

NexWeb:
With the current script, the `kill -9 $1 > /dev/null 2>&1 `, being outside the for loop and recursively calling terminateTree, is causing a segmentation fault and obviously not getting the job done.

Consider the following, which works:
```
#!/bin/bash
pids=(521 522 523 524 525)
terminateTree() {
for p in "${pids[@]}"  
do  
    echo "PID is $p"
done  
}
terminateTree 
```
If you place the echo statement after the done and call terminateTree 
 within the loop, the same as it is in terminateProcess.sh, it will fail with a Segmentation fault: 11.

NexWeb:
Having `kill -9 `outside the for loop kills **debug** as well as **debugserver** and **dlv**.
This signal cannot be caught or ignored, and the receiving process, **debug**, cannot perform any clean-up upon receiving this signal.

PR #1380, opened 2017-11-29
Update goDebug.js to enable the removal of the debug binary

NexWeb:
This modification works with PR #1379, RE #1345, to enable the removal of the debug binary when pressing the stop button.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -369,7 +369,7 @@ class Delve {
 	}
 
 	close() {
-		if (!this.debugProcess) {
+		if (this.debugProcess) {
 			this.call<DebuggerState>('Command', [{ name: 'halt' }], (err, state) => {
 				if (err) return logError('Failed to halt.');
 				this.call<DebuggerState>('Restart', [], (err, state) => {


ramya-rao-a:
I am guessing you want the `killTree` to be called when pressing the stop button,
And when the stop button is pressed, if the `debugProcess` still exists, the `killTree` in the `else` block doesnt get called.
And that's why you are changing the `if` condition?

ramya-rao-a:
Closng this PR as this change will break the original reason why the `killTree` was added in the first place.
When debugging is stopped, the processes spawned by the debug process were not getting terminated. That's the reason `killTree` was put in place.

NexWeb:
It's simple. You could try the 2 changes on your locally copy and see it works as expected. I have been using it this way since I made the pull request.
 With the code below,  for pid in $*; terminateTree gets called, and continues to get called recursively without hitting kill because of the done. That's the error and the reason why the debug binary doesn't get deleted. 

```
#!/bin/bash

terminateTree() {
	for cpid in $(/usr/bin/pgrep -P $1); do
		terminateTree $cpid
	done
	kill -9 $1 > /dev/null 2>&1
}

for pid in $*; do
	terminateTree $pid
done

```

ramya-rao-a:
This change reverts the fix done for #438 which repros only in Linux.
The fix for #438 was to run the `killTree` on the id of the debug process so that all the processes spawned from the debug process get terminated.

This change to the `if` condition results in the `killTree` never getting called on the debug process.

PR #1388, opened 2017-12-01
Some proposals for code improvement 

JanKoehnlein:


src/goOutline.ts
```diff
@@ -122,7 +122,7 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
 				label,
 				this.goKindToCodeKind[decl.type],
 				new vscode.Range(document.positionAt(start), document.positionAt(end)),
-				undefined,
+				document.uri,
```
ramya-rao-a: The default value for the `uri` field in the `SymbolInformation` is the current document.
And we are passing blank `containerName`
Both are optional parameters.
We are better off not passing either when creating the `symbolInfo`

JanKoehnlein:
I have removed the uri commit. Are you OK with the other two?

I am currently creating a language server for Go based on the VSCode plug-in. So the errors spotted are mostly found during debugging. I have not yet taken a systematic approach using stricter compiler options or a stricter linter.

ramya-rao-a:
> I am currently creating a language server for Go based on the VSCode plug-in

That's very exciting. Let me know if you have any questions and do keep me updated. You can reach me in the Gophers slack channel.

>  Are you OK with the other two?

Yes.

PR #1410, opened 2017-12-14
Add custom debug command support

wader:
Makes it possible lunch dlv in some custom way.
Demo test project at https://github.com/wader/vscode-go-debug-test

I've used remote debug quite a bit but it is a bit cumbersome to manually start dlv and then attach a  debugger. I did some attempts to add a "docker" mode that would be more smart but it ended up very messy. So instead i thought that it might be just easier to let people start dlv in whatever way they please.

Is this a good idea? should it be called something else than "custom"?

Todo:
- [ ] Add help and mode to go debug launch.json documentation

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -140,6 +140,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 	/** Optional path to .env file. */
 	envFile?: string;
 	backend?: string;
+	custom?: string[]
 }
 
 process.on('uncaughtException', (err: any) => {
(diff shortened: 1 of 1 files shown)


wader:
Not related to this PR but I noticed two things while working on this:

Setting breakpoint in go code where there is no code gives no feedback that breakpoint will be ignored. Create issue?

For some reason it does not work for me to set breakpoint in the extension js code, it used to work some mother ago but now they just go grey on launching the extension in debug mode. I followed the steps in https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code

ramya-rao-a:
Hey @wader,

Can you elaborate on how this can help the remote debugging scenario?

wader:
@ramya-rao-a Hey! sorry i've updated the text in my first comment to make it more clear. Must have written in a hurry or something :)

If you look at the demo test project it probably makes it a bit more clear https://github.com/wader/vscode-go-debug-test/blob/master/.vscode/launch.json My idea was to be able to start dlv in a custom way instead of attaching to a running one. If i remember correctly none of the other debug modes made that possible.

ramya-rao-a:
So you are trying to launch a docker container with PTRACE capability and starting delve inside this container.

How are you passing the image to be used?

wader:
Yes. The image name is hardcoded in the custom lunch argument `vscode-go-debug-test` and is built from the Dockerfile https://github.com/wader/vscode-go-debug-test/blob/master/Dockerfile

So my main idea was to cut down number of steps need for start each debug session. From start program in container and then attach dlv to just start debugger.

marco-m-pix4d:
Hello, is this PR still valid ?

ramya-rao-a:
@wader
Thanks for your efforts here. Unfortunately, we will have to close this PR as we are halting development on the current debug adapter and shifting focus to a new solution based on the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/).

ramya-rao-a:
If you can log an issue describing the issue that you were hoping this PR would resolve, we can consider it for the future.

wader:
@ramya-rao-a Ok! thanks ill look into the debug adapter protocol thingy

My original issue was that i wanted to be able to debug programs and tests using delve inside container using the normal "debug test" etc and not have to start container manually and then attach debugger manually.

I was actually doing some debugging of tests in a container some weeks ago and that made do this hack https://gist.github.com/wader/531c437171b17d66bcfc7be032a3760b maybe can give some ideas. What that hack "debug test" works quite smoothly.

ramya-rao-a:
Thanks @wader 

cc @quoctruong, @polinasok

PR #1416, opened 2017-12-19
Multiline func definition support for go signatures

hummerd:
Go signature does not work for functions declared like this:
``` golang
func VeryLongLongLongLongLongName(
  with int,
  a int,
  lot int,
  of int,
  parameters int) {
}
```
This patch fixes it.


hummerd:
Thank you!

PR #1425, opened 2017-12-25
Check and notify if the file seems to be generated. Fixes #1295.

### LINKED ISSUE #1295: Open generated Go files in readonly mode
The Go community has a standard for files marked as machine-generated: https://golang.org/s/generatedcode
It is expected that the generator may ran again and that any changes made by a human would be overwritten. So those files should not be modified by humans. It would be helpful if VSCode could help to enforce this.

So I propose that when a .go file is opened and it contains the `^// Code generated .* DO NOT EDIT\.$`pattern the readonly mode would be enabled.

cheesedosa:


src/goCheck.ts
```diff
@@ -28,6 +28,16 @@ export function removeTestStatus(e: vscode.TextDocumentChangeEvent) {
 	statusBarItem.text = '';
 }
 
+export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
+	if (e.document.isUntitled) {
```
ramya-rao-a: you can also return if `e.document.languageId !== 'go` so that we can skip the first line check on non Go files

PR #1428, opened 2017-12-26
Fix goreturns error

### LINKED ISSUE #613: Format removes imports of vendored packages in use

halfcrazy:
fix  #613 #630
After https://github.com/sqs/goreturns/pull/42 was merged, then this could be merged.

### DIFF
--- a/src/goFormat.ts
+++ b/src/goFormat.ts
@@ -7,13 +7,14 @@
 
 import vscode = require('vscode');
 import cp = require('child_process');
-import { promptForMissingTool } from './goInstallTools';
+import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 import { sendTelemetryEvent, getBinPath, getToolsEnvVars } from './util';
 
 const missingToolMsg = 'Missing tool: ';
 
 export class Formatter {
 	public formatDocument(document: vscode.TextDocument): Thenable<vscode.TextEdit[]> {
+		let self = this;
 		return new Promise((resolve, reject) => {
 			let filename = document.fileName;
 			let goConfig = vscode.workspace.getConfiguration('go', document.uri);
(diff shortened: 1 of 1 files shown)


src/goFormat.ts
```diff
@@ -27,7 +27,7 @@ export class Formatter {
 			}
 
 			// Fix for https://github.com/Microsoft/vscode-go/issues/613 and https://github.com/Microsoft/vscode-go/issues/630
-			if (formatTool === 'goimports') {
+			if (formatTool === 'goimports' || formatTool === 'goreturns') {
```
ramya-rao-a: For people who haven't updated their `goreturns` this would result in an error of the form `flag provided but not defined: -srcdir`

In that case, we need to prompt the user to update their `goreturns` tool. An example of how we do this is here:
https://github.com/Microsoft/vscode-go/blob/0.6.70/src/goSymbol.ts#L89

halfcrazy: Update. But meet a error. Load the extension and edit a go file then save, the prompt for update show and disappear and show and disappear...

ramya-rao-a: Are you saying that the prompt shows up for each save? If yes, then its fine. If the user presses the `Close` button or presses the `Escape` key, then the prompt will no longer show up in the current session of VS Code.

halfcrazy: Nope, in my case, the prompt shows up and disappear quickly again and again.

ramya-rao-a: So the prompt appears and disappears even without you typing anything? If it happens only when you type, then do you have auto-save on?

halfcrazy: I have auto-save disabled, it happens on manually save.

ramya-rao-a: If you click on the Close button or escape key, does it still keep appearing and disappearing on each file save?

ramya-rao-a:
@halfcrazy I had some refactoring in mind for the `goFormat.ts` file, so I included this change in it as well. See https://github.com/Microsoft/vscode-go/commit/1d7f4b2157ae9c6ba27a3061eee3dca2643f356b

Can you try out the latest vsix and see if you get the repeated prompts on save?
- run `code --install-extension Go-latest.vsix`
- If the above fails with `Error: end of central directory record signature not found`, then get latest from the master branch of this repo and use the `Go-latest.vsix` file from there
- Reload VS Code and test

PR #1438, opened 2018-01-03
Cyclometric complexity code lens

astutecat:
This PR add a new (optional) code lens to show the complexity of functions in a given go source file, using the gocyclo package (https://github.com/fzipp/gocyclo).

The cyclomatic complexity score of a function can be a useful indicator for when a given function has too many branches of logic and needs to be broken apart. 

image

At the present moment the complexity is merely shown to the user, but it could be easily enhanced to include configurable thresholds and associated messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -697,6 +697,11 @@
               "type": "boolean",
               "default": true,
               "description": "If true, enables code lens for running and debugging tests"
+            },
+            "complexity": {
+              "type": "boolean",
+              "default": false,
+              "description": "If true, enables a code lens showing cyclometric complexity. Uses gocyclo."
             }
           },
           "default": {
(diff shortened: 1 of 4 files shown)


ramya-rao-a:
I do believe this is a cool feature to have, although I dont believe this extension is the right place for it. 

As VS Code and this extension has gained popularity, we have seen a gradual scope creep over the past year. Being a sole maintainer of this project, it is getting a little hard to support all incoming requests and maintaining the ever growing set of features. Therefore,one criteria I am using now in the triaging process is whether the feature request would make sense as a separate extension. This current feature request appears to be so.

Thanks for all the work. Let me know if you need any help creating the new extension if you choose to do so.

azr:
Hey @astutecat, I would be interested in such an extension, can it be found anywhere ?

PR #1460, opened 2018-01-14
Show build/vet/lint status in Status bar. Fix #1456

### LINKED ISSUE #1456: Show build/vet/lint status in status bar instead of opening output pane when run manually
Running the build/lint/vet commands manually results in the output pane popping up.
This was initially done to show the linting/building/vetting progress.

There has been feedback that this popup gets annoying after a while.

This issue is to instead have a status bar item or the blue progress bar up top.

This is very easy to do, PRs are welcome and I can help anyone who wants to take a stab at this.

cheesedosa:
I am also maintaining the state in the `outputChannel` which is revealed with the registered command `go.show.outputPanel` when the user clicks on the `statusBarItem` (in case the user wants debug info). Hope it is fine?

### DIFF
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -26,6 +26,7 @@ export function buildCode(buildWorkspace?: boolean) {
 	let documentUri = editor ? editor.document.uri : null;
 	let goConfig = vscode.workspace.getConfiguration('go', documentUri);
 
+	outputChannel.clear(); // Ensures stale output from build on save is cleared
 	diagnosticsStatusBarItem.show();
 	diagnosticsStatusBarItem.text = 'Building...';
 
@@ -33,7 +34,6 @@ export function buildCode(buildWorkspace?: boolean) {
 		.then(errors => {
 			handleDiagnosticErrors(editor ? editor.document : null, errors, vscode.DiagnosticSeverity.Error);
 			diagnosticsStatusBarItem.hide();
-			outputChannel.clear();
 		})
 		.catch(err => {
 			vscode.window.showInformationMessage('Error: ' + err);
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -28,10 +28,12 @@ export function buildCode(buildWorkspace?: boolean) {
 
 	diagnosticsStatusBarItem.show();
 	diagnosticsStatusBarItem.text = 'Building...';
+
 	goBuild(documentUri, goConfig, buildWorkspace)
 		.then(errors => {
 			handleDiagnosticErrors(editor ? editor.document : null, errors, vscode.DiagnosticSeverity.Error);
 			diagnosticsStatusBarItem.hide();
+			outputChannel.clear();
 		})
 		.catch(err => {
 			vscode.window.showInformationMessage('Error: ' + err);
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -6,7 +6,7 @@ import os = require('os');
 import { getNonVendorPackages } from './goPackages';
 import { getTestFlags } from './testUtils';
 import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
-
+import { diagnosticsStatusBarItem } from './goStatus';
 /**
  * Builds current package or workspace.
  */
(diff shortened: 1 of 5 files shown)


src/goBuild.ts
```diff
@@ -23,13 +23,18 @@ export function buildCode(buildWorkspace?: boolean) {
...
+	statusBarItem.show();
+	statusBarItem.text = 'Building in progress';
 	outputChannel.clear();
-	outputChannel.show();
 	outputChannel.appendLine('Building in progress...');
 	goBuild(documentUri, goConfig, buildWorkspace)
-		.then(errors => handleDiagnosticErrors(editor ? editor.document : null, errors, vscode.DiagnosticSeverity.Error))
+		.then(errors => {
+				handleDiagnosticErrors(editor ? editor.document : null, errors, vscode.DiagnosticSeverity.Error);
+				statusBarItem.text = 'Build Complete';
```
ramya-rao-a: I wonder if we should not show anything when the build/lint/vet process via commands is complete.
We might end up in a case where the status bar says "complete" but the on save build/lint/vet processes might be running.

src/goStatus.ts
```diff
@@ -10,8 +10,18 @@ import vscode = require('vscode');
 
 export let outputChannel = vscode.window.createOutputChannel('Go');
 
+export let statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
+statusBarItem.command = 'go.show.outputPanel';
```
ramya-rao-a: We dont need to bind the status bar to a command. It doesnt provide any added benefits.

src/goBuild.ts
```diff
@@ -23,13 +23,18 @@ export function buildCode(buildWorkspace?: boolean) {
 
 	let documentUri = editor ? editor.document.uri : null;
 	let goConfig = vscode.workspace.getConfiguration('go', documentUri);
+	statusBarItem.show();
+	statusBarItem.text = 'Building in progress';
```
ramya-rao-a: Lets shorten this to `Building...`
Some users might have isntalled extensions that are already occupying some space in the status bar. Its best if we occupy as less space as possible

src/goStatus.ts
```diff
@@ -10,8 +10,18 @@ import vscode = require('vscode');
 
 export let outputChannel = vscode.window.createOutputChannel('Go');
 
+export let statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
```
ramya-rao-a: name this `diagnosticsStatusBarItem` so that its intutive

src/goBuild.ts
```diff
@@ -23,13 +23,18 @@ export function buildCode(buildWorkspace?: boolean) {
 
 	let documentUri = editor ? editor.document.uri : null;
 	let goConfig = vscode.workspace.getConfiguration('go', documentUri);
+	statusBarItem.show();
+	statusBarItem.text = 'Building in progress';
 	outputChannel.clear();
-	outputChannel.show();
 	outputChannel.appendLine('Building in progress...');
```
ramya-rao-a: We don't need this anymore. Let's not maintain progress indicators in 2 places.

src/goBuild.ts
```diff
@@ -25,13 +25,17 @@ export function buildCode(buildWorkspace?: boolean) {
 
 	let documentUri = editor ? editor.document.uri : null;
 	let goConfig = vscode.workspace.getConfiguration('go', documentUri);
-	outputChannel.clear();
-	outputChannel.show();
-	outputChannel.appendLine('Building in progress...');
+
+	diagnosticsStatusBarItem.show();
+	diagnosticsStatusBarItem.text = 'Building...';
 	goBuild(documentUri, goConfig, buildWorkspace)
```
ramya-rao-a: One last thing...

What we are trying to achieve here is
- Use the statusbaritem to show progress/failure when build/lint/vet is run on demand.
- Use the outputChannel to show progress/failure when build/lint/vet is run on save.

We should make sure that the statusbaritem is hidden when running build/lint/vet in save and that the outputChannel is cleared when running build/lint/vet on demand.

cheesedosa: Ah, okay. That makes sense.

ramya-rao-a: Actually I meant this: https://github.com/Microsoft/vscode-go/pull/1460/commits/5598cb70b662aa992b65c20085f22a2cd59b6b94

PR #1488, opened 2018-01-29
Only start language server on localhost

philk:
Both for security and avoiding the firewall message on OSX.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -102,7 +102,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				'go-langserver',
 				{
 					command: getBinPath('go-langserver'),
-					args: ['-mode=stdio', ...langServerFlags],
+					args: ['-mode=stdio', '-addr=localhost:4389', ...langServerFlags],
 				},
 				{
 					documentSelector: ['go'],


ramya-rao-a:
@philk Can you elaborate on the need for this a little?

philk:
Sorry. Currently if you start go-language server it binds to `:4389` which as you know in go is actually `0.0.0.0:4389`. Exposing the language server to the local network is a potential security risk, or at least annoyance as on OSX binding an unsigned binary to 0.0.0.0 triggers the firewall to ask if you want to allow/deny go-langserver to listen to network connections.

ramya-rao-a:
I want to keep the option open for a user to override the `addr` flag
Wouldn't it be better to use `localhost` as default in the language server itself?
cc @slimsag and @keegancsmith from sourcegraph

keegancsmith:
Sorry this is an oversight. The language server uses stdio to communicate to vscode-go, but it listens on `:6060` by default to expose a pprof handler. The default should not be to start the pprof server, will update the defaults.

keegancsmith:
Fixed by https://github.com/sourcegraph/go-langserver/commit/c0ef041f24a52e734b83c2f69075ad95df7ff97e

philk:
That fixes it. I assumed it was the -addr in the flags and didn't notice the pprof. Thanks for the (better) fix @keegancsmith !

ramya-rao-a:
Thanks @keegancsmith!

Issue #1491, opened 2018-01-31
Option to use go-find-references for Find all All References 


jonbonazza:
I found this past issue #340 and I understand the problem, but in the current state, Find All References is completely unusable for large projects. The way I see it, we have a couple options here:
1. Take a page from JetBrains' book and index all the files in memory and update the index asynchronously when they change. This way, when you do a Find All References, you don't need to rerun guru and can just consult the index. This is the ideal solution in terms of performance, but also comes with an increase in memory usage, as well as being more rather to implement.
2. Add the option to use the old Find All References method. While not as accurate as Guru, at least it's usable. Some people (myself included) would much rather a slightly less accurate find all references feature that is fast than one that is more accurate but unbearably slow.

I'm not sure what everyone's thoughts are here, but I think option 2 should be easy enough to implement and expose as a configuration item until a better solution is discovered.

ramya-rao-a:
Option 1 is what a language server ought to do, and we have a step forward in that direction. If you are not on Windows, you can try our support for the  [language server from Sourcegraph](https://github.com/Microsoft/vscode-go/blob/master/README.md#go-language-server-experimental). The issue with Windows support is tracked in https://github.com/sourcegraph/go-langserver/issues/113

Option 2 is definitely doable and am open to consider any PRs and help if anyone has questions.

jonbonazza:
I actually implemented this already. I can submit a PR tomorrow.

ramya-rao-a:
Great

ijesonchen:
I tried Go Language Server but it seems has some problem, and Find All References still cost a few seconds (about half of gore?). 
BUT it seem can not find references when projects not in GOPATH. I don't know why. 
It cost half of a day. Any suggest?
Thank you.

jonbonazza:
So I actually realized recently that if you don't have any vendored code, Find All References is actually really fast. It's when you have a lot of vendored code that its speed becomes an issue.

Is there a way to ignore the vendor directory when running Guru?

ramya-rao-a:
@ijesonchen 

> Go Language Server is not a suitable solution, since it does not support multiple GOPATH.

The language server should support multiple GOPATH as far as I know. 
cc @slimsag and @keegancsimth to confirm.

Do you see the multiple gopath when running the command `Go: Current GOPATH`? That should be the same value passed to the language server

@jonbonazza That's a great find. I don't know if guru has a way to ignore the vendor code. Ask in the Gophers slack. Someone there must know.

[...]

Issue #340, opened 2016-05-24
Find all References is very slow


drinktee:
Hellow, I'm a rookie in vscode and have some questions.
I downloaded go tools from https://github.com/golang/tools. Everytime I click 'find all references' , it takes a few seconds to show the results.

vscode version :1.1.1
go version 1.6.2

lukehoban:
We did recently move to use golang's Guru tool to provide Find All References, as it provides a more well maintained foundation than than the previous (https://github.com/redefiance/go-find-references).  

However, Guru is definitely slower.  I expect that's something that would have to be improved in Guru.  I recall seeing some work on improving performance, but not sure how much more progress to expect. 

/cc @alandonovan

adonovan:
I recently changed guru so that it outputs each result as soon as it is available instead of waiting for the complete set before printing the first one.  This makes it much more responsive, though of course if the workspace is very large it may still take a while to run.

Does vscode display guru results in a streaming fashion?  If not, I recommend making it do that.

lukehoban:
Unfortunately `vscode-go` doesn't currently use the streaming results because `vscode`'s UX for find references doesn't support streaming back results.  I expect that's intentional since there are different UX considerations when results may not be complete.  This may be something VS Code could add in the future.

I'm curious though why https://github.com/redefiance/go-find-references was so much faster for this.  Was it cutting some corner in providing results that `guru` is not cutting?  Or is there room to still improve `guru` performance even without relying on streaming results?

adonovan:
Yes, `redefiance` finds all occurrences of a name, including those that resolve to a different declaration, whereas `guru` resolves names precisely using the same algorithm as a compiler.  Since the latter requires type-checking in the general case, it is significantly slower.

ramya-rao-a:
Closing this issue as there is nothing we do in the Go extension for this at the moment. If anyone has any ideas, please drop a comment here and we can re-visit this

ramya-rao-a:
Update: One alternative (if you are not on Windows) is to use the [language server from Sourcegraph](https://github.com/Microsoft/vscode-go/blob/master/README.md#go-language-server-experimental)

PR #1494, opened 2018-02-01
Added configuration to use legacy Find All References

jonbonazza:
This is in response to #1491 .

So there are a couple things here that I am still not sure about, having never written a VSC plugin before and not being super fluent in TS/JS:

There is still one issue that I can't figure out though. Maybe someone else has an idea:

When sideloaded, while there doesn't seem to be a regression when ```go.legacyReference``` is _disabled_ (the default)--that is, it still works as usual without issue, however when ```go.legacyReference``` is _enabled_, it seems to find _some_ references, but most symbols that it should find are missing, and sometimes no symbols are found at all. This legacy code was taken directly from commit history, so Im not sure what I am missing here. Furthermore, I ran the ```go-find-references``` command externally and even it only finds a subset of the expected symbols. I'm not sure if go-find-references is just broken or if I am using it incorrectly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -417,6 +417,11 @@
       "type": "object",
       "title": "Go configuration",
       "properties": {
+        "go.legacyReference": {
+          "typ": "boolean",
+          "default": false,
+          "description": "Use legacy Find All References."
+        },
         "go.buildOnSave": {
           "type": "string",
           "enum": [
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -38,7 +38,8 @@ const allTools: { [key: string]: string } = {
 	'gometalinter': 'github.com/alecthomas/gometalinter',
 	'megacheck': 'honnef.co/go/tools/...',
 	'go-langserver': 'github.com/sourcegraph/go-langserver',
-	'dlv': 'github.com/derekparker/delve/cmd/dlv'
+	'dlv': 'github.com/derekparker/delve/cmd/dlv',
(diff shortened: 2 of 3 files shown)


jonbonazza:
Also, I am not really sure why Travis is failing. Any insight here?

ramya-rao-a:
> Also, I am not really sure why Travis is failing. Any insight here?

Travis is failing due to linting errors. Run `npm run lint` to see what errors these are.

> go-find-references is just broken or if I am using it incorrectly.

Based on https://github.com/Microsoft/vscode-go/issues/340#issuecomment-224313736 I'd guess that `go-find-references` should atleast give all the results that `guru` gives with some additional false positives, but your experiments seem to be concluding otherwise.

One thing to check is the value being passed to `root`. When you tested, were your expected references all inside the folder that was opened in VS Code?

ramya-rao-a:
I wouldnt add the support for `go-find-references` back until we are more satisfied with its results.
If all it does is a fancy text search, then VS Code can do the same out of the box

jonbonazza:
I agree. I was hoping someone here would be able to see sonething stupid i
was doing or missing. Mind giving it a try yourself and see if your results
are different?

jadengore:
Any status updates on this? Would be cool to test out the performance of this. I find the Go Language Server from Sourcegraph to still be quite slow when looking up references.

ramya-rao-a:
Closing as per https://github.com/Microsoft/vscode-go/issues/1491#issuecomment-468966736

PR #1496, opened 2018-02-03
Support resolving symlinks when navigating to symbols

scele:
When building go apps outside of a GOPATH environment (e.g. using
bazel), users may want to construct an artificial GOPATH directory
structure using symlinks, in order to keep using standard go
productivity tools.  In particular, pointing vscode to such an
artificial GOPATH directory allows features like "go to symbol" work
without further configuration.

However, when navigating around a GOPATH that consists of symlinks, we
would like to open the real files, not the symlinks, within vscode.  To
support this, add support for a new "go.gotoSymbol.resolveSymlinks"
configuration setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -679,6 +679,11 @@
           "description": "If false, the import statements will be excluded while using the Go to Symbol in File feature",
           "scope": "resource"
         },
+        "go.gotoSymbol.resolveSymlinks": {
+          "type": "boolean",
+          "default": false,
+          "description": "If true, symlinks will be resolved when using the Go to Symbol feature"
+        },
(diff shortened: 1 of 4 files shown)


scele:
For context, I'm trying to set up vscode debugging + tools with bazel-built binaries (and go dep vendoring) using a setup like this: https://github.com/scele/go-dep-bazel-vscode-example.  To make it work, I needed this feature and also https://github.com/Microsoft/vscode-go/pull/1495.

bennyscetbun:
Hey Scele why have you closed this pr? you ve found an other solution?

PR #1506, opened 2018-02-09
add fillstruct tool

### LINKED ISSUE #1227: Feature request: add fillstruct support
Recently added to vim-go, this seems pretty awesome:

https://github.com/davidrjenni/reftools/tree/master/cmd/fillstruct

I will probably take a stab at adding it as a command myself, but wanted to note it here as an issue first.

buyology:
Closes #1227 

This adds the nice [`fillstruct`](https://github.com/davidrjenni/reftools/tree/master/cmd/fillstruct) tool.

fillstruct


src/goFillStruct.ts
```diff
@@ -0,0 +1,100 @@
...
+		return;
+	}
+	let args = ['-modified', '-file', editor.document.fileName];
+	if (editor.selection.isEmpty) {
+		let offset = byteOffsetAt(editor.document, editor.selection.start);
+		args.push('-offset');
+		args.push(offset.toString());
+	} else if (editor.selection.start.line <= editor.selection.end.line) {
+		args.push('-line');
+		args.push(`${editor.selection.start.line + 1},${editor.selection.end.line + 1}`);
```
ramya-rao-a: When the command is run on a selection, `fillstruct` fails with `invalid value "..." for flag -line: strconv.ParseInt: parsing "...": invalid syntax
`
Looks like the `-line` flag takes a single number and we are passing 2 here.

Also, in what scenario do we expect the user select text and then run the command?

src/goMain.ts
```diff
@@ -434,6 +440,7 @@ function sendTelemetryEventForConfig(goConfig: vscode.WorkspaceConfiguration) {
 		includeImports: goConfig['gotoSymbol'] && goConfig['gotoSymbol']['includeImports'] + '',
 		addTags: JSON.stringify(goConfig['addTags']),
 		removeTags: JSON.stringify(goConfig['removeTags']),
+		fillStruct: JSON.stringify(goConfig['fillStruct']),
```
ramya-rao-a: This is only for configuration. Since there is no config called `fillStruct`, this change is not needed

buyology: Done.

src/goFillStruct.ts
buyology: I had that change sitting on my computer :) Since it takes a `-line`, I thought we might as well support selecting a whole line.

buyology:
Thanks for your comments. Added the tests. 

However — after the refactoring I'm back at that same problem that we discussed on Slack that I can see during the tests that stuff is properly edited, but the tests fail. I cannot spot the obvious missing return in this case however.

ramya-rao-a:
Sorry about that, fixed the issue with the tests.
Also updated to use the first line of the selection as a best guess when selection spans multiple lines.

buyology:
Nice! :)

ramya-rao-a:
Will just wait for the tests to complete and then merge
Thanks!

Issue #1508, opened 2018-02-10
0.6.74 breaks envVars['Path']


malashin:
After updating **vscode-go** to 0.6.74 I could not load **gcc** anymore, even though it was preset in the %PATH% and I was able to call it from builtin terminal.

![img](https://i.imgur.com/KWYDXv1.png)

Checking if that is the case on my other PC proved that it happens right after updating **vscode-go** to 0.6.74.

I was able to pin point the issue to this code https://github.com/Microsoft/vscode-go/commit/37570abab239ffec084cd67cdbd6ce952e8c89cb:
```go
if (goroot && (<string>envVars['PATH'] || '').split(path.delimiter).indexOf(goroot) === -1) {
	envVars['PATH'] += (envVars['PATH'] ? path.delimiter : '') + goroot;
}
```

`envVars['PATH']` does not actually exist _(at least on Windows 7)_, `envVars['Path']` does.

The code can't find the `envVars['PATH']`, but is still able to update `envVars['Path']` with 
`"undefinedC:\tools\go"`. JavaScript is awesome...

trigun117:
![image](https://user-images.githubusercontent.com/23389076/36066155-cc1c0a8a-0ead-11e8-9171-5e73b5ea2a1c.png)
Same problem after update

ramya-rao-a:
Released an update (0.6.75) with the fix.
Thanks for reporting!

PR #1509, opened 2018-02-10
Fix envVars['Path'] in getToolsEnvVars()

malashin:
#1508


ramya-rao-a:
Thanks for the investigation on this issue @malashin!

Turns out I've been using the `PATH` in a few other places, need to fix them as well. So
closing this PR in favor of https://github.com/Microsoft/vscode-go/commit/ef7b65f8e6b93b409357a21ddb117fa3e59243a1

Issue #1291, opened 2017-10-19
Autocompletion of current function doesn't happen in Godoc comment


atombender:
![vsc](https://user-images.githubusercontent.com/50314/31794547-afc28074-b4f0-11e7-9031-3c11685cbf1a.gif)

Say I have this:

```go
// SetT<caret>
func SetThing() {
}
```

I expect code completion to automatically suggest `SetThing` for me. If I type `SetT`, nothing happens. If hit cmd-space to force autocompletion, I might get `SetThing` (or something else starting with `SetT`), but it's a textual autocompletion, not a code completion.

Apologies if this has come up before. I distinctly remember writing about it in a comment, a long time ago.

ramya-rao-a:
There are multiple layers at play here

1. Completions do not appear as you type when inside comments. This is a VS Code feature. You can tweak this by editing the `editor.quickSuggestions` setting
2. When you do force a autocompletion, the Go extension intentionally does not invoke `gocode` for suggestions when typing inside a line comment as it was considered as noise
3. Funnily `gocode` does get invoked when you are inside a block comment. This is because there was no simple way of figuring out the start and end of the block comment without parsing the whole file contents

One solution I have in mind is to invoke gocode for line comments if `editor.quickSuggestions` supported comments. i.e if `editor.quickSuggestions` was set to `true` or `editor.quickSuggestions.comments` was set to `true`, then the Go extension will return suggestions when inside comments.

Or we will have to introduce a new setting just to enable suggestions inside comments

Thoughts?

atombender:
That sounds like a good solution. I didn't know about the `editor.quickSuggestions` setting, thanks.

zh-h:
![default](https://user-images.githubusercontent.com/13273151/36098857-883154f8-103b-11e8-9eec-d3e96f1dfbc9.png)

@atombender  d42e48bf30a3da3936bf1d7f5bbbd8b2e51dd66a This is my solution, but i thing it is useless now😥

ramya-rao-a:
Closing this in favor of #1005

PR #1515, opened 2018-02-12
feat: auto complete exported item`s comment

zh-h:
auto-complete

Auto-completion of exported `type` or `func` .
Inspired by #1291

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -50,8 +50,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let lineTillCurrentPosition = lineText.substr(0, position.character);
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
-				if (lineText.match(/^\s*\/\//)) {
-					return resolve([]);
+				if (lineText.match(/\//)) {  // complete exported item`s comment
+					if (!lineText.match(/^\/\//)) {
+						return resolve([]);
+					}
+					let nextLineText = document.lineAt(position.line + 1).text;
+					if (nextLineText) {
+						let tokens = nextLineText.split(' ');
+						if (tokens.length < 2 || ['type', 'func'].indexOf(tokens[0]) < 0 || tokens[1].match(/^[a-z]/)) {
+							return resolve([]);
+						}
+						let kind = tokens[0];
+						let symbol = tokens[1].split('(')[0];
+						let suggestItem = new vscode.CompletionItem(` ${symbol} `, vscodeKindFromGoCodeClass(kind));
+						suggestItem.detail = 'Exported item`s comment snippet';
+						return resolve([suggestItem]);
+					}
 				}
 
 				let inString = isPositionInString(document, position);
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -52,19 +52,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 
 				if (lineText.match(/\//)) {  // complete exported item`s comment
 					if (!lineText.match(/^\/\//)) {
-						return resolve([]);
+						return resolve([]);  // skip single slash
 					}
 					let nextLineText = document.lineAt(position.line + 1).text;
 					if (nextLineText) {
-						let tokens = nextLineText.split(' ');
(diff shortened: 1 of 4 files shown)


src/goSuggest.ts
```diff
@@ -50,8 +50,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let lineTillCurrentPosition = lineText.substr(0, position.character);
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
-				if (lineText.match(/^\s*\/\//)) {
-					return resolve([]);
+				if (lineText.match(/\//)) {  // complete exported item`s comment
```
ramya-rao-a: This should be `lineText.match(/^\/\//)` and regex test in the next line can be skipped

Other wise in the below scenario where `|` denotes the cursor position, no completions will be returned
`fmt.Print(|)  // Some comment`

src/goSuggest.ts
```diff
@@ -50,8 +50,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+						return resolve([]);
+					}
+					let nextLineText = document.lineAt(position.line + 1).text;
+					if (nextLineText) {
+						let tokens = nextLineText.split(' ');
+						if (tokens.length < 2 || ['type', 'func'].indexOf(tokens[0]) < 0 || tokens[1].match(/^[a-z]/)) {
+							return resolve([]);
+						}
+						let kind = tokens[0];
+						let symbol = tokens[1].split('(')[0];
```
ramya-rao-a: This will fail to find method names with receivers. For example: `func (s *Server) poll() `

src/goSuggest.ts
```diff
@@ -50,8 +50,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
 
-				if (lineText.match(/^\s*\/\//)) {
-					return resolve([]);
+				if (lineText.match(/\//)) {  // complete exported item`s comment
+					if (!lineText.match(/^\/\//)) {
+						return resolve([]);
+					}
+					let nextLineText = document.lineAt(position.line + 1).text;
+					if (nextLineText) {
+						let tokens = nextLineText.split(' ');
```
ramya-rao-a: When there are extra spaces between `func` and the name of the function, this will result in empty text being suggested

src/goSuggest.ts
zh-h: I think the function`s comments only using to completing the named functions😂, it is shot of tests. Thinks your reviews, i will make it pretty.

src/goSuggest.ts
zh-h: Is having other util for picking up function`s name?

ramya-rao-a: I would suggest to use regular expressions

`/^\s*type\s+([A-Z]\S*)/` for types
`/^\s*func\s+\(\s*\w+\s+\*?\w+\s*\)\s+([A-Z]\S*)/` for methods with receivers
`/^\s*func\s+([A-Z]\S*)/` for functions

When you use `let matches = nextLineText.match()` on the above, `matches[1]` should give you the symbol name

src/goSuggest.ts
zh-h: using `lineText.match(/^\/\//)` at first, it will sugguest alot of completions. My solution can skip one slash`s completion.

zh-h:
Duplicated feature https://github.com/Microsoft/vscode-go/issues/1291

zh-h:
Duplicated feature https://github.com/Microsoft/vscode-go/issues/1291

ramya-rao-a:
@zh-h Why is this closed?

zh-h:
@ramya-rao-a Duplicated with `editor.quickSuggestions.comments` feature, also I did not know why the test failed
```


[main 23:47:22] update#ctor - updates are disabled

Error: Error: Cannot find module 'C:\Users\zh-h\NodejsProjects\vscode-go\test'
        at t.handleExtensionTests (c:\Users\zh-h\NodejsProjects\vscode-go\.vscode-test\resources\app\out\vs\workbench\node\extensionHostProcess.js:710:691)
```
Windows 10 Home 64bit
npm@5.5.1
node@v8.9.0

ramya-rao-a:
When `editor.quickSuggestions.comments` is set to `true`, we will only get word based suggestions and snippets because `gocode` doesnt return anything when the current position is inside a comment

PR #1517, opened 2018-02-14
Switch to testfile if already open/visible instead of reopening

kwiesmueller:
Implementing the first feature mentioned in #610.
This PR prevents vscode from overriding the currently open file if the test is already open in another visible editor.
To improve this even further, this could benefit from https://github.com/Microsoft/vscode/issues/15178 to even open the file if it is not visible in the other open editors.

First PR here (even first contribution to vscode) so if something is not ok, please tell me :-)


ramya-rao-a:
Looks good, thanks @kwiesmueller!

ramya-rao-a:
This is now out in the latest update to the Go extension (0.6.77)
Thanks @kwiesmueller!

PR #1535, opened 2018-02-23
Fix internal package import

uudashr:
**Fixes for case:**
package to import: `github.com/uudashr/myproject/internal/config'`
importer package: `/Users/uudashr/go/src/github.com/uudashr/myproject-monitoring/cmd/app-monitor`

Expect: _Shouldn't allowed_


uudashr:
@ramya-rao-a please check

uudashr:
@ramya-rao-a maybe it's a good idea to add unit tests for this project, test that doesn't rely to UI, and relatively fast to run.
I think we have lot codes can be test without requiring UI framework, this will become big saviour.

Haven't figure out how to do it in this project, the idea to separate the test command.
Ex: 
- use variable
  - unit test `SCOPE=unit npm test`
  - ui test `SCOPE=ui npm test`
- dedicated script `npm test-unit`

not sure which follows best practice

src/goPackages.ts
```diff
@@ -238,7 +238,7 @@ export function isAllowToImportPackage(toDirPath: string, currentWorkspace: stri
 	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
 	if (internalPkgFound) {
 		let rootProjectForInternalPkg = path.join(currentWorkspace, pkgPath.substr(0, internalPkgFound.index));
-		return toDirPath.startsWith(rootProjectForInternalPkg);
+		return toDirPath.startsWith(rootProjectForInternalPkg + '/') || toDirPath === rootProjectForInternalPkg;
```
ramya-rao-a: Any chance in Windows this would be `\` instead?

uudashr: Not sure... it based on pkg. It will fail on `let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);` also..
Need Windows machine

uudashr: I see.. will push latest changes

ramya-rao-a:
Refactoring this project for tests has been one thing that I wanted to do for a long time. Having tests that do not rely on the extension host spinning up a new window will definitely help. I am not sure how to go about that at the moment, but I'll give that a try soon

uudashr:
Awesome 👍

ramya-rao-a:
This fix is now available in the latest update (0.6.78) to the Go extension

PR #1560, opened 2018-03-06
WIP: Suggestion to expand func types to func snippets (#1553)

### LINKED ISSUE #1553: Feature Request: Expand func types to func snippets
Given the code:

```go
type HandlerFunc func(ResponseWriter, *Request)
```

I would like to see an expansion setting in vscode-go for expanding `http.HandlerFunc` into one of the following (I'm impartial on which is used):

```go
http.HandlerFunc(func(arg1 http.ResponseWriter, arg2 *http.Request) {
	
})
```

or without the http.HandlerFunc prefix...

```go
func(arg1 http.ResponseWriter, arg2 *http.Request) {
	
}
```

Ideally the cursors would be at each arg, and then inside the function like so:

```go
func(${1:"arg1 http.ResponseWriter"}, ${2:"arg2 *http.Request"}) {
	$3
}
```

`gocode` returns enough information to do this, and it was done in Atom's Go package here: https://github.com/joefitzgerald/go-plus/pull/719

I'm willing to help with this, but have no experience working on extensions for vscode so I'll have to find some time and figure the basics out. I suspect the actual PR will be fairly small once I learn how to get going. 

Any suggestions on how to start developing the vscode-go extension are appreciated.

joncalhoun:
Given a type that defines a function, this change adds a code completion suggestion to expand the type into a closure. Eg:

```go
type Demo func(int) string
```

Will expand into the the suggest with the name `Demo` and upon triggering it the following snippet will be used:

```
Demo(func(${1:arg1} int) {
  $2
})
```

Similarly, given the following type:

```go
type Demo func(i int) string
```

The variable name that exists will be used (`i` in this example).

If accepted, this resolves #1553

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -184,6 +184,23 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 								}
 								item.insertText = new vscode.SnippetString(suggest.name + '(' + paramSnippets.join(', ') + ')');
 							}
+							if (conf.get('useCodeSnippetsOnFunctionSuggest') && suggest.class === 'type' && suggest.type.startsWith('func(')) {
+								let params = parameters(suggest.type.substring(4));
+								let paramSnippets = [];
+								for (let i = 0; i < params.length; i++) {
+									let param = params[i].trim();
+									if (param) {
+										param = param.replace('${', '\\${').replace('}', '\\}');
+										if (!param.includes(' ')) {
+											// If we don't have an argument name, we need to create one
+											param = 'arg' + (i + 1) + ' ' + param;
+										}
+										let arg = param.substr(0, param.indexOf(' '));
+										paramSnippets.push('${' + (i + 1) + ':' + arg + '}' + param.substr(param.indexOf(' '), param.length));
+									}
+								}
+								item.insertText = new vscode.SnippetString(suggest.name + '(func(' + paramSnippets.join(', ') + ') {\n	$' + (params.length + 1) + '\n})');
+							}
 
 							if (wordAtPosition && wordAtPosition.start.character === 0 &&
 								suggest.class === 'type' && !goBuiltinTypes.has(suggest.name)) {


joncalhoun:
@ramya-rao-a I suspect this needs a test, but at first glance I'm not sure I understand the tests very well. I suspect I need to add something here - https://github.com/Microsoft/vscode-go/blob/0.6.77/test/go.test.ts#L698 - but that's about all the further I got. I'll probably dive into them more tomorrow, but I don't use typescript often so any help would probably speed this up.

joncalhoun:
@ramya-rao-a I'd love to figure out what needs done to get this submitted.

I think it needs a test of some sort; can you point me to an area in the tests where you think this would be most easily tested?

After a test is added, is there anything else you feel will need altered in the PR?

ramya-rao-a:
There is just one test failing and that too only on tip, that's unrelated to your PR, I'll take a look at it soon.

Pointer to existing test for completion that you can use as reference: https://github.com/Microsoft/vscode-go/blob/0.6.78/test/go.test.ts#L747

You can add another test file to https://github.com/Microsoft/vscode-go/tree/0.6.78/test/fixtures/completions for your case and test it

joncalhoun:
Thank you!

OneOfOne:
ping

joncalhoun:
So I'm currently swamped and won't have time to finish this for at least a week. If you want to close this in the meantime its fine - I'll just reference it with a new PR when it is ready.

joncalhoun:
:+1: Thanks for the help on this. Psyched to see it make it into master finally 😄

ramya-rao-a:
Yup, I finally got around to adding some tests (See https://github.com/Microsoft/vscode-go/commit/8dd3f331bf4e0f4f6f58d1005132f91acf319bdd)

I'll release an update early next week.

ramya-rao-a:
@joncalhoun I just realized that if the function has a return type, then the snippet doesnt reflect that.

ramya-rao-a:
Take the below code
```
package main

type Demo func(int) string

func main() {

	mycallback :=

	takeDemo(mycallback)
}

func takeDemo(callback Demo) {
	callback(1)
}
```
Next to `mycallback :=`  type `D` to get the completion. You will get

```
mycallback := Demo(func(arg1 int) {

	})
```

which wont compile as there is no return value in the func declaration

joncalhoun:
I think you are right looking the code - I see nothing there for return types. I must have forgot about that use case when writing it, and I haven't touched this code in a while.

Want me to take a stab at fixing it or are you going to? Regardless it might make sense to roll back the change since it can break things 😦

ramya-rao-a:
Fixed with https://github.com/Microsoft/vscode-go/commit/8bb08f6e7b4697f154da1f74539f4d421302db34

There is no need to roll back.
A very small percentage of users have the setting to expand snippets and I doubt among them everyone will encounter this issue. Even if they do, the fix is easy, they just need to add the return type manually.

PR #1564, opened 2018-03-07
Add the output option for the delve

1984weed:
Hi

I added the output option for delve. 
You would be able to use delve, even if there has been `debug` directory already.

Thank you

### DIFF
--- a/package.json
+++ b/package.json
@@ -412,6 +412,11 @@
                   "lldb"
                 ],
                 "description": "Backend used by delve. Only available in delve version 0.12.2 and above."
+              },
+              "output": {
+                "type": "string",
+                "description": "Output path for the bynary of delve",
+                "default": "debug"
               }
             }
           }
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@1984weed  I dont see the `--output` option when running `dlv -h` on version "1.0.0-rc.2". Is this available only on a newer version?

1984weed:
Sorry. It's only about `debug` mode.
https://github.com/derekparker/delve/blob/master/Documentation/usage/dlv_debug.md

ramya-rao-a:
You will need to add the `output` entry at https://github.com/Microsoft/vscode-go/blob/master/package.json#L319 as well, else you won't get a default or intellisense when editing the launch.json file

ramya-rao-a:
@1984weed Can you sign the CLA?

1984weed:
Done !!

ramya-rao-a:
@1984weed This feature is now available in the latest update (0.6.78) to the Go extension. Thanks for the PR!

package.json
```diff
@@ -412,6 +412,11 @@
                   "lldb"
                 ],
                 "description": "Backend used by delve. Only available in delve version 0.12.2 and above."
+              },
+              "output": {
+                "type": "string",
+                "description": "Output path for the bynary of delve",
```
levrik: Typo `binary`

ramya-rao-a: That is fixed in master

PR #1566, opened 2018-03-08
Add benchmark debuging codelens

harkal:
Sometimes it is useful to be able to step into and debug benchmarks. This pull request adds this codelen above benchmark functions.

### DIFF
--- a/src/goRunTestCodelens.ts
+++ b/src/goRunTestCodelens.ts
@@ -113,8 +113,28 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
 				};
 
 				codelens.push(new CodeLens(func.location.range, runBenchmarkCmd));
-			});
 
+				const args = [
+					'-test.bench', func.name,
+					'-test.run', 'a^' // Don't match any test
+				]; 
+				const program = path.dirname(document.fileName);
+				const env = Object.assign({}, this.debugConfig.env, vsConfig['testEnvVars']);
+				const envFile = vsConfig['testEnvFile'];
+				let buildFlags = getTestFlags(vsConfig, null);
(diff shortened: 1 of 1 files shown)


src/goRunTestCodelens.ts
```diff
@@ -113,8 +113,28 @@ export class GoRunTestCodeLensProvider extends GoBaseCodeLensProvider {
 				};
 
 				codelens.push(new CodeLens(func.location.range, runBenchmarkCmd));
-			});
 
+				const args = [
+					'-test.bench', func.name,
+					'-test.run', 'a^' // Don't match any test
+				]; 
+				const program = path.dirname(document.fileName);
```
ramya-rao-a: A lot of code here is similar to whats needed for the debug test codelens. Instead of repeating it, it would be better to pull the common code out of the promise creation

harkal:
@ramya-rao-a I pulled the code out to a single function that generates the command. Let me know if you like it this way, or you want it a member of the class or something.

ramya-rao-a:
We don't really need a function for this. We can have the `config` as a variable without args inside the `getCodeLensForFunctions` function. Then, append the right `args` inside the promise creations to create the `debugTestCmd` and `debugBenchmarkCmd`.

harkal:
config and debugCmd would then need to be copied and then append args and title to the copies or all promises will share the last one. It didn't look that tighty with all the coping and replacing of keys. The function looked more easy to follow. I might find time to refactor it that way at some point, but for now I am afraid I am over my time quota for a quick contribution like this. But feel free to change it anyway you like, the important thing is to have the feature in as it's quite useful! :) Thanks!

ramya-rao-a:
@harkal This is now out in the latest update (0.6.78) to the Go extension. Thanks for the PR!

Issue #1572, opened 2018-03-10
Feature request: use line number + byte offset to highlight errors


dhowden:
It would be great if the error highlighting would begin at the byte offset reported by the compiler (since go 1.9:https://github.com/golang/go/issues/10324).

dhowden:
So it looks like the correct information is making it in, but just not being passed to the renderer correctly: https://github.com/Microsoft/vscode-go/blob/master/src/goLiveErrors.ts#L87

dhowden:
Closing as this has been merged.

Note: go1.10 is known to report some counter-intuitive column locations for certain errors (for instance https://github.com/golang/go/issues/24339, which has been fixed and will be released in go1.11).

ramya-rao-a:
This feature is now available in the latest update (0.6.78) to the Go extension.

PR #1573, opened 2018-03-10
Use col numbers from cmd outputs to give src locations for error/warnings

dhowden:
Errors/warnings parsed from compiler output now include column numbers in source locations.

Both source annotations and "Problems" use the column data.

Starts #1572.

Later improvements could include trying to extract further information so that annotations can be isolated to the problematic tokens instead of marking the whole line after the location column.

Probably needs testing on pre-go1.9 when the column information was not provided in compiler output.


ramya-rao-a:
@dhowden This feature is now available in the latest update (0.6.78) to the Go extension. Thanks for the PR!

PR #1589, opened 2018-03-22
Make tools re-compiling warnings scoped by toolsGopath

FiloSottile:
When switching between workspaces with different goroot and toolsGopath
settings, the editor is suggesting that re-compiling tools is required
because the GOROOT state is global and not toolsGopath-aware.

This makes the global context a map keyed on toolsGopath.

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -66,45 +66,45 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	let useLangServer = vscode.workspace.getConfiguration('go')['useLanguageServer'];
 	setGlobalState(ctx.globalState);
 
-	updateGoPathGoRootFromConfig().then(() => {
+	updateGoPathGoRootFromConfig().then(async () => {
 		const updateToolsCmdText = 'Update tools';
 		interface GoInfo {
 			goroot: string;
 			version: string;
 		}
+		const toolsGoInfo: { [id: string]: GoInfo; } = ctx.globalState.get('toolsGoInfo') || {};
 		const toolsGopath = getToolsGopath() || getCurrentGoPath();
-		let toolsGoInfo: { [id: string]: GoInfo; } = ctx.globalState.get('toolsGoInfo') || {};
-		if (!toolsGoInfo[toolsGopath]) toolsGoInfo[toolsGopath] = { goroot: null, version: null };
+		if (!toolsGoInfo[toolsGopath]) {
+			toolsGoInfo[toolsGopath] = { goroot: null, version: null };
+		}
 		const prevGoroot = toolsGoInfo[toolsGopath].goroot;
 		const currentGoroot: string = process.env['GOROOT'];
-		vscode.window.showInformationMessage(currentGoroot);
-		if (prevGoroot !== currentGoroot && prevGoroot) {
-			vscode.window.showInformationMessage('Your goroot is different than before, a few Go tools may need recompiling', updateToolsCmdText).then(selected => {
+		if (prevGoroot && prevGoroot !== currentGoroot) {
+			vscode.window.showInformationMessage(`Your current goroot (${currentGoroot}) is different than before (${prevGoroot}), a few Go tools may need recompiling`, updateToolsCmdText).then(selected => {
 				if (selected === updateToolsCmdText) {
 					installAllTools(true);
 				}
 			});
 		} else {
-			getGoVersion().then(currentVersion => {
-				if (currentVersion) {
-					const prevVersion = toolsGoInfo[toolsGopath].version;
-					const currVersionString = `${currentVersion.major}.${currentVersion.minor}`;
+			const currentVersion = await getGoVersion();
+			if (currentVersion) {
+				const prevVersion = toolsGoInfo[toolsGopath].version;
+				const currVersionString = `${currentVersion.major}.${currentVersion.minor}`;
 
-					if (prevVersion !== currVersionString) {
-						if (prevVersion) {
-							vscode.window.showInformationMessage('Your Go version is different than before, few Go tools may need re-compiling', updateToolsCmdText).then(selected => {
-								if (selected === updateToolsCmdText) {
-									installAllTools(true);
-								}
-							});
-						}
-						toolsGoInfo[toolsGopath].version = currVersionString;
+				if (prevVersion !== currVersionString) {
+					if (prevVersion) {
+						vscode.window.showInformationMessage('Your Go version is different than before, few Go tools may need re-compiling', updateToolsCmdText).then(selected => {
+							if (selected === updateToolsCmdText) {
+								installAllTools(true);
+							}
+						});
 					}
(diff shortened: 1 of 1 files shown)


FiloSottile:
I signed and emailed the Google / Microsoft CLA. I think you'll need to tick that box manually.

FiloSottile:
@ramya-rao-a ping. I think you'll need to toggle CLA state manually.

ramya-rao-a:
@FiloSottile We usually dont need to do anything from our side. Did you click on the link "sign now" above and sign there?

FiloSottile:
@ramya-rao-a Microsoft has a custom CLA for Google employees, I can't use the online system when contributing as part of my job. But I already signed the non-standard CLA and emailed it to the relevant Microsoft address (I think cla-automation@) and I'm told that now you'll have to check it manually on your side.

ramya-rao-a:
:) There is no button/link here for me to manually do anything in this area, I'll check with a few folks and get back to you.

https://github.com/Microsoft/vscode-go/pull/949 was a PR from another person from Google. I didnt have to manually check anything. 

@stamblerre Can you share what you had to do for the cla?

FiloSottile:
I think something might have indeed happened behind the scenes in #949 because the bot marked it as "cla-already-signed" instead of just signed.

ramya-rao-a:
@FiloSottile Our internal team for Open Source has reached out to the Google legal team which now has agreed that Google employees should use the normal CLA approach. Can you recheck your internal guidance in this matter?

src/goMain.ts
```diff
@@ -57,8 +57,16 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
-		const prevGoroot = ctx.globalState.get('goroot');
-		const currentGoroot = process.env['GOROOT'];
+		interface GoInfo {
+			goroot: string;
+			version: string;
+		}
+		const toolsGopath = getToolsGopath() || getCurrentGoPath();
+		let toolsGoInfo: { [id: string]: GoInfo; } = ctx.globalState.get('toolsGoInfo') || {};
+		if (!toolsGoInfo[toolsGopath]) toolsGoInfo[toolsGopath] = { goroot: null, version: null };
+		const prevGoroot = toolsGoInfo[toolsGopath].goroot;
```
ramya-rao-a: @FiloSottile Is this to cater to the case where you have different goroots as well as toolgopaths for different workspaces. Each of these toolGopaths have the Go tools compiled using the right version of go and so recompiling is not needed?

src/goMain.ts
```diff
@@ -57,8 +57,16 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+		const currentGoroot: string = process.env['GOROOT'];
+		vscode.window.showInformationMessage(currentGoroot);
```
ramya-rao-a: Is this info message needed? This will show up everytime the extension activates which is for every new Go project opened.

src/goMain.ts
```diff
@@ -79,12 +87,13 @@ export function activate(ctx: vscode.ExtensionContext): void {
 								}
 							});
 						}
-						ctx.globalState.update('goVersion', currVersionString);
+						toolsGoInfo[toolsGopath].version = currVersionString;
```
ramya-rao-a: The global state gets updated before waiting for this promise to be resolved, so the current version will never get updated in the global state

src/goMain.ts
```diff
@@ -57,8 +57,16 @@ export function activate(ctx: vscode.ExtensionContext): void {
 
 	updateGoPathGoRootFromConfig().then(() => {
 		const updateToolsCmdText = 'Update tools';
-		const prevGoroot = ctx.globalState.get('goroot');
-		const currentGoroot = process.env['GOROOT'];
+		interface GoInfo {
+			goroot: string;
+			version: string;
+		}
+		const toolsGopath = getToolsGopath() || getCurrentGoPath();
```
ramya-rao-a: In Go versions less than 1.8, it is possible that gopath will not be set, so `toolsGopath` here would be null or undefined

src/goMain.ts
ramya-rao-a: One way around is to check for goroot as well as the version only after the promise returned by `getGoVersion` is resolved

njgraf512:
This would be a super helpful PR to land. I've been having to continuously patch every new release with this to be able to work with multiple workspaces. Are there plans to push this forward?

ramya-rao-a:
@njgraf512 I was just waiting @FiloSottile to get back on my comments. Feel free to jump aboard and address them if you got some time.

src/goMain.ts
vladbarosan: @ramya-rao-a is this still a concern given that 1.12 is out and we dont test for 1.8 anymore

vladbarosan:
@ramya-rao-a @jhendrixMSFT made requested changes. please comment if there is anything else.
Still need to test the scenario. will let you know

vladbarosan:
Tested, everything seems fine. 
How: 
  - Installed 2 versions of Go and set the goroot and the gotoolspath and workspace settings and switched between them
 - Before: kept being asked every time.
 - Now: not anymore

src/goMain.ts
jhendrixMSFT: Only thing that comes to mind is the future when modules have taken over and `GOPATH` becomes obsolete.  It's a larger issue though and I don't think it needs to hold up this PR.

vladbarosan:
@ramya-rao-a since you requested changes initially , all good on your side ?

src/goMain.ts
```diff
@@ -66,36 +66,45 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
-							vscode.window.showInformationMessage('Your Go version is different than before, few Go tools may need re-compiling', updateToolsCmdText).then(selected => {
-								if (selected === updateToolsCmdText) {
-									installAllTools(true);
-								}
-							});
-						}
-						ctx.globalState.update('goVersion', currVersionString);
+			const currentVersion = await getGoVersion();
+			if (currentVersion) {
+				const prevVersion = toolsGoInfo[toolsGopath].version;
```
ramya-rao-a: When this update goes out, `prevVersion` will be `null` for the first time, and users will get the prompt for recompiling.
We need a null check for `prevVersion` just like we do for `prevGoroot`

vladbarosan: yess, good catch

vladbarosan: we do, but just taht is nested , ill update to remove the nested if

vladbarosan: actually i think its fine how it is ? cuz we show the worning only if prev version exists

ramya-rao-a: Ah! I missed the nested check. looks good

FiloSottile:
Thanks everyone for picking this up. ✨ I kept meaning to get back to it but never managed to.

ramya-rao-a:
This feature is out in the latest release(0.10.0) of the Go extension

Issue #1497, opened 2018-02-06
Prompt to add/update buildtags setting, if current file has buildtags


ramya-rao-a:
If current file has build tags and the user hasnt added the same via the `go.buildtags` setting, then the below features may not work as expected
- build
- test
- go to definition
- go to implementation
- find references
- rename

This feature request is to prompt the user in such scenarios to update the `go.buildtags` setting.
The prompt should have an option "Don't show again" which if chosen results in the prompt never being shown again
The prompt should have an option "Open settings" that would open the settings file so that user can add/edit the `go.buildtags` setting

PRs are welcome
I can help anyone who is interested to give this a try

urohit011:
I would like to give it a try, but it will my first time contributing to this project so can I get some pointers to get started?

ramya-rao-a:
Sure!

Basic set up steps are described here: [Building and Debugging the Go extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)

It assumes you have node installed.

Once you are set up, you can follow how we have the "notify user if a generated file is being edited" feature. This is done in the `notifyIfGeneratedFile` function. You can rename it to say `checksOnFileEdit` and add the build tags check in it.

urohit011:
Thank you, will get started right away.

ramya-rao-a:
After looking at attempts to implement this feature #1590 and #3020, I am having second thoughts about the feature being requested in #1497

This is not going to be trivial as ideally, each feature will need to trigger the check for build tags on current file.

When the feature is on a package or workspace level, these checks cannot be done as we the check assumes we are working with a single file. 

When feature is about testing, one will also need to look at `go.testTags` setting.

Since the error message when build tags are not set are simple enough to unpack, I would vote for not implementing this feature

@marcel-basel, @stamblerre, @hyangah, thoughts?

hyangah:
I agree that this is non-trivial in the current state. To add more complexity to the problem, also we have to consider the `GOOS`, `GOARCH`, and other env vars.
  
There are other efforts related to handling sources with build tags - e.g.
https://github.com/golang/go/issues/39005 proposal: build configuration
https://github.com/golang/go/issues/29202 gopls: improve handling of build tags

I leaning towards not implementing this feature yet.

Instead, maybe showing clear messages and indicating that the source is not included in the current build configuration (with some reference) will be nice.

[...]

PR #1590, opened 2018-03-24
prompt user to update go.buildTags if file has the build tags

urohit011:
This one is for #1497, I have added check to prompt the user to update "go.buildTags" settings if the file has build tags that are not specified in "go.buildTags". The prompt has two options, "Don't show again" and "Open settings".

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -51,17 +51,20 @@ export function checksOnFileEdit(e: vscode.TextDocumentChangeEvent) {
 	const openSettings = {title: 'Open settings'};
 
 	for (let i = 0; i < e.document.lineCount; i++) {
-		if (e.document.lineAt(i).text.match(/^\/\/ \+build [\w\s]+$/)) {
-			fileBuildTags.push(...e.document.lineAt(i).text.split(' '));
+		if (e.document.lineAt(i).isEmptyOrWhitespace) {
+			break;
+		}
+		let foundBuildTags = e.document.lineAt(i).text.match(/^\/\/\s*\+build\s+(.+)$/)[1].split(' ');
+		if (foundBuildTags) {
+			foundBuildTags = foundBuildTags.filter(item => item !== '');
+			fileBuildTags.push(...foundBuildTags);
 		}
 	}
-
-	fileBuildTags = fileBuildTags.filter(tag => tag !== '+build' && tag !== '//' && tag !== '' && tag !== '//+build');
 
 	if (fileBuildTags) {
 		for (let i = 0; i < fileBuildTags.length; i++) {
 			if (settingsBuildTags === null || settingsBuildTags.indexOf(fileBuildTags[i]) === -1) {
-				vscode.window.showInformationMessage('Build tags mentioned in the file don not match with "go.buildTags"', neverAgain, openSettings).then(result => {
+				vscode.window.showInformationMessage('Build tags mentioned in the file don not match with "go.buildTags"'+fileBuildTags, neverAgain, openSettings).then(result => {
 					if (result === neverAgain) {
 						ctx.globalState.update('ignoreGeneratedCodeWarning', true);
 					}
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -29,7 +29,7 @@ export function removeTestStatus(e: vscode.TextDocumentChangeEvent) {
 	statusBarItem.text = '';
 }
 
-export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
+export function checksOnFileEdit(e: vscode.TextDocumentChangeEvent) {
 	let ctx = this;
 	if (e.document.isUntitled || e.document.languageId !== 'go') {
 		return;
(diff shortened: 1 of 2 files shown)


urohit011:
@ramya-rao-a will you please take a look at this, I really don't understand why the tests are failing.

urohit011:
Thanks for your observation @ramya-rao-a, and sorry for late reply. I'll make the changes asap.

src/goCheck.ts
```diff
@@ -45,6 +45,34 @@ export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
 			}
 		});
 	}
+
+	let settingsBuildTags = goConfig['buildTags'] ? goConfig['buildTags'].split(' ') : null;
+	let fileBuildTags = [];
+	const openSettings = {title: 'Open settings'};
+
+	for (let i = 0; i < e.document.lineCount; i++) {
```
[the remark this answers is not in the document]
urohit011: You were right, we did not have to loop through all the lines, first I thought of breaking out of the loop after encountering a "package" statement, but empty line is fine.

src/goCheck.ts
```diff
@@ -45,6 +45,34 @@ export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
...
+		if (e.document.lineAt(i).text.match(/^\/\/ \+build [\w\s]+$/)) {
```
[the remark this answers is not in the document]
urohit011: I could not find the list of all the build tags on internet that is why I have used "." to match the build tags.

src/goCheck.ts
```diff
@@ -45,6 +45,34 @@ export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
...
+			fileBuildTags.push(...e.document.lineAt(i).text.split(' '));
+		}
+	}
+
+	fileBuildTags = fileBuildTags.filter(tag => tag !== '+build' && tag !== '//' && tag !== '' && tag !== '//+build');
```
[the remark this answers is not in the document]
urohit011: I could not get rid of the filter totally, but have reduced it. What do you suggest?

urohit011:
Hello @ramya-rao-a I have made a couple of changes. Will you please take a look at it?

src/goCheck.ts
```diff
@@ -45,6 +45,37 @@ export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
...
+
+	let settingsBuildTags = goConfig['buildTags'] ? goConfig['buildTags'].split(' ') : null;
+	let fileBuildTags = [];
+	const openSettings = {title: 'Open settings'};
+
+	for (let i = 0; i < e.document.lineCount; i++) {
+		if (e.document.lineAt(i).isEmptyOrWhitespace) {
+			break;
+		}
+		let foundBuildTags = e.document.lineAt(i).text.match(/^\/\/\s*\+build\s+(.+)$/)[1].split(' ');
```
ramya-rao-a: When there is no match, the `match` function returns `null`. Accessing the array member on `null` will throw error

src/goCheck.ts
```diff
@@ -45,6 +45,37 @@ export function notifyIfGeneratedFile(e: vscode.TextDocumentChangeEvent) {
...
+		if (foundBuildTags) {
+			foundBuildTags = foundBuildTags.filter(item => item !== '');
+			fileBuildTags.push(...foundBuildTags);
+		}
+	}
+
+	if (fileBuildTags) {
+		for (let i = 0; i < fileBuildTags.length; i++) {
```
ramya-rao-a: Take the example of `// +build darwin freebsd netbsd openbsd`
This code here will show the prompt if any of the above is not present in the settings.
But according to https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool,  "a build tag is evaluated as the OR of space-separated options"

So having any one of the above in the build tags is a sufficient.

Also, reading on in the same article, we have , "each option evaluates as the AND of its comma-separated terms" and "A file may have multiple build tags. The overall constraint is the logical AND of the individual constraints"

So looks like a lot more logic needs to be invloved here

urohit011: Okay, I get it, I only have a one question:
If the file has build tags like 
```// +build linux```
```// +build 386```, should it be present in the settings file like
 ```goBuildTags: linux,darwin``` as the goBuildTags holds a string?

ramya-rao-a: @urohit011 The `go.buildTags` takes space separated words.

So in your example the setting should be `"go.buildTags": "linux 386"`

If the file has `// +build abc def xyz`, then the build tags setting should contain either `abc`, `def` or `xyz`

urohit011: Hello @ramya-rao-a, thanks for the explanation, In my question what I actually wanted to know was  what should the ```"go.buildTags"``` should look like if the file contains two separate lines such as:
```// +build linux```
```// +build 386```

ramya-rao-a: It should look like this: `"go.buildTags": "linux 386"`

urohit011: Got it @ramya-rao-a , working on it

ramya-rao-a:
Closing this PR as it hasnt seen any activity in over an year.
The general approach here is right, if anyone wants to pick it up and create a new PR, they are most welcome

Issue #1534, opened 2018-02-23
extension does not handle command-line args that need to be passed to tests


dvelitchkov:
the `go test` supports passing command-line arguments to the test binary. To quote `go test -help`:

     In addition to the build flags, the flags handled by 'go test' itself are:
     -args
            Pass the remainder of the command line (everything after -args)
            to the test binary, uninterpreted and unchanged.
            Because this flag consumes the remainder of the command line,
            the package list (if present) must appear before this flag.

Sadly, the extension isn't aware of this rule, and adding any "-args" to the "go.testFlags" setting does not work, as the extension does not put such arguments last in the arguments list. 

Steps to Reproduce:

1. Write a test file:
```go

import (
    "flag"
    "os"
    "testing"
)

var dbCnn string

func TestMain(m *testing.M) {
    flag.StringVar(&dbCnn, "dbcnn", "", "database connection")
    flag.Parse()
    os.Exit(m.Run())
}

func TestReports(t *testing.T) {
    if dbCnn == "" {
	t.Fatal("db connection not set")
    }
}


```
2. Change Go settings:
```json
"go.testFlags": [
        "-args",
        "-dbcnn",
        "postgres://foo:bar@baz/abc"
    ]
```
3. Run test

Result:
```sh
Running tool: /usr/local/Cellar/go/1.9.3/libexec/bin/go test -args -dbcnn postgres://foo:bar@baz/abc -timeout 30s ***/server/db -run ^TestReports$

flag provided but not defined: -timeout
```
The error is because "-args" is not placed after all other args to `go test`. 

I think the correct solution would be to define a new setting - perhaps "go.testArgs" - an array of arguments to the test binary. 

I would love to work on this myself and submit a PR

ramya-rao-a:
So instead of 

`go test -args -dbcnn postgres://foo:bar@baz/abc -timeout 30s ***/server/db -run ^TestReports$`

we should actually be running

`go test -timeout 30s ***/server/db -run ^TestReports$ -args -dbcnn postgres://foo:bar@baz/abc` ?

dvelitchkov:
Sounds good, I'll try to get this in ASAP!

rschmied:
Hey... I do face the exact same issue when clicking on the "run test" link next to a function in the editor... I want to pass extra args for the test but those have to be appended to the values that are being passed as defaults...

my go.testflags look like this:

```json
[...]

PR #1603, opened 2018-04-02
Enable setting test arguments with '-args' 

dvelitchkov:
Test arguments are now correctly passed in. Users have to set their `go.buildFlags`, e.g.:

```go
"go.buildFlags": [
        "-args",
        "-foo",
        "i_am_foo",
        "-bar",
        "i_am_bar"
    ]
```

### DIFF
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -125,11 +125,11 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 
 		let buildTags: string = testconfig.goConfig['buildTags'];
 		let args: Array<string> = ['test', ...testconfig.flags];
-		
+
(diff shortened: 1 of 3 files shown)


dvelitchkov:
This is for #1534

ramya-rao-a:
I tried the test code and flags you provided, but I get the below error
Am I missing something?

screen shot 2018-04-03 at 11 05 57 pm

dvelitchkov:
Sorry for the delay. 
I think this PR should be closed - this was a bad idea. What I'm doing can be accomplished by setting VS code project debug settings, no need to mess with build flags.

ramya-rao-a:
Can you update your solution in #1534 and close it as well?

PR #1604, opened 2018-04-03
When gotoSymbol.includeGoroot is set, also show symbols from GOROOT

### LINKED ISSUE #1567: Include the standard library in 'Go to Symbol in Workspace...' searching
It would be nice if symbols from the stdlib were included when using 'Go to Symbol in Workspace...'.

It appears that `go-symbols` only accepts one root and making multiple invocations might bet messy, so an alternative approach of having a 'Go to Symbol in Standard Library...' could be sufficient

m90:
Setting the option to true enables the user to include the standard
library located at GOROOT to be included in the results shown on
"Go To Symbol" - resolves #1567

### DIFF
--- a/src/goSymbol.ts
+++ b/src/goSymbol.ts
@@ -7,6 +7,7 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import { getBinPath, getToolsEnvVars, killProcess } from './util';
+import { getGoRuntimePath } from './goPath';
 import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 
 // Keep in sync with github.com/acroca/go-symbols'
@@ -60,8 +61,8 @@ export class GoWorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider
 		}
 
 		let workspaceSymbols = getSymbols(root, query, token, goConfig);
-		let gorootSymbols = goConfig.gotoSymbol.includeGoroot && process.env.GOROOT
-			? getSymbols(process.env.GOROOT, query, token, goConfig)
+		let gorootSymbols = goConfig.gotoSymbol.includeGoroot
+			? getGoroot().then(goroot => getSymbols(goroot, query, token, goConfig))
 			: <GoSymbolDeclaration[]>[];
 
 		return Promise.all([workspaceSymbols, gorootSymbols])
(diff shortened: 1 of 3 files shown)


src/goSymbol.ts
```diff
@@ -51,27 +51,37 @@ export class GoWorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider
...
 			return;
 		}
 
-		return getWorkspaceSymbols(root, query, token).then(results => {
-			let symbols: vscode.SymbolInformation[] = [];
-			convertToCodeSymbols(results, symbols);
-			return symbols;
-		});
+		let workspaceSymbols = getSymbols(root, query, token, goConfig);
+		let gorootSymbols = goConfig.gotoSymbol.includeGoroot && process.env.GOROOT
```
ramya-rao-a: `process.env.GOROOT` will be undefined unless the user has set the env var or added the `go.goroot` setting.
On my Mac, its neither.
We need to run `go env GOROOT` to get the goroot.
See https://github.com/Microsoft/vscode-go/blob/0.6.78/src/goInstallTools.ts#L298 for reference

test/go.test.ts
```diff
@@ -703,11 +703,11 @@ It returns the number of bytes written and any write error encountered.
 				}
 			}
 		});
-		let withoutIgnoringFolders = getWorkspaceSymbols(workspacePath, 'WinInfo', null, configWithoutIgnoringFolders).then(results => {
+		let withoutIgnoringFolders = getSymbols(workspacePath, 'WinInfo', null, configWithoutIgnoringFolders).then(results => {
```
ramya-rao-a: Can you update this test case and add 2 cases for with and without including goroot just like we do here for with and without ignoring folders?

m90: The function that is being tested here does not know anything about the logic that is being introduced in this PR. If we wanted to add test cases for the behavior we would need add 2 full "integration-style" tests for `#provideWorkspaceSymbols`, which is currently not being tested. I would be happy to add these, but wanted to double check if this is what you are talking about?

Alternatively it might also be an option to move the logic that is introduced here into this method.

ramya-rao-a: I meant something like this: https://github.com/Microsoft/vscode-go/pull/1604/commits/bbf329cc43c2dc43f9e9f1888022da902ebb65a9

m90: But does this change in config have any effect on `changeSymbols`? The check is currently done higher up in the call hierarchy if I'm not mistaken.

m90: But does this change in config have any effect on `changeSymbols`? The check is currently done higher up in the call hierarchy if I'm not mistaken.

test/go.test.ts
```diff
@@ -703,13 +703,34 @@ It returns the number of bytes written and any write error encountered.
...
 			assert.equal(results.length, 0);
 		});
+		let withoutIncludingGoroot = getSymbols(workspacePath, 'printf', null, configWithoutIncludeGoroot).then(results => {
+			assert.equal(results.length, 0);
+		});
+		let withIncludingGoroot = getSymbols(workspacePath, 'WinInfo', null, configWithIncludeGoroot).then(results => {
+			assert.equal(results[0].name, 'printf');
+		});
+
 		Promise.all([withIgnoringFolders, withoutIgnoringFolders]).then(() => done(), done);
```
m90: This would still need to include the test cases generated above in order to make the test fail or pass.

ramya-rao-a:
Ah! sorry about that.
I've pushed another commit that will test the symbol provider directly

ramya-rao-a:
The test still fails, I didnt have the time to debug it yet. Can you take a look?

m90:
I actually think it might be nicer to do it the way your intuition and the tests you wrote were suspecting and move the logic of potentially calling twice down a method. I will force push over your changes for doing that so I can reuse the tests you initially added.

m90:
@ramya-rao-a I refactored the code so that your initial testing approach works now, and I think it's a lot nicer that way as all the logic on the `gotoSymbolConfig` happens in the same place. Tests are passing and I'd be happy.

Let me know if you have further remarks.

Issue #1593, opened 2018-03-27
use completion feature comes from go-langserver


leaxoy:
I notice that go-langserver has provider an experimental completion feature, so if we use go-langserver in vscode, should we use completion feature provided by go-langserver?

ramya-rao-a:
Yes, definitely. We should integrate the completion feature from the language server.
PRs are welcome.

e-nikolov:
They seem to have added support for querying interface implementations as well.

https://github.com/sourcegraph/go-langserver/commit/7e016370b5dcb7c929e563a2eea65b1c42ba01e3

leaxoy:
This will meet some error when add new lines to file while using completion feature comes from go-langserver. So I suggest don't harry to integrate it.

leaxoy:
https://github.com/sourcegraph/go-langserver/issues/257

ramya-rao-a:
The changes for this are merged in and will be available in the next week.
The below settings are needed to get this feature
```
"go.useLanguageServer": true,
  "go.languageServerExperimentalFeatures": {
    "autoComplete": true
  }
```

ramya-rao-a:
This feature is now out in the latest update to the Go extension (0.6.81)

PR #1607, opened 2018-04-05
When using go-languageserver, also use code completion feature

m90:
Instead of using go-code, use go-languageserver for providing
code completion results when enabled. Solves #1593

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -264,7 +264,7 @@ function installTools(goVersion: SemVersion, missing?: string[]) {
 		outputChannel.appendLine(''); // Blank line for spacing
 		let failures = res.filter(x => x != null);
 		if (failures.length === 0) {
-			if (missing.indexOf('langserver-go') > -1) {
+			if (missing.indexOf('go-langserver') > -1) {
 				outputChannel.appendLine('Reload VS Code window to use the Go language server');
 			}
 			outputChannel.appendLine('All tools successfully installed. You\'re ready to Go :).');
@@ -388,13 +388,13 @@ export function checkLanguageServer(): boolean {
 
 	// we execute the languageserver using -help so that it will fail and
 	// print all the available flags
-	let langserverSupportsCompletion = true;
+	let helpText = '';
 	try {
-		cp.execFileSync(getBinPath('go-langserver'), ['-help']);
+		helpText = cp.execFileSync(getBinPath('go-langserver'), ['-help']).toString();
 	} catch (err) {
-		let stderr: string = err.stderr.toString();
-		langserverSupportsCompletion = stderr.includes('-gocodecompletion');
+		helpText = (<cp.SpawnSyncReturns<Buffer>>err).stderr.toString();
 	}
+	let langserverSupportsCompletion = helpText.includes('-gocodecompletion');
 
 	if (!langserverSupportsCompletion) {
 		promptForUpdatingTool('go-langserver');
@@ -412,8 +412,3 @@ function allFoldersHaveSameGopath(): boolean {
 	let tempGopath = getCurrentGoPath(vscode.workspace.workspaceFolders[0].uri);
 	return vscode.workspace.workspaceFolders.find(x => tempGopath !== getCurrentGoPath(x.uri)) ? false : true;
 }
-
-
-
-
-
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -370,26 +370,26 @@ function getMissingTools(goVersion: SemVersion): Promise<string[]> {
 }
 
 // If langserver needs to be used, but is not installed, this will prompt user to install and Reload
-// If langserver needs to be used, and is installed, this will return true
-// Returns false in all other cases
-export function checkLanguageServer(): boolean {
+// If langserver needs to be used, and is installed, this will return the list of supported flags
+// Returns null in all other cases
+export function checkLanguageServer(): string[] {
 	let latestGoConfig = vscode.workspace.getConfiguration('go');
-	if (!latestGoConfig['useLanguageServer']) return false;
+	if (!latestGoConfig['useLanguageServer']) return null;
 
 	if (process.platform === 'win32') {
 		vscode.window.showInformationMessage('The Go language server is not supported on Windows yet.');
-		return false;
+		return null;
 	}
 	if (!allFoldersHaveSameGopath()) {
 		vscode.window.showInformationMessage('The Go language server is not supported in a multi root set up with different GOPATHs.');
-		return false;
+		return null;
 	}
 
 	let langServerAvailable = getBinPath('go-langserver') !== 'go-langserver';
 	if (!langServerAvailable) {
 		promptForMissingTool('go-langserver');
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
-		return langServerAvailable;
+		return null;
 	}
 
 	// we execute the languageserver using -help so that it will fail and
@@ -400,14 +400,12 @@ export function checkLanguageServer(): boolean {
 	} catch (err) {
 		helpText = (<cp.SpawnSyncReturns<Buffer>>err).stderr.toString();
 	}
-	let langserverSupportsCompletion = helpText.includes('-gocodecompletion');
 
-	if (!langserverSupportsCompletion) {
-		promptForUpdatingTool('go-langserver');
-		vscode.window.showInformationMessage('Reload VS Code window after updating the Go language server');
-	}
-
-	return langserverSupportsCompletion;
+	// return the list of supported flags so consumers know what features
+	// can be supported
+	return helpText.split('\n')
+		.filter(line => line.trim().startsWith('-'))
+		.map(flag => flag.split(' ')[0]);
 }
 
 function allFoldersHaveSameGopath(): boolean {
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -383,8 +383,25 @@ export function checkLanguageServer(): boolean {
 	if (!langServerAvailable) {
 		promptForMissingTool('go-langserver');
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
+		return langServerAvailable;
 	}
-	return langServerAvailable;
+
+	// we execute the languageserver using -help so that it will fail and
+	// print all the available flags
+	let langserverSupportsCompletion = true;
+	try {
+		cp.execFileSync(getBinPath('go-langserver'), ['-help']);
+	} catch (err) {
+		let stderr: string = err.stderr.toString();
+		langserverSupportsCompletion = stderr.includes('-gocodecompletion');
+	}
+
+	if (!langserverSupportsCompletion) {
+		promptForUpdatingTool('go-langserver');
+		vscode.window.showInformationMessage('Reload VS Code window after updating the Go language server');
+	}
+
+	return langserverSupportsCompletion;
 }
 
 function allFoldersHaveSameGopath(): boolean {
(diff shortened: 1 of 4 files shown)


src/goInstallTools.ts
```diff
@@ -383,8 +383,26 @@ export function checkLanguageServer(): boolean {
 	if (!langServerAvailable) {
 		promptForMissingTool('go-langserver');
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
+		return langServerAvailable;
 	}
-	return langServerAvailable;
+
+	// the languageserver does not know about any -help feature or similar so
```
m90: This is pretty terrible, but it's the best I could come up with in case we want to make sure an up-to-date version of the language server is installed.

Alternatively we could call `go-langserver -version`, but I don't really know what do to with the output as it would currently just print sth like `v2-dev`: https://github.com/sourcegraph/go-langserver/blob/dd4c619b53ae9d18c94e372d23d8710e59f0766b/main.go#L41

src/goInstallTools.ts
```diff
@@ -383,8 +383,25 @@ export function checkLanguageServer(): boolean {
 	if (!langServerAvailable) {
 		promptForMissingTool('go-langserver');
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
+		return langServerAvailable;
 	}
-	return langServerAvailable;
+
+	// we execute the languageserver using --help so that it will fail and
```
m90: This is pretty terrible, but it's the best I could come up with in case we want to make sure an up-to-date version of the language server is installed.

Alternatively we could call go-langserver -version, but I don't really know what do to with the output as it would currently just print sth like v2-dev: https://github.com/sourcegraph/go-langserver/blob/dd4c619b53ae9d18c94e372d23d8710e59f0766b/main.go#L41

src/goInstallTools.ts
```diff
@@ -383,8 +383,25 @@ export function checkLanguageServer(): boolean {
 	if (!langServerAvailable) {
 		promptForMissingTool('go-langserver');
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
+		return langServerAvailable;
 	}
-	return langServerAvailable;
+
+	// we execute the languageserver using -help so that it will fail and
```
m90: This is pretty terrible, but it's the best I could come up with in case we want to make sure an up-to-date version of the language server is installed.

Alternatively we could call go-langserver -version, but I don't really know what do to with the output as it would currently just print sth like v2-dev: https://github.com/sourcegraph/go-langserver/blob/dd4c619b53ae9d18c94e372d23d8710e59f0766b/main.go#L41

OneOfOne:
Ignore my previous comment, this is related to https://github.com/nsf/gocode/issues/510, I opened a ticket https://github.com/sourcegraph/go-langserver/issues/268.

I can confirm this works perfectly with go 1.10.

@ramya-rao-a ping

primalmotion:
works like a charm for me

m90:
@ramya-rao-a Any input on this one?

ramya-rao-a:
@m90 Do you know the reason why completions are not supported by default by the language server? 
Is it because of the memory burden? The help for the `-gocodecompletion` flag has the comment `enable completion (extra memory burden)`. 

If this indeed is the case, then I'd prefer for users to opt-in to use the completions from the language server instead of us providing as a default.

cc @keegancsmith and @slimsag for input on why completions are not supported by default by the language server.

src/goInstallTools.ts
```diff
@@ -383,8 +383,25 @@ export function checkLanguageServer(): boolean {
...
+		helpText = (<cp.SpawnSyncReturns<Buffer>>err).stderr.toString();
+	}
+	let langserverSupportsCompletion = helpText.includes('-gocodecompletion');
+
+	if (!langserverSupportsCompletion) {
+		promptForUpdatingTool('go-langserver');
+		vscode.window.showInformationMessage('Reload VS Code window after updating the Go language server');
+	}
+
+	return langserverSupportsCompletion;
```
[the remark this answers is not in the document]
m90: > Instead, create a new function getLanguageServerFlags that returns all supported flags. 

Just to double check: for `getLanguageServerFlags` you envision sth that reads the output from calling the naked command just above? Or do we already do sth similar elsewhere?

m90:
@ramya-rao-a I think this should reflect the requested changes now, feel free to check and let me know if you feel it'd be nicer to put this behind a preference.

src/goMain.ts
```diff
@@ -87,14 +87,24 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
-		let langServerAvailable = checkLanguageServer();
-		if (langServerAvailable) {
-			let langServerFlags: string[] = vscode.workspace.getConfiguration('go')['languageServerFlags'] || [];
+		let supportedLangServerFlags = checkLanguageServer();
+		if (supportedLangServerFlags) {
+			let configuredLangServerFlags: string[] = vscode.workspace.getConfiguration('go')['languageServerFlags'] || [];
+			configuredLangServerFlags = [
+				...configuredLangServerFlags,
+				'-gocodecompletion',
+				`-func-snippet-enabled=${vscode.workspace.getConfiguration('go')['useCodeSnippetsOnFunctionSuggest']}`
```
ramya-rao-a: The 2 flags for completion should only be added if they are present in `supportedLangServerFlags`, else `go-langserver` will fail

src/goMain.ts
```diff
@@ -87,14 +87,24 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+			];
+
+			let applicableFlags = configuredLangServerFlags.filter(f => {
+				return supportedLangServerFlags.some(supported => f.startsWith(supported));
```
ramya-rao-a: why `startsWith` and not `===` ?

ramya-rao-a:
I've logged a proposal for the language server to use the `initialziationOptions` instead of flags to drive features. See https://github.com/sourcegraph/go-langserver/issues/280

Let's wait on what our friends at sourcegraph have to say there before moving forward here

src/goMain.ts
m90: Because for example the supported flags will contain `-func-snippet-enabled` but we also pass a value like `-func-snippet-enabled=false`.

ramya-rao-a: Ah! Makes sense

ramya-rao-a:
@m90 Can you update this PR as per changes in https://github.com/sourcegraph/go-langserver/issues/280?

m90:
@ramya-rao-a Sure thing. When are you planning to do the next release of the extension? I'm a little swamped right now, so it might take until the weekend or so.

ramya-rao-a:
@m90 In that case, no worries. I can make the changes in your branch and create a vsix file. You can then install that vsix and help with testing?

primalmotion:
I can test if you like. I’m on that branch anyway

m90:
@ramya-rao-a Sounds good, thanks. Let me know when I can give things a test drive.

m90:
@ramya-rao-a So I have been able to unswamp myself and would be happy to still do the requested updates in the coming days. Does that sound like a plan to you or have you already started updating things?

m90:
@ramya-rao-a I pushed the requested changes. Tbh I am not entirely sure how the entire graceful fallback mechanism works now when someone has an older version of `go-langserver` installed, so in case I missed something here, let me know and I will update accordingly. With an up-to-date language server it seems to work like a charm for me.

OneOfOne:
This is weird, I can no longer apply this PR, git apply fails.

src/goInstallTools.ts
```diff
@@ -380,27 +380,42 @@ function getMissingTools(goVersion: SemVersion): Promise<string[]> {
 }
 
 // If langserver needs to be used, but is not installed, this will prompt user to install and Reload
-// If langserver needs to be used, and is installed, this will return true
-// Returns false in all other cases
-export function checkLanguageServer(): boolean {
+// If langserver needs to be used, and is installed, this will return the list of supported flags
+// Returns null in all other cases
+export function checkLanguageServer(): string[] {
```
ramya-rao-a: To keep things simple, lets go with the assumption that the auto-completion feature from language server is available only if the user has the latest language server i.e if the user's version of the language server respects the `initalizationOptions` sent.

Towards that end, we can revert the changes made to `checkLanguageServer` here

src/goMain.ts
```diff
@@ -109,7 +117,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 					revealOutputChannelOn: RevealOutputChannelOn.Never,
 					middleware: {
 						provideDocumentFormattingEdits: (document: vscode.TextDocument, options: FormattingOptions, token: vscode.CancellationToken, next: ProvideDocumentFormattingEditsSignature) => {
-							if (languageServerExperimentalFeatures['format'] === true) {
+							if (languageServerExperimentalFeatures['format']) {
```
ramya-rao-a: I specifically check for `true` here as user can provide anything in the settings and that value will make it through.

src/goMain.ts
```diff
@@ -118,6 +126,12 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				}
 			);
 
+			c.onReady().then(() => {
```
ramya-rao-a: There is already a `c.onReady()` below, can you combine this with that?

src/goMain.ts
```diff
@@ -118,6 +126,12 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+				if (c.initializeResult && c.initializeResult.capabilities && !c.initializeResult.capabilities.completionProvider) {
+					ctx.subscriptions.push(vscode.languages.registerCompletionItemProvider(GO_MODE, new GoCompletionItemProvider(), '.', '\"'));
```
ramya-rao-a: We would get inside this `if` block when the user had enabled the `languageServerExperimentalFeatures['autoComplete']` setting, but the language server is old as well. In this case, we should notify the user that their version of the language server is old and does not support the `auto-complete` feature.

Same for the format feature.

src/goMain.ts
m90: I wasn't allowed to use a non-boolean value when testing, which is why thought this was possible, but maybe that's only a warning? Undid this in any case.

src/goMain.ts
```diff
@@ -118,15 +122,23 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				}
 			);
 
-			ctx.subscriptions.push(c.start());
-
 			c.onReady().then(() => {
+				const capabilities = c.initializeResult && c.initializeResult.capabilities;
+				if (capabilities && !capabilities.completionProvider) {
+					ctx.subscriptions.push(vscode.languages.registerCompletionItemProvider(GO_MODE, new GoCompletionItemProvider(), '.', '\"'));
+					vscode.window.showInformationMessage('Your installed version of `go-langserver` is out of date and does not support code completion. Falling back to default behavior.');
```
m90: @ramya-rao-a Is this an ok phrasing or would you also prompt the user to upgrade? It might also make sense to show a single generic message in case _any_ fallback condition inside this function body was triggered.

src/goMain.ts
```diff
@@ -118,15 +122,23 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+				}
+
 				if (!languageServerExperimentalFeatures['format'] ||
```
m90: Should this check for `=== true` then, too?

src/goMain.ts
ramya-rao-a: @m90 I'll look into the update option later. I've pushed a commit to use generic message

src/goMain.ts
ramya-rao-a: Good point, I've pushed a commit to fix it

PR #1624, opened 2018-04-12
Update CompletionItemKind

leaxoy:


ramya-rao-a:
This is now out in the latest update to the Go extension (0.6.79)
Thanks @leaxoy! I honestly would never have found this inconsistency :)

leaxoy:
You‘re welcome!

PR #1630, opened 2018-04-17
Fix smart package completion

uudashr:
Some people might want to test the `main` package with the existence of `main.go`.
On the `_test.go` file, it will suggest `main` as package name instead of `publisher_test` or `publisher`.

Dir structure example:
```
pubsub
- cmd
   - subscriber
      - main.go
   - publisher
      - main.go
      - main_test.go <-- package main
- doc.go
- pubsub.go
```

Q: Why not suggest `main_test`?
A:  Main package are not going to be reused by others, all function  might going to be non-exported. So the test package not will use the non `_test` pattern.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -531,14 +531,15 @@ export function guessPackageNameFromFile(filePath): Promise<string[]> {
 
 		const proposedPkgName = segments[segments.length - 1];
 
-		if (goFilename.endsWith('_test.go')) {
-			return resolve([proposedPkgName, proposedPkgName + '_test']);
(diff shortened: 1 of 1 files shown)


uudashr:
@ramya-rao-a Should done, but still not sure why it's failed on `tip` version

ramya-rao-a:
Thanks @uudashr! Sorry for the delay, work has been very busy

ramya-rao-a:
This change is now out in the latest update to the Go extension (0.6.79)

PR #1651, opened 2018-05-02
Fix file path expansion for subpackages in go test output.

### LINKED ISSUE #1626: Links to source files in go test output are not correct for when testing all pkgs in workspace
Sample project: [gotestbug.zip](https://github.com/Microsoft/vscode-go/files/1908084/gotestbug.zip)

Steps to Reproduce:

1. Load the sample project in VS Code
2. Invoke the Go Test All Packages in Workspace action.
3. go test output looks like the following:

```
Running tool: C:\Go\bin\go.exe test -short -timeout 30s ./...

--- FAIL: TestDoStuff (0.00s)
	c:\Users\gordon.tyler\go\src\gotestbug\stuff_test.go:10: oh noes!
FAIL
FAIL	gotestbug/stuff	0.060s
Error: Tests failed.
```

The link to the source file is incorrect. It should be `c:\Users\gordon.tyler\go\src\gotestbug\stuff\stuff_test.go`.

doxxx:
This fixes #1626 by using the package name from the go test output combined with the workspace root folder path to expand filenames to their correct path. go test itself buffers all test output in a package until all the tests in a package have completed, so adding our own per-package buffering does not have a visible effect on the latency of test reporting.

### DIFF
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -174,9 +174,9 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 			const processTestResultLine = (line: string) => {
 				testResultLines.push(line);
 				const result = line.match(packageResultLineRE);
-				if (result) {
+				if (result && currentGoWorkspace) {
 					const packageNameArr = result[2].split('/');
-					const baseDir = path.join(workspaceRoot, ...packageNameArr.slice(1));
+					const baseDir = path.join(currentGoWorkspace, ...packageNameArr);
 					testResultLines.forEach(line => outputChannel.appendLine(expandFilePathInOutput(line, baseDir)));
 					testResultLines.splice(0);
 				}
@@ -184,7 +184,14 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 
 			// go test emits test results on stdout, which contain file names relative to the package under test
 			outBuf.onLine(line => processTestResultLine(line));
-			outBuf.onDone(last => last && processTestResultLine(last));
+			outBuf.onDone(last => {
+				if (last) processTestResultLine(last);
+
+				// If there are any remaining test result lines, emit them to the output channel.
+				if (testResultLines.length > 0) {
+					testResultLines.forEach(line => outputChannel.appendLine(line));
+				}
+			});
 
 			// go test emits build errors on stderr, which contain paths relative to the cwd
 			errBuf.onLine(line => outputChannel.appendLine(expandFilePathInOutput(line, testconfig.dir)));
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -8,7 +8,7 @@ import path = require('path');
 import vscode = require('vscode');
 import util = require('util');
 import { parseEnvFile, getGoRuntimePath, getCurrentGoWorkspaceFromGOPATH } from './goPath';
-import { getToolsEnvVars, getGoVersion, LineBuffer, SemVersion, resolvePath, getCurrentGoPath } from './util';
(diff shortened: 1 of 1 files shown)


doxxx:
@ramya-rao-a Is there anything else needed here?

src/testUtils.ts
```diff
@@ -166,9 +168,25 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
...
-			outBuf.onDone(last => last && outputChannel.appendLine(expandFilePathInOutput(last, testconfig.dir)));
+			const packageResultLineRE = /^(ok|FAIL)[ \t]+(.+?)[ \t]+([0-9\.]+s|\(cached\))/; // 1=ok/FAIL, 2=package, 3=time/(cached)
+			const testResultLines: string[] = [];
+
+			const processTestResultLine = (line: string) => {
+				testResultLines.push(line);
+				const result = line.match(packageResultLineRE);
+				if (result) {
+					const packageNameArr = result[2].split('/');
+					const baseDir = path.join(workspaceRoot, ...packageNameArr.slice(1));
```
ramya-rao-a: Using `workspaceRoot` doesn't work always here.

For example, assume I have opened the [golang example repo](https://github.com/golang/example/) in vscode and assume that a test in the `stringutil` package is failing

`workspaceRoot` will be `/Users/myGopath/src/github.com/golang/example`
`packageNameArr` after the slice will be [golang, example, stringutil]
`baseDir` becomes `/Users/myGopath/src/github.com/golang/example/golang/example/stringutil` which is wrong

`currentGoWorkspace` variable on the other hand has what you need. So use that instead
`const baseDir = path.join(currentGoWorkspace, ...packageNameArr);

 Note that `currentGoWorkspace` can be undefined when we cannot determine the gopath or if current file doesn't not belong to current workspace

src/testUtils.ts
```diff
@@ -166,9 +168,25 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
...
+					testResultLines.forEach(line => outputChannel.appendLine(expandFilePathInOutput(line, baseDir)));
+					testResultLines.splice(0);
```
ramya-rao-a: this is to reset the array?

src/testUtils.ts
```diff
@@ -166,9 +168,25 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
...
+			const packageResultLineRE = /^(ok|FAIL)[ \t]+(.+?)[ \t]+([0-9\.]+s|\(cached\))/; // 1=ok/FAIL, 2=package, 3=time/(cached)
+			const testResultLines: string[] = [];
+
+			const processTestResultLine = (line: string) => {
+				testResultLines.push(line);
+				const result = line.match(packageResultLineRE);
+				if (result) {
+					const packageNameArr = result[2].split('/');
+					const baseDir = path.join(workspaceRoot, ...packageNameArr.slice(1));
+					testResultLines.forEach(line => outputChannel.appendLine(expandFilePathInOutput(line, baseDir)));
```
ramya-rao-a: For any which reason, in case `go test` changes the way it outputs result, then we will end up not showing anything in the output channel. I would prefer to pull the printing to the output channel out of the `if` block here

src/testUtils.ts
doxxx: Yes. It removes the elements without allocating a new array.

src/testUtils.ts
doxxx: I've fixed this by emitting any remaining lines at the end of the test output. If the regex doesn't match because the go test output has changed, this should emit all the results at the end. It's not ideal since none of the results will be visible until all of the tests have finished, but there's not much else I can do.

src/testUtils.ts
doxxx: Fixed!

ramya-rao-a:
Thanks!

Issue #1490, opened 2018-01-31
Go development is generally slow when using github.com/uudashr/gopkgs


codingconcepts:
I've found that when using the `github.com/uudashr/gopkgs` version of `gopkgs`, my CPU spikes for a long time and many on-save operations are extremely slow.  If you look in your your Activity Monitor (on MacOS) or presumably Task Manager (on Windows if affected), you'll see `gopkgs` whirring away and not a lot happening in vscode.

I do have a pretty busy GOPATH (96 github.com subfolders etc.) but only open the top-level directory of the project I'm working on.

Since removing `github.com/uudashr/gopkgs` and replacing with `github.com/tpng/gopkgs` my IDE has come back to life but I keep getting prompted to install `github.com/uudashr/gopkgs` for the best experience.  I can confirm that when I _do_ heed the suggestion and re-install the uudashr version, my IDE grinds to a halt again.

ramya-rao-a:
Thanks for reporting @codingconcepts!

We need both the import path and package name for the installed packages to provide certain features in the Go extension. `github.com/tpng/gopkgs` provides only the import path and so we made the move to `github.com/uudashr/gopkgs`

The results from `github.com/uudashr/gopkgs`  gets cached, but the cache expires in 5 seconds. I can see how this can be aggressive in a busy GOPATH

What we can do is set the expiry to be the maximum between 5 seconds and the time it took to run `gopkgs` the previous time.

ramya-rao-a:
@codingconcepts Can you try out the below version of the extension and see if it helps?

- Download https://github.com/Microsoft/vscode-go/blob/master/Go-latest.vsix
- Run `code --install-extensions Go-latest.vsix`
- If the above fails with `Error: end of central directory record signature not found`, then clone this repo and use the Go-latest.vsix file from the cloned repo
- Reload VS Code

codingconcepts:
Hey Ramya, I installed the vsix without issue and have re-installed the two packages I've been ignoring updates for since discovering the performance improvement:

- github.com/ramya-rao-a/go-outline
- github.com/uudashr/gopkgs

Thanks for your help with this, I'll keep an eye on things today and let you know how I get on.

codingconcepts:
Hey Ramya, vscode has become very slow again during the course of the day unfortunately.  I've seen gopkgs consistently spiking to ~13% CPU and guru consistently spiking to ~350% CPU.

ramya-rao-a:
[...]

PR #1658, opened 2018-05-05
Use -workDir on gopkgs to improve performance of packages listing

uudashr:
Use `-workDir` this should improve the performance.

The difference on my case:
Without -workDir
```
time gopkgs |wc -l
    5822
gopkgs  1.95s user 2.57s system 121% cpu 3.740 total
wc -l  0.00s user 0.00s system 0% cpu 3.739 total
```

vs
Using -workDir
```
time gopkgs -workDir=.|wc -l
     836
gopkgs -workDir=.  0.28s user 0.31s system 225% cpu 0.261 total
wc -l  0.00s user 0.00s system 1% cpu 0.259 total
```

The duration 3.740 vs 0.261, the different around 3 sec.

### DIFF
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -13,9 +13,9 @@ import { getImportablePackages } from './goPackages';
 
 const missingToolMsg = 'Missing tool: ';
 
-export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
+export function listPackages(excludeImportedPkgs: boolean = false, workDir?: string): Thenable<string[]> {
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
-	let pkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName);
+	let pkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName, false, workDir);
 
 	return Promise.all([pkgsPromise, importsPromise]).then(([pkgMap, importedPkgs]) => {
 		importedPkgs.forEach(pkg => {
--- a/src/goImport.ts
+++ b/src/goImport.ts
@@ -13,9 +13,9 @@ import { getImportablePackages } from './goPackages';
 
 const missingToolMsg = 'Missing tool: ';
 
-export function listPackages(excludeImportedPkgs: boolean = false, workDir?: string): Thenable<string[]> {
+export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
 	let importsPromise = excludeImportedPkgs && vscode.window.activeTextEditor ? getImports(vscode.window.activeTextEditor.document) : Promise.resolve([]);
-	let pkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName, false, workDir);
+	let pkgsPromise = getImportablePackages(vscode.window.activeTextEditor.document.fileName);
 
 	return Promise.all([pkgsPromise, importsPromise]).then(([pkgMap, importedPkgs]) => {
 		importedPkgs.forEach(pkg => {
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -151,7 +151,7 @@ export function getAllPackages(workDir?: string): Promise<Map<string, string>> {
  */
 export function getImportablePackages(filePath: string, useCache: boolean = false): Promise<Map<string, string>> {
 	let getAllPackagesPromise: Promise<Map<string, string>>;
-	let fileDirPath = path.dirname(filePath);
+	let fileDirPath = fixDriveCasingInWindows(path.dirname(filePath));
 
 	let foundPkgRootDir = pkgRootDirs.get(fileDirPath);
 	let workDir = foundPkgRootDir || fileDirPath;
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -6,19 +6,18 @@ import { isVendorSupported, getCurrentGoPath, getToolsEnvVars, getGoVersion, get
 import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 
 type GopkgsDone = (res: Map<string, string>) => void;
-interface Cache<V> {
-	entry: V;
+interface Cache {
+	entry: Map<string, string>;
 	lastHit: number;
 }
 
 let gopkgsNotified: boolean = false;
 let cacheTimeout: number = 5000;
 
-let gopkgsRunning: Set<string> = new Set<string>();
 let gopkgsSubscriptions: Map<string, GopkgsDone[]> = new Map<string, GopkgsDone[]>();
+let gopkgsRunning: Set<string> = new Set<string>();
 
-let allPkgsCache: Map<string, Cache<Map<string, string>>> = new Map<string, Cache<Map<string, string>>>();
-let allPkgsLastHit: number;
+let allPkgsCache: Map<string, Cache> = new Map<string, Cache>();
 
 function gopkgs(workDir?: string): Promise<Map<string, string>> {
 	let t0 = Date.now();
@@ -88,6 +87,7 @@ function getAllPackagesNoCache(workDir?: string): Promise<Map<string, string>> {
 		let subs = gopkgsSubscriptions.get(workDir);
 		if (!subs) {
 			subs = [];
+			gopkgsSubscriptions.set(workDir, subs);
 		}
 		subs.push(callback);
 
@@ -97,7 +97,6 @@ function getAllPackagesNoCache(workDir?: string): Promise<Map<string, string>> {
 
 			gopkgs().then((pkgMap) => {
 				gopkgsRunning.delete(workDir);
-
 				gopkgsSubscriptions.delete(workDir);
 				subs.forEach((callback) => callback(pkgMap));
 			});
@@ -107,7 +106,7 @@ function getAllPackagesNoCache(workDir?: string): Promise<Map<string, string>> {
 
 /**
  * Runs gopkgs
- * @argument workDir?: string is the working directory of the golang project.
+ * @argument workDir. The workspace directory of the project.
  * @returns Map<string, string> mapping between package import path and package name
  */
 export function getAllPackages(workDir?: string): Promise<Map<string, string>> {
@@ -128,8 +127,8 @@ export function getAllPackages(workDir?: string): Promise<Map<string, string>> {
 		}
 
 		allPkgsCache.set(workDir, {
-			lastHit: new Date().getTime(),
-			entry: pkgs
+			entry: pkgs,
+			lastHit: new Date().getTime()
 		});
 		return pkgs;
 	});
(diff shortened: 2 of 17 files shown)


uudashr:
Test on server failed.

I'cant test on my local machine, using node version 10. I always got `TypeError: Cannot read property 'packageJSON' of undefined`.
Not sure it caused by node or by the vscode. @ramya-rao-a maybe you know anything about this?

uudashr:
Nevermind.. fixed. Just need to delete the yarn.lock and node_modules

uudashr:
@ramya-rao-a please check

src/goPackages.ts
```diff
@@ -66,23 +77,27 @@ function gopkgs(): Promise<Map<string, string>> {
...
+		}
+		subs.push(callback);
 
-				let subs = gopkgsSubscriptions;
-				gopkgsSubscriptions = [];
+		// Ensure only single gokpgs running
+		if (!gopkgsRunning.has(workDir)) {
+			gopkgsRunning.add(workDir);
 
+			gopkgs().then((pkgMap) => {
```
ramya-rao-a: shouldn't we pass the `workDir` to `gopkgs` here?

src/goPackages.ts
```diff
@@ -2,21 +2,32 @@ import vscode = require('vscode');
...
+
+let allPkgsCache: Map<string, Cache> = new Map<string, Cache>();
+
+function gopkgs(workDir?: string): Promise<Map<string, string>> {
 	let t0 = Date.now();
 	return new Promise<Map<string, string>>((resolve, reject) => {
-		const cmd = cp.spawn(getBinPath('gopkgs'), ['-format', '{{.Name}};{{.ImportPath}}'], { env: getToolsEnvVars() });
+		const args = ['-format', '{{.Name}};{{.ImportPath}}'];
+		if (workDir) {
+			args.push('-workDir', workDir);
```
ramya-rao-a: This will fail for users who don't have the latest `gopkgs` i.e they have a version where `workDir` is not a supported.
You'll have to something similar to https://github.com/Microsoft/vscode-go/blob/0.6.79/src/goFormat.ts#L33 to prompt the user to update their tool, and fallback to not using the `workDir`

src/goPackages.ts
```diff
@@ -91,13 +106,15 @@ function getAllPackagesNoCache(): Promise<Map<string, string>> {
...
+	let cache = allPkgsCache.get(workDir);
+	let useCache = cache && (new Date().getTime() - cache.lastHit) < cacheTimeout;
 	if (useCache) {
-		allPkgsLastHit = new Date().getTime();
-		return Promise.resolve(allPkgsCache);
+		cache.lastHit = new Date().getTime();
+		return Promise.resolve(cache.entry);
 	}
 
 	return getAllPackagesNoCache().then((pkgs) => {
```
ramya-rao-a: shouldn't we pass the `workDir` here?

src/goPackages.ts
ramya-rao-a: This will fail for users who don't have the latest `gopkgs` i.e they have a version where `workDir` is not a supported.
You'll have to something similar to https://github.com/Microsoft/vscode-go/blob/0.6.79/src/goFormat.ts#L33 to prompt the user to update their tool, and fallback to not using the `workDir`

src/goPackages.ts
ramya-rao-a: pass `workDir` here

src/goPackages.ts
ramya-rao-a: pass `workDir` here

uudashr:
Wow... I missed a lot. Fixed.

src/goPackages.ts
uudashr: Done

src/goPackages.ts
uudashr: done

src/goPackages.ts
uudashr: done

src/goImport.ts
```diff
@@ -13,9 +13,9 @@ import { getImportablePackages } from './goPackages';
 
 const missingToolMsg = 'Missing tool: ';
 
-export function listPackages(excludeImportedPkgs: boolean = false): Thenable<string[]> {
+export function listPackages(excludeImportedPkgs: boolean = false, workDir?: string): Thenable<string[]> {
```
ramya-rao-a: why is this change needed when `getImportablePackages` can calculate the `workDir` by using the given `fileName`?

src/goPackages.ts
```diff
@@ -66,23 +82,27 @@ function gopkgs(): Promise<Map<string, string>> {
...
+		subs.push(callback);
 
-				let subs = gopkgsSubscriptions;
-				gopkgsSubscriptions = [];
+		// Ensure only single gokpgs running
+		if (!gopkgsRunning.has(workDir)) {
+			gopkgsRunning.add(workDir);
 
+			gopkgs(workDir).then((pkgMap) => {
+				gopkgsRunning.delete(workDir);
```
ramya-rao-a: Entries never get deleted from `gopkgsRunning` when `gopkgs` returns a rejected promise

src/goPackages.ts
```diff
@@ -29,6 +40,11 @@ function gopkgs(): Promise<Map<string, string>> {
 				return promptForMissingTool('gopkgs');
 			}
 
+			if (errchunks && errchunks.join('').startsWith('flag provided but not defined: -workDir')) {
+				promptForUpdatingTool('gopkgs');
+				return Promise.resolve(pkgs);
```
ramya-rao-a: fallback to not using `workdir` by `return gopkgs().then(result => resolve(result))` instead.

src/goPackages.ts
uudashr: No rejection on gopkgs. Positive and negative result will be return by `resolve()`.

Base on your comment, negative should be in form of rejection. Such as:
1. When `gopkgs` binary not found
2. When `gopkgs` needs to update

This is part of quality I think, which I'm okay, we can do this. But maybe after this PR.

src/goPackages.ts
uudashr: Ahh... good idea

src/goImport.ts
uudashr: I need this to pass the test of **"Replace vendor packages with relative path"** it will fail on the assert of *"Relative path for vendor package 9fans.net/go/acme not found"*.

This happen because `gopkgs` recognize vscode workspace as `workDir`, and the workspace dir for the running test is `vscode-go/out/test`, no vendor package under the directory. While actually we are testing vendor package on under `$GOPATH/src/github.com/rogpeppe/godef`, there the vendor package of `9fans.net/go/acme` should be found.

So it only for satisfying the test. The actual running vscode will never use `workDir` param or `listPackages()` (which kind of silly). I can't guess the project root/`workDir` based on  `fileName`. Any suggestion?

Available the options are:
1. Guess golang project dir by filename
2. Change the test workspace dir to `$GOPATH/src/github.com/rogpeppe/godef` above test
3. Pass `workDir` to `listPackages()` (currently use this)

src/goPackages.ts
uudashr: Done

src/goImport.ts
ramya-rao-a: @uudashr This can happen IRL too. A user might open and edit a Go file that is not under the current workspace.

src/goPackages.ts
```diff
@@ -66,23 +83,27 @@ function gopkgs(): Promise<Map<string, string>> {
...
+		if (!subs) {
+			subs = [];
+			gopkgsSubscriptions.set(workDir, subs);
+		}
+		subs.push(callback);
 
-				let subs = gopkgsSubscriptions;
-				gopkgsSubscriptions = [];
+		// Ensure only single gokpgs running
+		if (!gopkgsRunning.has(workDir)) {
```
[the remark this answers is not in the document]
uudashr: Since it already run, we can use the cache.

uudashr:
Fail on test. Mac machine on Travis CI seems out of resources

uudashr:
All good now

src/goPackages.ts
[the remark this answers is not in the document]
uudashr: Made some changes... all the go files under the same package root will share the same cache and `gopkgs` running  instance.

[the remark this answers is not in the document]
ramya-rao-a: @uudashr I was **just** thinking about the same thing over the weekend :) The fact that vendor packages under a higher ancestor is also available to the current file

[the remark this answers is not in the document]
uudashr: Ahaha.. cool 👍

uudashr:
@ramya-rao-a which test? on travis all good.
Is there a way to reproduce?

uudashr:
Issues on gopkgs... for symbolic link

uudashr:
Issues [godirwalk](https://github.com/karrick/godirwalk/pull/4). Just send the owner a PR https://github.com/karrick/godirwalk/pull/4.

Meanwhile the `gopkgs` use the patched version and already work as expected.

@ramya-rao-a could you re-run the test on your machine?

ramya-rao-a:
Re-installing `gopkgs` fails with "unexpected directory layout" for the `godirwalk`. Had to delete the `gopkgs` folder and then install from scratch

uudashr:
Due to this issue https://github.com/golang/go/issues/21057
You have to execute another `go get -u -github.com/uudashr/gopkgs/cmd/gopkgs` if the first attempt fail. 

Try to add -v `go get -u -v -github.com/uudashr/gopkgs/cmd/gopkgs` so we know what happen.

Let me know if still fail, I have to find another way then.

ramya-rao-a:
Second update worked. I'll run some tests tomorrow and get back here.

src/goPackages.ts
```diff
@@ -149,12 +182,21 @@ export function getImportablePackages(filePath: string, useCache: boolean = fals
 				return;
 			}
 
-			let relativePkgPath = getRelativePackagePath(currentFileDirPath, currentWorkspace, pkgPath);
+			if (!foundPkgRootDir) {
```
ramya-rao-a: @uudashr 
You will have to find the `pkgRootDir` upfront and not after getting the results from `gopkgs`. To understand why, take the below case

- If you dont have the `godef` code, then run `go get github.com/rogpeppe/godef` 
- Open the `godef` folder and then open the file `ast.go` in it
- Run the command `Go: Add Import` with the current Go extension and then with your changes and search for `acme`
- In the current Go extension, you will get the result for "9fans.net/go/acme", but you wont get the same with your changes.

This is because, the `acme` package is in the vendor folder which is not a sibling of the directory to which the file `ast.go` file belongs.

uudashr: No, it should works.

Don't know which version `gopkgs` did you use. Since I push the latest changes, gopkgs `-workDir` also works on sub packages. Try to update the `gopkgs`

```
$ gopkgs -format '{{.Name}};{{.ImportPath}}' -workDir=$(go env GOPATH)/src/github.com/rogpeppe/godef/go/ast|grep acme
acme;github.com/rogpeppe/godef/vendor/9fans.net/go/acme
```

And this also covered on the test case *"Replace vendor packages with relative path"*

ramya-rao-a: Works in my Mac but not Windows

uudashr: Just push an update... would you mind to test it on Windows?

ramya-rao-a:
@uudashr That's because linting was failing on master which I just fixed. Sorry about that

uudashr:
@ramya-rao-a no problem. If you have time please run the test on Windows.

ramya-rao-a:
@uudashr The problem in Windows needed your fix as well as https://github.com/Microsoft/vscode-go/pull/1658/commits/0055c20a80922c1261bf64398a1c1fbea6395240.

I made one more commit to print the error message from gopkgs to the console. This should help in troubleshooting I ran into `gopkgs` failure myself a few times as discussed in this PR.

I also added a retry logic for when the "unexpected directory layout" error is encountered.

The unit tests are passing on my Windows machine.

ramya-rao-a:
I've updated https://github.com/Microsoft/vscode-go/issues/1490 with the means to try out these changes

PR #1666, opened 2018-05-07
Update gocode to use mdempsky's fork

douglascamata:
According to nsf, the original maintainer of the project, he won't  be
able to update the project to work well with Go 1.10 because of the
introduction of package caching.


douglascamata:
Need help to make this conditional based on the golang version as it breaks compatibility on Golang version `< 1.10`.

douglascamata:
Will work more on this and reopen later.

Issue #1656, opened 2018-05-04
Add "goformat" to the allowed values for go.formatTool


mbenkmann:
Please add "goformat" to the allowed values for go.formatTool.
goformat is an extended version of gofmt with more options:

https://github.com/mbenkmann/goformat

This tool would also allow you to revisit issue #1011 but all I care about is having VSCode accept "goformat" as a value for go.formatTool. Usage is exactly as gofmt, so the backend can call it exactly the same way.

ramya-rao-a:
I believe you should be able to use `goformat` without any support from the Go extension if it works similar to the other tools we support.

- Add `"go.formatTool": "goformat"` to your settings. (There will be a warning that it is not valid, but ignore that)
- Install the `goformat` tool so that it is available in the $PATH

The extension should just pick it up and run it.

mbenkmann:
It runs it, but I get a perpetual entry in my Problems tab, so that it always has the blue bubble, which makes harder to notice when I have an actual problem.

ramya-rao-a:
That makes sense. PRs are welcome to this option to `go.formatTool`

Just add an entry at https://github.com/Microsoft/vscode-go/blob/0.6.79/package.json#L507. That should fix the warning in the settings.
Also add an entry in https://github.com/Microsoft/vscode-go/blob/0.6.79/src/goInstallTools.ts#L20 which will ensure you get the right prompt when you have chosen `goformat`, but its not installed

mbenkmann:
Thanks for the code pointers. I managed to implement it.
Since goformat allows style options I think it would make sense to implement the following behavior:

If go.formatFlags == [] and go.formatTool == "goformat", then use a default set of flags that corresponds to

* editor.insertSpaces
* editor.tabSize.

I would assume that this automatically fixes issue #1011 if the user is willing/able to set go.formatTool="goformat".

What do you think? Should I add this at
https://github.com/Microsoft/vscode-go/blob/0.6.79/src/goFormat.ts#L31

ramya-rao-a:
Let's not have hidden defaults that way. Sometimes, a user might indeed want to pass no flags. Also, it wouldn't be right for the Go extension to have formatting opinions :). That's purely a personal decision.

mbenkmann:
But that's exactly the point. We should honor the user's setting of editor.insertSpaces. VSCode contains this default:

"[go]": {
    "editor.insertSpaces": false,
    "editor.formatOnSave": true
  },

[...]

PR #1670, opened 2018-05-09
Add support for go.formatTool: "goformat" #1656

mbenkmann:
See issue #1656

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -33,7 +33,7 @@ const allTools: { [key: string]: string } = {
 	'gogetdoc': 'github.com/zmb3/gogetdoc',
 	'goimports': 'golang.org/x/tools/cmd/goimports',
 	'goreturns': 'github.com/sqs/goreturns',
-	'goformat': 'github.com/mbenkmann/goformat/goformat',
+	'goformat': 'winterdrache.de/goformat/goformat',
 	'golint': 'github.com/golang/lint/golint',
 	'gotests': 'github.com/cweill/gotests/...',
 	'gometalinter': 'github.com/alecthomas/gometalinter',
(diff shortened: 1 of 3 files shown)


src/goInstallTools.ts
```diff
@@ -33,6 +33,7 @@ const allTools: { [key: string]: string } = {
 	'gogetdoc': 'github.com/zmb3/gogetdoc',
 	'goimports': 'golang.org/x/tools/cmd/goimports',
 	'goreturns': 'github.com/sqs/goreturns',
+	'goformat': 'github.com/mbenkmann/goformat/goformat',
```
ramya-rao-a: @mbenkmann When I run `go get github.com/mbenkmann/goformat/goformat`, I get the error "package github.com/mbenkmann/goformat/goformat: cannot find package "github.com/mbenkmann/goformat/goformat" in any of:..."

Does it expect that the user has already cloned the repo?

mbenkmann:
I verified that the installation worked from VSCode by removing the package from my GOPATH. There were no issues.
I just tried it again with a blank GOPATH and it worked.

    mkdir /tmp/go
    export GOPATH=/tmp/go
    go get github.com/mbenkmann/goformat/goformat

mbenkmann:
I think I know what the problem is. I have symlinks in the repo to make it buildable outside of GOPATH (for people who like to download stuff in /tmp). I assume you are on Windows and those symlinks don't clone for you, right?

mbenkmann:
Okay. I've fixed the repository. It should work without symlinks now. I've also changed the URL in goInstallTools.ts to the canonical URL which uses my own domain, because I've noticed that go get github/... creates 2 copies of the source tree, one in src/github.com/goformat and one in src/winterdrache.de/goformat

PR #1671, opened 2018-05-12
Fix prevention of completion popup in line comments. (fixes #1659)

### LINKED ISSUE #1659: The "." (dot) inside comment open the suggestion/completion list
Steps to Reproduce:

1.  Create comment on the right side of the code
2. Type "." (dot) and the completion list will opened

screen shot 2018-05-06 at 01 40 54

Expect: The completion should not appear since it's inside the comment

doxxx:
Fixes #1659 by checking for a line comment at the _end_ of the current line and also checking that the cursor position is within the line comment.

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -54,7 +54,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 
 				// prevent completion when typing in a line comment
 				const commentIndex = lineText.indexOf('//');
-				if (commentIndex > 0 && position.character > commentIndex) {
+				if (commentIndex >= 0 && position.character > commentIndex) {
 					return resolve([]);
 				}
 
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -53,8 +53,8 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
 				// prevent completion when typing in a line comment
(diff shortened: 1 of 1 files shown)


src/goSuggest.ts
```diff
@@ -52,7 +52,9 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let lineTillCurrentPosition = lineText.substr(0, position.character);
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
-				if (lineText.match(/^\s*\/\//)) {
+				// prevent completion when typing in a line comment
+				const commentMatch = lineText.match(/\/\/.*$/);
```
ramya-rao-a: why not just find the index directly using `lineText.indexOf('//')` ?

doxxx: .... That's a very good point. :)

src/goSuggest.ts
```diff
@@ -52,7 +52,9 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let lineTillCurrentPosition = lineText.substr(0, position.character);
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
-				if (lineText.match(/^\s*\/\//)) {
+				// prevent completion when typing in a line comment
+				const commentIndex = lineText.indexOf('//');
+				if (commentIndex > 0 && position.character > commentIndex) {
```
ramya-rao-a: commentIndex being 0 is valid. This check should be `commentIndex > -1`

PR #1672, opened 2018-05-12
Fix isPositionInString recognize backtick strings.

### LINKED ISSUE #1231: No quick suggestion inside back-quoted strings
It works inside double-quoted strings, but not in strings like ``` `foobarbaz foo` ```

doxxx:
This fixes #1231 by fixing the isPositionInString function to recognize backtick strings as well as double-quoted strings. This allows normal textual completion to occur in both types of strings.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -287,12 +287,17 @@ export function isPositionInString(document: vscode.TextDocument, position: vsco
 	let lineText = document.lineAt(position.line).text;
 	let lineTillCurrentPosition = lineText.substr(0, position.character);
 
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
@doxxx This fix only helps the case of single line strings inside backticks.
The issue #1231 is mainly for multi-line strings inside backticks.

doxxx:
Yeah, you're right. It's not easy to detect whether the cursor is inside a multi-line backtick string without an AST of the file.

ramya-rao-a:
Exactly :)

PR #1673, opened 2018-05-12
Use code snippets for completion of vars of func type.

### LINKED ISSUE #1287: Function complete doesn't support inline functions
If a local variable like this is defined

```
ensureKeyword := func (k string) {}
```

then `ensureKeyword` autocompletes to `ensureKeyword` without arguments even if `useCodeSnippetsOnFunctionSuggestWithoutType` is true.

doxxx:
This fixes #1287 by enabling code snippet completion of vars declared as a func type.


ramya-rao-a:
Thanks!

Issue #1005, opened 2017-05-19
Autocompletion in declaration comment should suggest nearest declaration before others


atombender:
If I'm writing a comment for a declaration, I expect the struct to be the first thing suggested (and I swear this used to be the case, too):

```go
// <caret>
type HTTPResponseError struct {
```

I would expect autocompletion here to offer `HTTPResponseError` as the first option if I started typing. But I'm getting other suggestions at the top of the list:

![vsc](https://cloud.githubusercontent.com/assets/50314/26256117/e6e0e7ce-3c89-11e7-9c77-b88ef41cc35f.gif)

ramya-rao-a:
@atombender What you see being suggested in the comments is word based completion (coming from VS Code core and not the Go extension) that has no knowledge of the language semantics.

What you are saying can be achieved if the Go extension itself figures out the right completion. Since the Go extension uses `gocode` to get completion items, you can log this feature request in the repo for `gocode`

This will also be easy to implement by a language server, which makes me think I should have a new label called "language-server-candidate" :)

ramya-rao-a:
When we do have a stable language server to use and which implements code completion, we can add this feature to then.

atombender:
Thanks, I [created an issue](https://github.com/nsf/gocode/issues/454).

ramya-rao-a:
If anybody wants to give this a try from the Go extension's side, here are a few pointers
- Completions are not triggered by default in comments as you type. Users have to manually press Ctrl+Space or enable quick suggestions in comments using the setting `editor.quickSuggestions`
- Once completion does get triggered, the Go extension [specifically return empty result](https://github.com/Microsoft/vscode-go/blob/0.6.79/src/goSuggest.ts#L55) after determining that the cursor is in a line comment. You will need to skip this if the next line declares an exported member. Keep in mind to do this only right after `//` or right after the first word in the comment. If the comment already as a whitespace after the first word, we shouldn't be providing any completions

shreyu86:
I am getting familiar with the `vscode-go` extension and adding this functionality. As a first pass I am experimenting with the following:
```typescript
if (lineText.match(/^\s*\/\//)) {
					let nextLine = document.lineAt(position.line + 1).text;
					let part0 = nextLine.split(" ");
					if (!part0[1].match(/^[A-Z]/)) {
						return resolve([]);
					}
				}
```

* If line is comment, then check if the next line has an exported member.
* Checking for exported member by splitting the next line and checking for exported member. 
[...]

PR #1675, opened 2018-05-12
Triggering autocompletion in comments if the next item is an exported member #1005

shreyu86:
This pull request adds functionality described #1005 by triggering autocompletion in comments if the next line is an exported member. This would be handy in adding documentation for exported members.

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -59,12 +59,11 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				if (lineCommentRegex.test(lineTillCurrentPosition) && position.line + 1 < document.lineCount) {
 					let nextLine = document.lineAt(position.line + 1).text.trim();
 					let memberType = nextLine.match(exportedMemberRegex);
+					let suggestionItem: vscode.CompletionItem;
 					if (memberType && memberType.length === 3) {
-						let suggestionItem = new vscode.CompletionItem(memberType[2]);
-						suggestionItem.kind = vscodeKindFromGoCodeClass(memberType[1]);
-						return resolve([suggestionItem]);
+						suggestionItem = new vscode.CompletionItem(memberType[2], vscodeKindFromGoCodeClass(memberType[1]));
 					}
-					return resolve([]);
+					return resolve(suggestionItem ? [suggestionItem] : []);
 				}
 				// prevent completion when typing in a line comment that doesnt start from the beginning of the line
 				const commentIndex = lineText.indexOf('//');
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -56,7 +56,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 					let nextLine = document.lineAt(position.line + 1).text;
 					let memberName = nextLine.replace(/\s+/g, ' ').trim().split(" ");
 					if (!nextLine.startsWith("\t")) {
-						// check for exported qmembers declared in block
+						// check for exported members declared in block
 						memberName.shift()
 					}
 					if (!memberName[0].match(/^[A-Z]/)) {
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -53,13 +53,18 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
 				if (lineText.match(/^\s*\/\//)) {
-					let nextLine = document.lineAt(position.line + 1).text;
-					let memberName = nextLine.replace(/\s+/g, ' ').trim().split(' ');
-					if (!nextLine.startsWith('\t')) {
-						// check for exported members declared in block
-						memberName.shift();
+					if (position.line + 1 < document.lineCount) {
+						let nextLine = document.lineAt(position.line + 1).text;
+						let memberName = nextLine.replace(/\s+/g, ' ').trim().split(' ');
+						if ((!nextLine.startsWith('\t'))) {
+							// check for exported members declared in block
+							memberName.shift();
+						}
+						if (!memberName[0].match(/^[A-Z]/)) {
+							return resolve([]);
+						}
 					}
-					if (!memberName[0].match(/^[A-Z]/)) {
+					else {
 						return resolve([]);
 					}
 				}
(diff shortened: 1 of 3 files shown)


src/goSuggest.ts
```diff
@@ -53,7 +53,15 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				let autocompleteUnimportedPackages = config['autocompleteUnimportedPackages'] === true && !lineText.match(/^(\s)*(import|package)(\s)+/);
 
 				if (lineText.match(/^\s*\/\//)) {
-					return resolve([]);
+					let nextLine = document.lineAt(position.line + 1).text;
```
ramya-rao-a: If cursor is in the last line of the file, this will result in error

src/goSuggest.ts
```diff
@@ -53,7 +53,15 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+					let memberName = nextLine.replace(/\s+/g, ' ').trim().split(' ');
+					if (!nextLine.startsWith('\t')) {
```
ramya-rao-a: I am guessing your assumption here is that the file is always formatted, which might not be the case. Instead look for `func`, `var` and `type` declarations in the `nextLine`.

src/goSuggest.ts
```diff
@@ -53,7 +53,15 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+						// check for exported members declared in block
+						memberName.shift();
+					}
+					if (!memberName[0].match(/^[A-Z]/)) {
```
ramya-rao-a: Say we found an exported member. In that case we should return a single item (the member name) in the suggestion list. Currently `gocode` gets called and it can return multiple results.

src/goSuggest.ts
shreyu86: Thanks for the feedback. I will add logic to check for this and include an test case.

src/goSuggest.ts
shreyu86: Will take these considerations into account and modify code.

src/goSuggest.ts
shreyu86: In this case gocode should not be invoked right?

ramya-rao-a: Yes, we shouldn't invoke `gocode`. Instead we should return a single completion item with the member name right here

ramya-rao-a:
Also, there have been conflicting changes in master. So first, merge from the master branch.

src/goSuggest.ts
shreyu86: Cool, will do.

shreyu86:
Given the change and removal of `if (lineText.match(/^\s*\/\//))` what would be the preference to proceed and check for condition where it is not a line comment and trigger the completion?

src/goSuggest.ts
```diff
@@ -55,7 +55,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
 				const commentIndex = lineText.indexOf('//');
 				if (commentIndex >= 0 && position.character > commentIndex) {
-					return resolve([]);
+					if (!lineText.trim().startsWith('//')) {
+						return resolve([]);
+					}
+					// triggering completions in comments on exported members
+					if (position.line + 1 < document.lineCount) {
+						let nextLine = document.lineAt(position.line + 1).text.trim();
+						let memberType = nextLine.match(/(const|func|type|var)\s*(\w+)/);
```
ramya-rao-a: We need atleast one space between the keyword and member name, so `\s*` here should be `s+`

test/fixtures/completions/exportedMemberDocs.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+const (
+	//
+	HelloStatus = 0
+	//
+	GreetingStatus = 1
+)
+
+//
```
ramya-rao-a: Keep the cursor between the two `/` and manually trigger completions by pressing `Ctrl+Space`. We get the completion for `SayHello`, but we shouldn't

src/goSuggest.ts
```diff
@@ -55,7 +55,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				// prevent completion when typing in a line comment
 				const commentIndex = lineText.indexOf('//');
 				if (commentIndex >= 0 && position.character > commentIndex) {
-					return resolve([]);
+					if (!lineText.trim().startsWith('//')) {
```
ramya-rao-a: To avoid a lot of nesting, you can pull your code out and place it before line 55. You can start with
```
// triggering completions in comments on exported members
if (lineText.trim().startsWith('//') && (position.line + 1 < document.lineCount)) {
   let nextLine = .....
}
```

src/goSuggest.ts
```diff
@@ -55,7 +55,22 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+						return resolve([]);
+					}
+					// triggering completions in comments on exported members
+					if (position.line + 1 < document.lineCount) {
+						let nextLine = document.lineAt(position.line + 1).text.trim();
+						let memberType = nextLine.match(/(const|func|type|var)\s*(\w+)/);
+						if (memberType && memberType.length === 3) {
```
ramya-rao-a: In the test file, at lines 11, 13, 18, 20 and 35 we get multiple completions from `gocode` and not from the code we specifically wrote for exported members.

This is because we don't exit with empty results when there is no match with the regex above.

shreyu86: So in case of a block declaration, we should not trigger any completions and exit with empty results?

ramya-rao-a: Yes, unless you want to scan through the lines above and below the current line to look for the starting and ending of the block

shreyu86: I think I am fine with skipping, wanted to clarify before I made any changes.

src/goSuggest.ts
shreyu86: Added the right regex.

test/fixtures/completions/exportedMemberDocs.go
shreyu86: Have added logic to handle this case.

shreyu86:
Thanks a lot! @ramya-rao-a

ramya-rao-a:
Thanks for all the work @shreyu86!
I've made 2 small changes, take a look if you want to :)

shreyu86:
Thanks a lot for the refactor, I really appreciate the swift feedback.

PR #1678, opened 2018-05-15
Add package-lock.json to source control

m90:
Right now this repo seems to not check in `package-lock.json`. Still it will be created on each `npm install` (which has me undo each commit at least once :jack_o_lantern:). This would prevent npm from generating a lockfile.

Alternatively we could also check in the lock file (personally I'm very undecided about what is the better option in this context).

I just think it'd be a nice addition to prevent accidental check-ins of `package-lock.json`  from happening.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3584 @@
+{
+  "name": "Go",
+  "version": "0.6.79",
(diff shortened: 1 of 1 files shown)


m90:
@ramya-rao-a Here we go.

PR #1681, opened 2018-05-17
Disallow import internal root package

uudashr:
This should ignore any import of internal packages on the root such as:
```
internal/syscall/windows/registry
internal/nettrace
internal/singleflight
internal/cpu
internal/testenv
internal/trace
internal/testlog
internal/syscall/windows/sysdll
internal/syscall/windows
internal/syscall/unix
internal/poll
internal/race
```

### DIFF
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -277,10 +277,14 @@ export function getNonVendorPackages(folderPath: string): Promise<string[]> {
 // see: https://golang.org/doc/go1.4#internalpackages
 // see: https://golang.org/s/go14internal
 function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string) {
-	let internalPkgFound = pkgPath.match(/(^|\/)internal(\/|$)/);
+	if (pkgPath.startsWith('internal/')) {
+		return false;
+	}
+
+	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
 	if (internalPkgFound) {
 		let rootProjectForInternalPkg = path.join(currentWorkspace, pkgPath.substr(0, internalPkgFound.index));
-		return (rootProjectForInternalPkg !== currentWorkspace && toDirPath.startsWith(rootProjectForInternalPkg + path.sep)) || toDirPath === rootProjectForInternalPkg;
+		return toDirPath.startsWith(rootProjectForInternalPkg + path.sep) || toDirPath === rootProjectForInternalPkg;
 	}
 	return true;
 }
--- a/src/goPackages.ts
+++ b/src/goPackages.ts
@@ -279,8 +279,12 @@ export function getNonVendorPackages(folderPath: string): Promise<string[]> {
 function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string) {
 	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
 	if (internalPkgFound) {
+		internalPkgFound = pkgPath.match(/^internal\b/);
+	}
+
+	if (internalPkgFound) {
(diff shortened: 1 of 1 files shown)


src/goPackages.ts
```diff
@@ -278,9 +278,13 @@ export function getNonVendorPackages(folderPath: string): Promise<string[]> {
 // see: https://golang.org/s/go14internal
 function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string) {
 	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
+	if (internalPkgFound) {
```
ramya-rao-a: shouldnt this be `if(!internalPkgFound)`?

Your case is that `pkgPath` has a value like `internal/nettrace` for which `internalPkgFound` would be null at this point.

ramya-rao-a:
Also, am curious. If no one can import such packages, how do they get used?

src/goPackages.ts
uudashr: Thank for the correct.. will fix soon

uudashr:
based on https://golang.org/s/go14internal
> An import of a path containing the element “internal” is disallowed if the importing code is outside the tree rooted at the parent of the “internal” directory.

```
$ grep -r 'internal/nettrace' "$(go env GOROOT)/src"
/usr/local/Cellar/go/1.10.2/libexec/src/go/build/deps_test.go:		"internal/nettrace", "internal/poll",
/usr/local/Cellar/go/1.10.2/libexec/src/go/build/deps_test.go:		"internal/nettrace",
/usr/local/Cellar/go/1.10.2/libexec/src/go/build/deps_test.go:	"net/http/httptrace": {"context", "crypto/tls", "internal/nettrace", "net", "reflect", "time"},
/usr/local/Cellar/go/1.10.2/libexec/src/net/dial.go:	"internal/nettrace"
/usr/local/Cellar/go/1.10.2/libexec/src/net/http/httptrace/trace.go:	"internal/nettrace"
/usr/local/Cellar/go/1.10.2/libexec/src/net/http/transport_test.go:	"internal/nettrace"
/usr/local/Cellar/go/1.10.2/libexec/src/net/lookup.go:	"internal/nettrace"
```
It can be used by all the source reside on GOROOT

uudashr:
@ramya-rao-a please check

src/goPackages.ts
```diff
@@ -277,10 +277,10 @@ export function getNonVendorPackages(folderPath: string): Promise<string[]> {
 // see: https://golang.org/doc/go1.4#internalpackages
 // see: https://golang.org/s/go14internal
 function isAllowToImportPackage(toDirPath: string, currentWorkspace: string, pkgPath: string) {
-	let internalPkgFound = pkgPath.match(/\/internal\/|\/internal$/);
+	let internalPkgFound = pkgPath.match(/(^|\/)internal(\/|$)/);
 	if (internalPkgFound) {
 		let rootProjectForInternalPkg = path.join(currentWorkspace, pkgPath.substr(0, internalPkgFound.index));
-		return toDirPath.startsWith(rootProjectForInternalPkg + path.sep) || toDirPath === rootProjectForInternalPkg;
+		return (rootProjectForInternalPkg !== currentWorkspace && toDirPath.startsWith(rootProjectForInternalPkg + path.sep)) || toDirPath === rootProjectForInternalPkg;
```
ramya-rao-a: Wouldn't this return `true` for pkgPath which is say `internal/something` and user is working on a file directly under `GOPATH/src`?

uudashr: Impossible to work under `GOPATH/src`

Ahh.. yes.. correct.

uudashr:
More simple with your way

PR #1699, opened 2018-05-31
Add support for revive as lintTool

mgechev:
Introduces support for revive. Revive is an extensible, configurable, and faster alternative of golint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This extension adds rich language support for the Go language to VS Code, includ
 - Workspace symbol search (using `go-symbols`)
 - Rename (using `gorename`. Note: For Undo after rename to work in Windows you need to have `diff` tool in your path)
 - Build-on-save (using `go build` and `go test`)
-- Lint-on-save (using `golint` or `gometalinter` or `megacheck` or `golangci-lint`)
+- Lint-on-save (using `golint` or `gometalinter` or `megacheck` or `golangci-lint` or `revive`)
 - Format on save as well as format manually (using `goreturns` or `goimports` or `gofmt`)
 - Generate unit tests skeleton (using `gotests`)
 - Add Imports (using `gopkgs`)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -98,6 +98,14 @@ You can configure golangci-lint with `go.lintFlags`, for example to show issues
   "go.lintFlags": ["--enable-all", "--new"],
 ```
 
+An alternative of golint is [revive](https://github.com/mgechev/revive). It extensible, configurable, provides superset of the rules of golint rules, and has significantly better performance.
+
+To configure revive, use:
+
+```javascript
+  "go.lintFlags": ["-config=[PATH_TO_CONFIG.toml]"]
```
ramya-rao-a: Users might want to use the `${workspaceFolder}` or `${workspaceRoot}` variables for the config file. We support this for the config file used by `gometalinter`. You might want to do something similar for revive.

See https://github.com/Microsoft/vscode-go/blob/master/src/goLint.ts#L67

wolfeidau:
@mgechev are you going to fix this feedback so we can try it out?

Keen to use this!

Cheers

README.md
mgechev: Just updated the readme & tested the `${workspaceRoot}`/`${workspaceFolder}` support locally. Everything seems to be functioning correctly.

ramya-rao-a:
Thanks for the PR!

I'll update this thread soon with steps to try this out before I release the next update.

mgechev:
Sounds good! I'm already using it in my project:
 
cyclomatic

Would definitely verify again once you share the steps.

ramya-rao-a:
To test this,

- Reload VS Code
- Change the `go.lintTool` setting to `revive`

mgechev:
@ramya-rao-a verified that everything works as expected. Tested the config file as well 💯

Issue #1698, opened 2018-05-31
Test Plan for May 2018 update (0.6.81)


ramya-rao-a:
If you got a notification from this issue, then you are part of the next update to the Go extension.
Thanks and Welcome!

As most of you know already, I am the sole maintainer of this project at the moment. I wanted to try something different for this next update. 

Here I will be listing the features and the bug fixes that will go in the next update along with the the folks who logged the bugs/feature requests, those who upvoted it, and those who fixed it by sending PRs.

The idea is to get the folks behind this update to give the update a trial run before I release it.
I always do basic testing before every update. But because I dont work with Go in my day job, I always feel that someone who does Go more than me can help in testing tremendously.

And when I say "test", you can do a minimum pass of just installing the update and then do nothing special other than going about using VS Code like you do everyday. Or you can test the feature you requested/implemented or the bug you logged/fixed. Or you can test all of the below. Every bit helps.

Drop in a comment in this issue after 2 or 3 days of installing the update with your findings.
Also, let me know in the comments what you think of this model of me asking your help in testing an update before releasing it.

Feature Requests:

- #1005 Include stub with exported member name on the comment above it in completions. Use Ctrl+Space to trigger completions inside comments. cc @shreyu86, @atombender, @shurcooL, @ruedap
- #1593 Use code-completion feature from language server. Ensure to use both `go.useLanguageServer` and `go.languageServerExperimentalFeatures` to test this. cc @m90, @leaxoy, @OneOfOne, @primalmotion, @Contextualist
- #1555 Use version 2 apis of delve that enables watching strings and arrays whose size is > 64. First set `useApiV1` to `false` in the debug config to test this. Then configure `dlvLoadConfig` in the same place to control the array size to display. cc @lggomez, @Sirikon, @michaelmello, @jgfet, @Arman92, @carusyte
- #1693 Support to use `golangci-lint` as a linter. cc @golangci, @atombender, @caarlos0, @lafin, @nikitabokovoy
- #1697 Support to use `revive` as a linter. cc @Southclaws, @mgechev, @wolfeidau
- #1302 Make the test coverage colors/styles more colorblind friendly. cc @kentquirk, @dcormier, @sigwinch28, @Mistobaan

Bugs:

- #1667 Issue with formatting when using the language server. cc @OneOfOne 
- #1490 Perf issue with using `gopkgs`. cc @codingconcepts, @uudashr
- #1683 Fix code coverage when code is covered by multiple tests. cc @kentquirk

[...]

PR #1703, opened 2018-06-01
doc(README) Fix the revive flags examples.

wolfeidau:
Road testing revive as a part of #1698 and found some small corrections required.


ramya-rao-a:
Added support to resolve the file path passed in the `-config` flag to revive with https://github.com/Microsoft/vscode-go/commit/d31636a89931add2b799610d91dce1f67b27d5d8

@wolfeidau Can you try using file paths that use `~` for home directory or `${workspaceRoot}` ?

PR #1704, opened 2018-06-01
Fix lint tool cancellation

doxxx:
When a lint tool takes a non-trivial amount of time to execute (e.g. gometalinter), and the user has `lintOnSave` enabled, saving a file twice in quick succession causes the previous lint tool execution to be cancelled.

However, subsequent executions of the lint tool will be [terminated immediately](https://github.com/Microsoft/vscode-go/blob/2075d98808e10bbe475500e2a1752d3790e55251/src/util.ts#L584) because the "global" CancellationTokenSource used for cancelling the lint tool is still in the cancelled state.

This is fixed by recreating the CancellationTokenSource after it has been cancelled.

### DIFF
--- a/src/goLint.ts
+++ b/src/goLint.ts
@@ -45,6 +45,7 @@ export function lintCode(lintWorkspace?: boolean) {
 export function goLint(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfiguration, lintWorkspace?: boolean): Promise<ICheckResult[]> {
 	if (running) {
 		tokenSource.cancel();
+		tokenSource = new vscode.CancellationTokenSource();
 	}
 
 	const currentWorkspace = getWorkspaceFolderPath(fileUri);
--- a/src/goLint.ts
+++ b/src/goLint.ts
@@ -45,6 +45,7 @@ export function lintCode(lintWorkspace?: boolean) {
 export function goLint(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfiguration, lintWorkspace?: boolean): Promise<ICheckResult[]> {
 	if (running) {
 		tokenSource.cancel();
+		tokenSource.dispose();
 		tokenSource = new vscode.CancellationTokenSource();
 	}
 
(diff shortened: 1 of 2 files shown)


src/goLint.ts
```diff
@@ -45,6 +45,7 @@ export function lintCode(lintWorkspace?: boolean) {
 export function goLint(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfiguration, lintWorkspace?: boolean): Promise<ICheckResult[]> {
 	if (running) {
 		tokenSource.cancel();
+		tokenSource = new vscode.CancellationTokenSource();
```
ramya-rao-a: Also, should we dispose the current token before creating a new one?

doxxx:
It's not clear to me whether I need to wrap the cancel in a try/finally so that the dispose will always be called. I don't know how closely vscode's implementation of CancellationTokenSource mirrors that of .NET's.

ramya-rao-a:
I spent a long time looking at how the CancellationTokenSource works and came at the below conclusion:
- Yes, like you figured out, once cancelled, the token remains in a cancelled state and every new `onCancellationRequest` handler will get called
- The token source needs to be disposed after each use. Take the case of a completed lint process followed by a cancelled one. If the token source is not disposed after the first use, then the  `onCancellationRequest` handler will get called even for the completed process

https://github.com/Microsoft/vscode-go/pull/1704/commits/7f8e68df058bd69bf7c64313cf5b26234bd6544e should take care of both cases.

> It's not clear to me whether I need to wrap the cancel in a try/finally so that the dispose will always be called.

Are you talking about the case that one of the cancel event handlers might throw an exception and so the `dispose` won't be called?

doxxx:
> Are you talking about the case that one of the cancel event handlers might throw an exception and so the dispose won't be called?

Yes.

ramya-rao-a:
Just tried throwing an exception from the cancel event handler. It doesn't affect the `dispose`. So we are good.

PR #1707, opened 2018-06-03
Added support for tests using stretchr/testify

mhr3:
This adds support for tests using the widely used [stretchr/testify](https://github.com/stretchr/testify) framework, and allows running and debugging individual test methods with instances which use github.com/stretchr/testify/suite interfaces.

image

### DIFF
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -8,7 +8,7 @@
 import path = require('path');
 import vscode = require('vscode');
 import os = require('os');
-import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions } from './testUtils';
+import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions, extractInstanceTestName, findTestFnForInstanceTest } from './testUtils';
 import { getCoverage } from './goCover';
 
 // lastTestConfig holds a reference to the last executed TestConfig which allows
@@ -59,11 +59,21 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, isBenchmar
 				return;
 			}
 
+			const testConfigFns = [testFunctionName];
+
+			if (!isBenchmark && extractInstanceTestName(testFunctionName)) {
+				// find test function with corresponding suite.Run
+				const t = findTestFnForInstanceTest(testFunctionName, editor.document, testFunctions);
+				if (t) {
+					testConfigFns.push(t.name);
+				}
+			}
+
 			const testConfig: TestConfig = {
 				goConfig: goConfig,
 				dir: path.dirname(editor.document.fileName),
 				flags: testFlags,
-				functions: [testFunctionName],
+				functions: testConfigFns,
 				isBenchmark: isBenchmark,
 			};
 
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -59,13 +59,13 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, isBenchmar
 				return;
 			}
 
-			const testConfigFns = [testFunctionName];
+			let testConfigFns = [testFunctionName];
 
 			if (!isBenchmark && extractInstanceTestName(testFunctionName)) {
 				// find test function with corresponding suite.Run
-				const t = findTestFnForInstanceTest(testFunctionName, editor.document, testFunctions);
-				if (t) {
-					testConfigFns.push(t.name);
+				const testFns = findTestFnForInstanceTest(testFunctionName, editor.document, testFunctions);
+				if (testFns) {
+					testConfigFns = testConfigFns.concat(testFns.map(t => t.name));
 				}
 			}
 
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -8,7 +8,7 @@
 import path = require('path');
 import vscode = require('vscode');
 import os = require('os');
-import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions, isInstanceTestMethod } from './testUtils';
+import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions } from './testUtils';
 import { getCoverage } from './goCover';
 
 // lastTestConfig holds a reference to the last executed TestConfig which allows
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -8,7 +8,7 @@
 import path = require('path');
 import vscode = require('vscode');
 import os = require('os');
-import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions, extractInstanceTestName, findTestFnForInstanceTest } from './testUtils';
+import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions, extractInstanceTestName, findAllTestSuiteRuns } from './testUtils';
 import { getCoverage } from './goCover';
 
 // lastTestConfig holds a reference to the last executed TestConfig which allows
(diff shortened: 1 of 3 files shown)


mhr3:
@ramya-rao-a Could you review please?

ramya-rao-a:
@mhr3 Yes, I'll get to this in a few days.

ramya-rao-a:
Never mind, I used the one from your screenshot above

src/testUtils.ts
```diff
@@ -89,10 +90,36 @@ export function getTestFunctions(doc: vscode.TextDocument, token: vscode.Cancell
...
+
+/**
+ * Extracts test method name of a suite test function.
+ * For example a symbol with name "(*testSuite).TestMethod" will return "TestMethod".
+ *
+ * @param symbol Symbol to extract method name from.
+ */
+export function extractInstanceTestName(symbol: vscode.SymbolInformation | string): string {
+	let symbolName = typeof symbol === 'string' ? symbol : symbol.name;
+	const match = symbolName.match(/\.(Test.*)$/);
```
ramya-rao-a: Why not re-use the `testSuiteMethodRe` regex here? We can do that if it is updated to `/^\([^)]+\)\.(Test.*)$/`.

src/goTest.ts
```diff
@@ -8,7 +8,7 @@
 import path = require('path');
 import vscode = require('vscode');
 import os = require('os');
-import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions } from './testUtils';
+import { goTest, TestConfig, getTestFlags, getTestFunctions, getBenchmarkFunctions, isInstanceTestMethod } from './testUtils';
```
ramya-rao-a: `isInstanceTestMethod` is an unused import here

src/testUtils.ts
```diff
@@ -89,10 +90,36 @@ export function getTestFunctions(doc: vscode.TextDocument, token: vscode.Cancell
...
+			)
 		);
 }
 
+/**
+ * Checks whether given symbol is a test method of a test suite instance.
+ *
+ * @param symbol Symbol to check.
+ */
+export function isInstanceTestMethod(symbol: vscode.SymbolInformation | string): boolean {
```
ramya-rao-a: Since this function isn't used anywhere, it need not be exported.

src/testUtils.ts
```diff
@@ -89,10 +90,36 @@ export function getTestFunctions(doc: vscode.TextDocument, token: vscode.Cancell
...
+	return testSuiteMethodRe.test(symbolName);
+}
+
+/**
+ * Extracts test method name of a suite test function.
+ * For example a symbol with name "(*testSuite).TestMethod" will return "TestMethod".
+ *
+ * @param symbol Symbol to extract method name from.
+ */
+export function extractInstanceTestName(symbol: vscode.SymbolInformation | string): string {
```
ramya-rao-a: All uses of this function passes in a string. So we don't need to support both string and `vscode.SymbolInformation`

src/testUtils.ts
```diff
@@ -242,7 +269,26 @@ function expandFilePathInOutput(output: string, cwd: string): string {
...
+			let testFunctions = testconfig.functions;
+			let testifyMethods = testFunctions.filter(isInstanceTestMethod);
+			if (testifyMethods.length > 0) {
+				// filter out testify methods
+				testFunctions = testFunctions.filter(fn => !isInstanceTestMethod(fn));
+				testifyMethods = testifyMethods.map(extractInstanceTestName);
+			}
+
+			if (testFunctions.length > 0) {
+				params = params.concat(['-run', util.format('^%s$', testFunctions.join('|'))]);
```
ramya-rao-a: We have to pass `-run ^$` even if `test.Functions` array is empty. Otherwise all tests will be run

mhr3:
Thanks for the review, and the refactor ;)

> We have to pass -run ^$ even if test.Functions array is empty. Otherwise all tests will be run

That was intended, with `-run ^$` none of the tests will run - including the ones that invoke `suite.Run(...)`. Ideally we'd analyze the code deeper and match which Test function contains the necessary `Run` call, but that sounded like an overkill to me, you probably aren't mixing suite and non-suite tests in a single package - and even if, we'd just end up running more tests than necessary, which shouldn't be a bid deal.

ramya-rao-a:
> you probably aren't mixing suite and non-suite tests in a single package - and even if, we'd just end up running more tests than necessary, which shouldn't be a big deal.

Take the case where the non testify test is failing and the user clicks on the "run test" codelens above a test which should be run using testify. Now the non-testify test will run and fail and the failure will be printed out in the output which can be misleading

mhr3:
Yes, I get what you're saying, but if we ran with `-run ^$ -testify.m ^TestFoo$`, nothing would actually run, cause testify relies on something invoking the `suite.Run()`.

ramya-rao-a:
>  but if we ran with -run ^$ -testify.m ^TestFoo$, nothing would actually run

Ah! I missed that. I understand it now.

Regardless, having all the normal tests run when the `run test` or `debug test` codelens are clicked or when the command `Go:  Test Function At Cursor` is run, feels very wrong.

How about this?

When creating the codelens as well as in the `testAtCursor` function, we have access to all the test functions with their range. 

If the current function is actually a method, then scan the text content of the other functions to find the one with `suite.Run()` call. Both names should get included in the `TestConfig` object in the `testAtCursor` function

This way, in `targetArgs`, you will have access to both the function and the method.

mhr3:
Pushed the changes you requested, have another look please

src/testUtils.ts
```diff
@@ -89,10 +90,64 @@ export function getTestFunctions(doc: vscode.TextDocument, token: vscode.Cancell
...
+	// we have multiple matches, filter even more
+	const match = symbolName.match(testSuiteMethodRegex);
+	if (!match || match.length !== 3) {
+		return candidates[0];
+	}
+	// drop pointers, lowercase for more lenient matching
+	const instanceType = match[1].replace(/[*]/g, '').toLowerCase();
+	const filtered = candidates.filter(t => {
+		const text = doc.getText(t.location.range).toLowerCase();
+		if (text.includes(instanceType)) {
```
ramya-rao-a: This may give false positives if say there is a comment inside the function with another suite's name.
It can give false negatives if the suite was instantiated outside the function.

Why not return all the functions that have a `suite.Run()`?

Since we also pass `-testify.m`, the other suites will be a no-op right?

src/testUtils.ts
```diff
@@ -89,10 +90,64 @@ export function getTestFunctions(doc: vscode.TextDocument, token: vscode.Cancell
...
+ * @param doc Editor document
+ * @param allTests All test functions
+ */
+export function findTestFnForInstanceTest(symbolName: string, doc: vscode.TextDocument, allTests: vscode.SymbolInformation[]): vscode.SymbolInformation {
+	// get non-instance test functions
+	const testFunctions = allTests.filter(t => !testSuiteMethodRegex.test(t.name));
+	// filter further to ones containing suite.Run()
+	const candidates = testFunctions.filter(t => {
+		const text = doc.getText(t.location.range);
+		return /suite\.Run\(/.test(text);
```
ramya-rao-a: Since this is a simple string check, wont `text.includes('suite.Run(')` be simpler?
Unless we make this a useful regex like `/suite\.Run\([^\)]+\)/

src/testUtils.ts
mhr3: Changed, clearly this isn't going to work in 100% of cases, but should be close enough.

mhr3:
@ramya-rao-a Anything still missing here?

ramya-rao-a:
@mh3 I dont see any changes related to my last comment https://github.com/Microsoft/vscode-go/pull/1707#discussion_r194812979

You have replied with "Changed, clearly this isn't going to work in 100% of cases, but should be close enough.", but I dont see which commit corresponds to this comment

mhr3:
Ok, refactored per your comment.

ramya-rao-a:
Thanks!
I'll release an update in a few days.

alex-kovoy:
how can i disable now this new default behavior? You cannot run any tests now without this required framework

ramya-rao-a:
@alex-kovoy That is not true. Tests not using the testify framework run just as before. If they don't for you, please log a new issue with a sample code we can try replicate your issue on.

alex-kovoy:
It looks like this implementation is based on string parsing (regex) in order to determine if this framework is installed which I am not sure is the right way to implement this feature. Removing this line https://github.com/Microsoft/vscode-go/pull/1707/files#diff-869d156f8d53e34e379625eb2aab2d2cR292 fixed my problem.

ramya-rao-a:
For anyone else seeing the same issue as @alex-kovoy, the issue is being tracked in #1911

Issue #1645, opened 2018-04-25
gocode is broken on current go git master


OneOfOne:
Steps to Reproduce:

1. use current go git master.
2. get PANIC when you try to auto complete.

Possible fixes:

- https://github.com/mdempsky/gocode, slightly slower than nsf/gocode but NOT that slow, not noticeable really.
- https://github.com/OneOfOne/gocode my fork based on mdempsky's +  myitcv's vgo PR.

Just installing either of those will override nsf's gocode and you just can run `killall -9 gocode` without restarting vscode.

OneOfOne:
With https://github.com/Microsoft/vscode-go/pull/1607, go-langserver will handle that.

- https://github.com/sourcegraph/go-langserver/issues/268

@slimsag @keegancsmith

FiloSottile:
From https://github.com/nsf/gocode/blob/9d1e0378d/README.md:

> IMPORTANT: consider switching to https://github.com/mdempsky/gocode if you have problems starting with Go version 1.10, due to changes in binary packages architecture (introduction of package cache) I'm not going to adjust gocode for it for quite some time. There is a higher chance that fork under the given link will have some solution to the problem sooner or later.

mccolljr:
Out of curiousity, has this been addressed? Is this package going to move to `github.com/mdempsky/gocode`? Manually installing and switching to that fork of `gocode` stops the `PANIC` messages and gets me completion again. Even `gocode`'s original author recommends switching to the fork, so I think this would be the right move. Or, at least provide a setting for which fork to use?

localvar:
just FYI: To improve the performance of [https://github.com/mdempsky/gocode](https://github.com/mdempsky/gocode), I create another temporary folk at: [https://github.com/localvar/gocode](https://github.com/localvar/gocode)

ramya-rao-a:
Well, I did give this a try, but the completion tests for all versions of Go started failing in travis.
See https://github.com/Microsoft/vscode-go/commits/new-gocode

If anyone can figure that out, then yes, we can move to `github.com/mdempsky/gocode`

ramya-rao-a:
Fixed with https://github.com/Microsoft/vscode-go/pull/1710 thanks to @m90
The fix is out in today's update to the Go extension (0.6.82)

PR #1710, opened 2018-06-04
Migrate from nsf/gocode to mdempsky/gocode for completion results

m90:
As package nsf/gocode is unmaintaned this is now consuming mdempsky/gocode.
In order to cater for changes, correctly handle empty results and promise
rejections.

This would solve #1645 although for reasons unclear to me, the updated `gocode` package will not provide any results for members of unimported packages (hence the skipped test case). Don't know if this is a dealbreaker or can be fixed deeper down in `goSuggest.ts`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
   -  nvm install $TRAVIS_NODE_VERSION;
   - npm install
   - npm run vscode:prepublish
-  - go get -u -v github.com/nsf/gocode
+  - go get -u -v github.com/mdempsky/gocode
   - go get -u -v github.com/rogpeppe/godef
   - if [[ "$(go version)" =~ "go version go1.5" ]]; then echo hello; else go get -u -v github.com/zmb3/gogetdoc; fi
   - if [[ "$(go version)" =~ "go version go1.5" ]]; then echo cannot get golint; else go get -u -v github.com/golang/lint/golint; fi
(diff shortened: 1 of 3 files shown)


m90:
Scrap the comment about the test that needs to be skipped, that was an issue I encountered locally as `math` can actually be an ambiguous package (there is `github.com/gohugoio/tpl/math`) and the extension therefore decides to not display anything here: https://github.com/Microsoft/vscode-go/blob/5bc4ce40557a044cd790b9941271908b9135bc68/src/goSuggest.ts#L361-L363

m90:
@ramya-rao-a Should we also already update the parts of the wiki ([this](https://github.com/Microsoft/vscode-go/wiki/Go-tools-that-the-Go-extension-depends-on) and [this](https://github.com/Microsoft/vscode-go/wiki/Go-with-VS-Code-FAQ-and-Troubleshooting)) that still point to the `nsf` gocode or should that wait until it has actually been released?

ramya-rao-a:
@m90  Let's wait until the release for the wiki update.

I was wondering if we should prompt the users using Go 1.10 and above to update their `gocode` if the current one panics. Because otherwise, this change will only help first time installs and those who already have `go get` the new gocode

m90:
From what I understand the `mdempsky` gocode should work fine on older supported Go versions as well, so ideally a prompt like this would show up for everyone.

Unfortunately I have a hard time coming up with an idea on how to know which `gocode` is installed on a user's machine as the binary name is the same and it will also be installed into `$GOPATH/bin`. It looks like the `nsf` gocode prints slightly different output on `gocode -h`, but I'm not too sure if this is the most robust way of doing it.

ramya-rao-a:
Ok, In that case, we can leave it as is.
I can update the FAQ to include a note on this.

PR #1724, opened 2018-06-08
Add x icon to 'Cancel Running Tests

uudashr:
I think it just nice to have the x icon, attract more on user attention

Before
screen shot 2018-06-08 at 13 23 46

After
screen shot 2018-06-08 at 13 24 18

### DIFF
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -16,7 +16,7 @@ const sendSignal = 'SIGKILL';
 const outputChannel = vscode.window.createOutputChannel('Go Tests');
 const statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
 statusBarItem.command = 'go.test.cancel';
-statusBarItem.text = 'Cancel Running Tests';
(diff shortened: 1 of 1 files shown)


src/testUtils.ts
```diff
@@ -16,7 +16,7 @@ const sendSignal = 'SIGKILL';
 const outputChannel = vscode.window.createOutputChannel('Go Tests');
 const statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
 statusBarItem.command = 'go.test.cancel';
-statusBarItem.text = 'Cancel Running Tests';
+statusBarItem.text = '$(x) Cancel Running Tests';
```
ramya-rao-a: How does `${x}` relate the icon and how did you find it :) ?

uudashr: Here https://code.visualstudio.com/docs/extensions/example-word-count#_customizing-the-status-bar

We can use https://octicons.github.com/ on vscode

ramya-rao-a: Nice!

PR #1743, opened 2018-06-19
go.gopath and go.toolsGopath now expand ${env:XXX}

### LINKED ISSUE #1732: go.toolsGopath doesn't expand ${env:GOPATH}
go.toolsGopath doesn't expand ${env:GOPATH}
I'm using ms-vscode.go v0.6.82

reproduce step

1. create .vscode/setting.json

{
    "go.gopath": "${env:GOPATH};${workspaceFolder}",
    "go.toolsGopath": "${env:GOPATH}"
}

2. press ctrl + shift + p, select Go: Install/Update Tools

sphawk:
Fixes #1732 
but... I solved my problem like this.
"go.toolsGopath" : '~/go'
....

anyway, I'll create pull request for someone who wants expand ${env:XXX}


src/util.ts
```diff
@@ -368,6 +368,12 @@ export function getToolsEnvVars(): any {
 	return envVars;
 }
 
+export function substituteEnv(input: string): string {
+	return input.replace(/\${env:([^}]+)}/mg, function (match, capture) {
```
ramya-rao-a: I believe we dont need the `m` flag here. Any reason we should have it?

sphawk: I can't assume input string is single line. someone could use that function elsewhere.

ramya-rao-a: Thats true, but isnt the `m` flag make sense only if you are using either `^` or `$` in the regular expression?

sphawk: sorry for late response, I'll remove `m` flag and add some tests within few hours

ramya-rao-a:
Thanks @sphawk!

Issue #1733, opened 2018-06-15
Feature request: Add import folder to workspace


freman:
I'm not sure how feasible this one is but it's something I'm constantly doing.

It'd be super cool if you could select an import, right click it, and add folder to workspace.

ramya-rao-a:
PRs are welcome for this feature

References:
- The `Go: Get Package` is an example of a command that acts on the import statement.
-  `vscode.workspace.updateWorkspaceFolders` can help in adding folder to workspace

freman:
So, I'll be the first to admin, I've never used typescript before and my abuse of it is probably a reflection but I have a patch https://github.com/Microsoft/vscode-go/pull/1745 that is a start in a direction. It's not vendor aware but realistically your vendors are already in your workspace.

Happy for tips to further develop it?

ramya-rao-a:
This feature is now out in the latest update (0.6.85) to the Go extension.

PR #1745, opened 2018-06-20
Add Import to Workspace

freman:
Make it possible to add the path for the selected import, or part there of to the current workspace for convenience.

Resolves https://github.com/Microsoft/vscode-go/issues/1733


ramya-rao-a:
Thanks for the PR @freman 

Any reason to prompt the user to choose among the imported packages in current document rather than say choose the package under the cursor like we do for the `Go: Get Package` command?

freman:
Hi @ramya-rao-a no reason really, I was just excited to be getting a feature that worked around the macos path selection box.

It did have the benefit of working no matter where one was in the file.

I've taken your suggestion and expanded on it, it'll now use the current selected import line, or if the user has selected text it'll try to use that.

It'll also try to prevent adding redundant workspaces, one future improvement I could see is removing redundant workspaces (ie you add foo/bar/package but then later add foo/bar it might be worth removing foo/bar/package automatically)

freman:
Errm, I don't think I broke travis... did I?

src/goImport.ts
```diff
@@ -104,3 +107,52 @@ export function addImport(arg: string) {
...
 	});
 }
+
+export function addImportToWorkspace(arg: string) {
+	const editor = vscode.window.activeTextEditor;
+	const selection = editor.selection;
+
+	let importPath = '';
+	if (selection.isSingleLine) {
+		// Attempt to load a partial import path based on currently selected text
```
ramya-rao-a: Can you provide an example for the case that you are attempting to cover here?

src/goImport.ts
```diff
@@ -104,3 +107,52 @@ export function addImport(arg: string) {
...
+	}
+
+	if (importPath === '') {
+		vscode.window.showErrorMessage('No import path to add');
+		return;
+	}
+
+	let filePath = fixDriveCasingInWindows(editor.document.fileName);
+	let fileDirPath = path.dirname(filePath);
+	let currentWorkspace = getCurrentGoWorkspaceFromGOPATH(getCurrentGoPath(), fileDirPath);
```
ramya-rao-a: Take the case of multiple gopaths. Someone might have GOPATH set to `/Users/folderA;/Users/folderB`.
The current workspace is under /Users.folderA
The package in question could be under /Users/folderB
In this case, the current logic wont work.

How about passing the import path to `go list -f {{.Dir}}`? This would return the actual path for the package. You wouldnt need to to a `fs.exists` check either.

See `showPackageFiles` function in `goBrowsePackage.ts` for a reference on how to use `go list`.

src/goImport.ts
```diff
@@ -104,3 +107,52 @@ export function addImport(arg: string) {
...
+	if (!fs.existsSync(globalPackagePath)) {
+		vscode.window.showErrorMessage('Import path not found on disk');
+		return;
+	}
+
+	const importPathUri = vscode.Uri.file(globalPackagePath);
+	const existingWorkspaceFolder = getRootWorkspaceFolder(importPathUri);
+	if (existingWorkspaceFolder !== undefined) {
+		vscode.window.showInformationMessage('Already available under ' + existingWorkspaceFolder.name);
+		return;
```
ramya-rao-a: What happens when we pass in a uri for an already existing folder to `vscode.workspace.updateWorkspaceFolders`? If `updateWorkspaceFolders` gracefully handles this case, then we dont need to do this check beforehand

src/goMain.ts
```diff
@@ -274,6 +274,10 @@ export function activate(ctx: vscode.ExtensionContext): void {
 		return addImport(typeof arg === 'string' ? arg : null);
 	}));
 
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.import.workspace', (arg: string) => {
```
ramya-rao-a: What is the scenario where an arg would be passed via the command?

package.json
```diff
@@ -154,6 +154,10 @@
         "title": "Go: Add Import",
         "description": "Add an import declaration"
       },
+      {
+        "command": "go.import.workspace",
+        "title": "Go: Add import to workspace"
```
ramya-rao-a: The name is not very self explanatory....
How about `go.add.package.workspace` and `Go: Add Package to Workspace` ?

src/goImport.ts
freman: Absolutely. The easiest example that comes to mind.

`import "github.com/aws/aws-sdk-go/service/rds/rdsutils"`

Realistically being able to select `github.com/aws/aws-sdk-go` and add that to the workspace is probably going to be what you want more than `github.com/aws/aws-sdk-go/service/rds/rdsutils`

src/goImport.ts
freman: True, I'll sort this out.

src/goImport.ts
freman: I can't remember if it did or didn't, I think my motivation was to let the user know why nothing happened. I can double check but I still think it's nice to let the user know that the workspace is already imported.

IE: they try to add `github.com/aws/aws-sdk-go/service/rds/rdsutils` after having `github.com/aws/aws-sdk-go` loaded.

In an ideal world I'd actually make the workspace navigate/expand the path but I've no idea how to do this in vscode

src/goMain.ts
freman: Uh that's a bit of left over code, it probably wouldn't at this point, I'll remove that.

src/goImport.ts
freman: if (existingWorkspaceFolder !== undefined) {
+		vscode.window.showInformationMessage('Already available under ' + existingWorkspaceFolder.name);
+		return;

freman:
@ramya-rao-a is there any way to trigger travis without pushing again?

src/goImport.ts
ramya-rao-a: Good point. I would change the `if (selection.isSingleLine)` to `if (!selection.isEmpty)`.
There could be a package with a very long import path or someone with a very zoomed in window.

src/goImport.ts
```diff
@@ -104,3 +105,58 @@ export function addImport(arg: string) {
...
+		importPath = getImportPath(selectedText);
+	}
+
+	if (importPath === '') {
+		vscode.window.showErrorMessage('No import path to add');
+		return;
+	}
+
+	const goRuntimePath = getBinPath('go');
+	const env = Object.assign({}, process.env, { GOPATH: getCurrentGoPath() });
```
ramya-rao-a: Use `getToolsEnvVars()`

src/util.ts
```diff
@@ -789,4 +789,15 @@ export function killTree(processId: number): void {
...
 		}
 	}
-}
\ No newline at end of file
+}
+
+export function getRootWorkspaceFolder(forPathUri: vscode.Uri): vscode.WorkspaceFolder {
+	let workspaceFolder = vscode.workspace.getWorkspaceFolder(forPathUri);
+	let previousWorkspaceFolder: vscode.WorkspaceFolder;
+	while (workspaceFolder !== undefined) {
```
ramya-rao-a: Can you explain the scenario that would need this loop?
Taking your example of aws, if `github.com/aws/aws-sdk-go` is already part of the workspace and we run the command for `github.com/aws/aws-sdk-go/service/rds/rdsutils`, wouldnt `vscode.workspace.getWorkspaceFolder(forPathUri)` give you the uri of the parent workspace folder? And isnt that enough?

PR #1766, opened 2018-07-03
Increase overall performance by speeding up GoDocumentSymbolProvider

### LINKED ISSUE #1668: formatOnSave extremely slow at piping
`formatOnSave` is being extremely slow, and failing if `editor.formatOnSaveTimeout` is not set to north of 5000ms on moderately large (8000 lines) files.

```
[Extension Host] Formatting took too long(5647ms). Format On Save feature could be aborted.
```

**The issue is not in the formatting tool, but in how the input is sent to it.**

I replaced `goimports` (my `go.formatTool`) with this script to compare invocations from VS Code:

```
#! /bin/bash

echo "$@"
echo "$PWD"
date
wc
date
```

When run within VS Code:

```
Mon May  7 20:13:27 EDT 2018
    2879   12052   90302
Mon May  7 20:13:31 EDT 2018
```

When run with the exact same input, invocation and environment from a shell with `pbpaste`:

```
Mon May  7 20:14:04 EDT 2018
    2879   12052   90302
Mon May  7 20:14:04 EDT 2018
```

You can see that it's taking 4 seconds just to pipe the input to `wc`, between the two `date` calls.

Reporting it here instead of to VS Code because in https://github.com/Microsoft/vscode-go/issues/1419#issuecomment-355654879 it says this extension implemented its own version of `formatOnSave`.

I have the global `editor.formatOnSave` on, in case it matters. Running on macOS.

JacksonKearl:
This removes massive copy overhead from recreating Buffers of the entire document for every symbol in the file. Also reduces the quadratic time approach to converting from byte-offsets to character-offsets to log-linear time.

Hopefully closes #1668.

### DIFF
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -109,7 +109,7 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
 		decls: GoOutlineDeclaration[],
 		symbols: vscode.SymbolInformation[],
 		containerName: string,
-		byteOffsetToDocumentOffset: (offset: number) => number): void {
+		byteOffsetToDocumentOffset: (byteOffset: number) => number): void {
 
 		let gotoSymbolConfig = vscode.workspace.getConfiguration('go', document.uri)['gotoSymbol'];
 		let includeImports = gotoSymbolConfig ? gotoSymbolConfig['includeImports'] : false;
@@ -137,7 +137,6 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
 	}
 
 	public provideDocumentSymbols(document: vscode.TextDocument, token: vscode.CancellationToken): Thenable<vscode.SymbolInformation[]> {
-		let t0 = Date.now();
 		let options = { fileName: document.fileName, document: document };
 		return documentSymbols(options, token).then(decls => {
 			let symbols: vscode.SymbolInformation[] = [];
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -7,7 +7,7 @@
 
 import vscode = require('vscode');
 import cp = require('child_process');
-import { getBinPath, getFileArchive, getToolsEnvVars, killProcess, makeMemoizedOffsetConverter } from './util';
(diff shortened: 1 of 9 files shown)


OneOfOne:
I have been using this for over a week, didn't run into any issues.

JacksonKearl:
@OneOfOne were you experiencing perf issues before?

JacksonKearl:
@FiloSeattle can you try this vsix and see if the performance improves?

src/goOutline.ts
```diff
@@ -104,7 +104,13 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
 		'function': vscode.SymbolKind.Function
 	};
 
-	private convertToCodeSymbols(document: vscode.TextDocument, decls: GoOutlineDeclaration[], symbols: vscode.SymbolInformation[], containerName: string): void {
+	private convertToCodeSymbols(
+		document: vscode.TextDocument,
+		decls: GoOutlineDeclaration[],
+		symbols: vscode.SymbolInformation[],
+		containerName: string,
+		byteOffsetToDocumentOffset: (offset: number) => number): void {
```
ramya-rao-a: rename the parameter from `offset` to `byteoffset` here and every other place we are talking about converting byte offset to offset

src/goOutline.ts
```diff
@@ -126,16 +131,17 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
...
 			symbols.push(symbolInfo);
 			if (decl.children) {
-				this.convertToCodeSymbols(document, decl.children, symbols, decl.label);
+				this.convertToCodeSymbols(document, decl.children, symbols, decl.label, byteOffsetToDocumentOffset);
 			}
 		});
 	}
 
 	public provideDocumentSymbols(document: vscode.TextDocument, token: vscode.CancellationToken): Thenable<vscode.SymbolInformation[]> {
+		let t0 = Date.now();
```
ramya-rao-a: unused variable

ramya-rao-a:
This will definitely help for VS Code 1.25 and above as the outline view in the explorer now opens up by default. Not sure if  it helps #1668 when the outline view in the explorer is disabled.

PR #1782, opened 2018-07-10
Prioritize current project import suggestions

marwan-at-work:
This PR will make it so that "import suggestions" will always put your project's packages at the top. This really helps when there are very common package names like `log` or `errors` and you want to import them through the autocomplete feature. Before this pr, you'd have to scroll through 10s if not 100s of packages to find the one form you're workspace. 

From the screenshot below, you can see that if you're inside `upspin.io`'s sourcode, then the autocomplete shows you imports from that package first.
screen shot 2018-07-09 at 11 11 48 pm

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -355,7 +355,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 	// Return importable packages that match given word as Completion Items
 	private getMatchingPackages(document: vscode.TextDocument, word: string, suggestionSet: Set<string>): vscode.CompletionItem[] {
 		if (!word) return [];
-	
+
 		const cwd = path.dirname(document.fileName);
 		const goWorkSpace = getCurrentGoWorkspaceFromGOPATH(getCurrentGoPath(), cwd);
 		const workSpaceFolder = vscode.workspace.getWorkspaceFolder(document.uri);
@@ -375,7 +375,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 					arguments: [pkgPath]
 				};
 				item.kind = vscode.CompletionItemKind.Module;
-				
+
 				// Unimported packages should appear after the suggestions from gocode
 				const isStandardPackage = !item.detail.includes('.');
 				item.sortText = isStandardPackage ? 'za' : pkgPath.startsWith(currentPkgRootPath) ? 'zb' : 'zc';
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -266,7 +266,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 					}
 
 					// Add importable packages matching currentword to suggestions
-					let importablePkgs = includeUnimportedPkgs ? this.getMatchingPackages(currentWord, suggestionSet) : [];
+					let importablePkgs = includeUnimportedPkgs ? this.getMatchingPackages(document, currentWord, suggestionSet) : [];
 					suggestions = suggestions.concat(importablePkgs);
 
 					// 'Smart Snippet' for package clause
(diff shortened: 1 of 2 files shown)


src/goSuggest.ts
```diff
@@ -346,6 +346,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 	private getMatchingPackages(word: string, suggestionSet: Set<string>): vscode.CompletionItem[] {
 		if (!word) return [];
 		let completionItems = [];
+		const rootPath = vscode.workspace.rootPath.slice(getCurrentGoPath().length + 5);
```
ramya-rao-a: This will fail in a few cases
- Multi-root mode. To fix this you can use `vscode.workspace.getWorkspaceFolder(fileUri)`
- The file being edited is not part of the folder opened in VS Code or the file alone has been opened in VS Code. No folders. In this case `vscode.workspace.getWorkspaceFolder(fileUri)` would not return a workspace, so you can fall back to the parent folder of the file.
- The file being edited is not part of the gopath
- The `getCurrentGopath` returns multiple gopaths like `projectA:projectB` or `projectA;projectB`. To solve this you can use `getCurrentGoWorkspaceFromGOPATH` method from `goPath.ts`

src/goSuggest.ts
```diff
@@ -363,9 +364,15 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				// Add same sortText to the unimported packages so that they appear after the suggestions from gocode
 				const isStandardPackage = !item.detail.includes('.');
 				item.sortText = isStandardPackage ? 'za' : 'zb';
-				completionItems.push(item);
+				if (pkgPath.startsWith(rootPath)) {
+					item.sortText = 'a';
```
ramya-rao-a: Why change the sortText?

marwan-at-work: @ramya-rao-a this is so that the unimported packages from the same project (or workspace) are prioritized over the rest of the package suggestions. Say you are in project `a` and you start typing `b` in main.go, when the imports are suggested i'd like to see `a/b` before `someOtherProject/b` so i can just hit enter and `a/b` is imported. Most of the time when you wanna import a package, it's most likely either from your own project, or from the standard library, or a third party. Third party libraries have special cool names so they usually come first anyway. But when you type generic package name like `log` or `errors`, you get too many suggestions from GOPATH and you most likely just want the `log` package from your own project.

marwan-at-work:
@ramya-rao-a i updated and commented the pr, thanks for taking a look!

src/goSuggest.ts
ramya-rao-a: Take a hypothetical case 
- lab, lag, large, lamb are standard pkgs
- github.com/mine/log, github.com/you/log are custom packages

The user is currently working on the github.com/mine/myproject and types `l`
The suggestions would now appear as
- log -> github.com/mine/log
- lab
- lag
- large
- lamb
- log -> github.com/you/log

This would be counter intutive, as all the std pkgs would be expected to be sorted above alphabetically.

The solution to your problem should be such that your package is sorted to the top but only among other packages with the same name. So taking the same example as above, typing `l` should result in

- lab
- lag
- large
- lamb
- log -> github.com/mine/log
- log -> github.com/you/log

ramya-rao-a: My previous comment assumed that the unimported package suggestions were sorted among themselves alphabetically. Which is not the case. My bad.

How about `item.sortText = isStandardPackage ? 'za' : pkgPath.startsWith(rootPath) ? 'zb' : 'zc'`?

We need the `z` as we dont want any of the unimported packages to appear before other normal symbol suggestions.

marwan-at-work:
@ramya-rao-a I've applied your last suggested edit, which seems to be working as intended. Thanks again for pointing out all the edge cases! Let me know if this needs further edits.

marwan-at-work:
@ramya-rao-a ok to merge on my end. 

Something kind of related but probably for another pr: would it be possible to just remove all packages nested under `GOPATH/src/mod` from suggestions and imports? This is because of vgo (and soon Go 1.11) new module system. We're never gonna import anything directly from there. 

I'm happy to push another PR for it if you can give me some direction. Thanks!

ramya-rao-a:
There is an ongoing discussion for vgo in https://github.com/Microsoft/vscode-go/issues/1532. Add you note there with some links to documentation and we can take it from there.

ramya-rao-a:
This fix is now out in the latest update (0.6.85) to the Go extension.

Issue #1779, opened 2018-07-09
Make autocomplete for inline functions optional


kaskavalci:
https://github.com/Microsoft/vscode-go/issues/1287 addresses the issue on autocomplete for inline functions. However when passing those functions around, autocomplete can be very annoying. Can we make this feature optional?

![](http://g.recordit.co/8jfCMnGTXJ.gif)

ramya-rao-a:
We could skip using the snippet with function parameters when the cursor is followed by a comma or `)`. 
It wouldnt cover 100% of the cases, but should work for most common cases.

The reason we cannot do a correct fix for all cases is that we are only aware that the symbol being completed is a variable of func type. We are not aware if the cursor is inside another function call.

If that sounds good enough, then PRs are welcome.

The fix would go here: https://github.com/doxxx/vscode-go/blob/ba71beed263765892fd6b49d0136954785b47ba8/src/goSuggest.ts#L171

kaskavalci:
Thanks for the response. I'll submit a PR this weekend.

doxxx:
#1788 only appears to have fixed this when completing variables of type `func`. When completing the name of a standalone function, it completes with the parameter list.

For example:

```go
func HandleRequest(w http.ResponseWriter, r *http.Request) {
}

func Register(r *mux.Router, jm *job.Manager) {
	r = r.PathPrefix("/api/v1").Subrouter()
	r.HandleFunc("/", HandleReq***CURSOR***)
}
```

When I invoke completion at the marked cursor position, it completes to:

```go
func HandleRequest(w http.ResponseWriter, r *http.Request) {
}

func Register(r *mux.Router, jm *job.Manager) {
	r = r.PathPrefix("/api/v1").Subrouter()
	r.HandleFunc("/", HandleRequest(w, r))
}
```

However, if I start with this instead:

```go
func Register(r *mux.Router, jm *job.Manager) {
	r = r.PathPrefix("/api/v1").Subrouter()
	handleRequest := func(w http.ResponseWriter, r *http.Request) {}
	r.HandleFunc("/", handleReq***CURSOR***)
}
```

Completion results in the following:

```go
func Register(r *mux.Router, jm *job.Manager) {
	r = r.PathPrefix("/api/v1").Subrouter()
	handleRequest := func(w http.ResponseWriter, r *http.Request) {}
	r.HandleFunc("/", handleRequest)
}
```

ramya-rao-a:
[...]

PR #1788, opened 2018-07-15
Do not suggest function parameters if passing it as an argument

kaskavalci:
As per the suggestion: https://github.com/Microsoft/vscode-go/issues/1779#issuecomment-403675528

This PR disables suggesting function parameters if passing the suggestion as an argument. If cursor ends with `,` or `)` it skips parameter suggestion.

For more info: https://github.com/Microsoft/vscode-go/issues/1779


kaskavalci:
True, I missed that.

ramya-rao-a:
Never mind, I've added the tests. Thanks for the PR!

PR #1790, opened 2018-07-16
Debugger compares length of bytes instead of length of string for hov…

### LINKED ISSUE #1777: debug hover string variable ...more appears
```go
	s1 := "你好"
	s2 := "abc"
	fmt.Println(s1)
	fmt.Println(s2)
```
image

hover on s1, I get message "你好...+4 more"

“你好” is expected.

xmattstrongx:
Fixes #1777

image

image


src/debugAdapter/goDebug.ts
```diff
@@ -819,8 +819,8 @@ class GoDebugSession extends DebugSession {
 			};
 		} else if (v.kind === GoReflectKind.String) {
 			let val = v.value;
-			if (v.value && v.value.length < v.len) {
-				val += `...+${v.len - v.value.length} more`;
+			if (v.value && Buffer.byteLength(v.value) < v.len) {
```
ramya-rao-a: We can avoid calling Buffer.byteLength twice by pulling it out say like `let byteLength = Buffer.byteLength(val || '');`

PR #1792, opened 2018-07-17
Detach to clean up debug file

ramya-rao-a:
This is an attempt to clean up the `debug` file that gets created after every debugging session.

@lggomez The Detach command deletes the said `debug` file successfully, but there are a few issues where I thought you could help me out
- It only works in Windows. On Mac, the debug file doesnt get removed
- The callback for the detach call never gets called, so I had to resort to `settimeout`

Thoughts?

@aarzilli  @derekparker What is the recommended way for a client to stop an headless dlv server such that
- the debug file that gets created is taken care of
- all processes started by the debug server (for eg: web servers) are killed (We had to manually kill the process corresponding to the headless server and all its children to achieve this)

From the docs I can see `Detach`, `Disconnect` (only available in v2) and `Kill` as possible choices.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -587,7 +587,7 @@ class GoDebugSession extends DebugSession {
 	protected disconnectRequest(response: DebugProtocol.DisconnectResponse, args: DebugProtocol.DisconnectArguments): void {
 		verbose('DisconnectRequest');
 		this.delve.close();
-		
+
 		// Timeout to ensure detach is complete.
 		setTimeout(() => {
 			if (this.delve.debugProcess) {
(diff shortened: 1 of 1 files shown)


aarzilli:
Detach is the intended way, if it doesn't remove the file on macOS it's a bug.

ramya-rao-a:
@aarzilli Is Detach expected to close all the processes that might have been started by the headless server as well? For example: When debugging a program that provides a web server, the web server continues to live in Mac but not Windows once the debugging session is complete.

aarzilli:
If you are passing Kill=true then yes.

ramya-rao-a:
Thanks @aarzilli

PR #1793, opened 2018-07-17
Clean up a message displayed to users.

UserRandom:


src/goInstallTools.ts
```diff
@@ -350,7 +350,7 @@ function installTools(missing: string[]) {
 			if (missing.indexOf('go-langserver') > -1) {
 				outputChannel.appendLine('Reload VS Code window to use the Go language server');
 			}
-			outputChannel.appendLine('All tools successfully installed. You\'re ready to Go :).');
+			outputChannel.appendLine('All tools successfully installed. You\'re ready to Go.');
```
ramya-rao-a: @UserRandom This was an intentional addition pointing to the pun of using the word  "go" in the context of the Go language itself. We would like to keep it. You are most welcome to contribute via bug fixes or other documentation updates.

PR #1794, opened 2018-07-17
Add single process management for format and build

JacksonKearl:


src/goFormat.ts
```diff
@@ -34,7 +34,7 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
 			formatFlags.push('-style=indent=' + options.tabSize);
 		}
 
-		return this.runFormatter(formatTool, formatFlags, document).then(edits => edits, err => {
+		return this.runFormatter(formatTool, formatFlags, document, token).then(edits => edits, err => {
```
ramya-rao-a: Have you found a scenario where this `token` was actually cancelled by vscode?

ramya-rao-a: In that case what is the need to have our own token?

JacksonKearl: Bit confused. The token just signals that the result will be discarded. If we don't listen to it and kill the process ourselves wouldn't it keep going?

PR #1795, opened 2018-07-17
Move to DocumentSymbol

### LINKED ISSUE #1772: Adopt DocumentSymbol API
With the 1.25 release we have finalised the API for the [outline tree](https://code.visualstudio.com/updates/v1_25#_outline-view). A `DocumentSymbolProvider` should now return [document symbols](https://github.com/Microsoft/vscode/blob/bd953248ad6a8e2f8896bfb78b3ae40e0fd47426/src/vs/vscode.d.ts#L2474) which can represent a hierarchy and that can differentiate between the full range of  a symbol and the selection range, e.g. the name of a function. 

*Note* that we synthesise trees using range-containment when a provider returns `SymbolInformation`-objects and that that isn't perfect. Adopting the new API makes the outline more robust and support features like follow-cursor and diagnostics association. 

Let me know if need more information is needed. Happy Coding!

JacksonKearl:
this closes #1772

### DIFF
--- a/src/goGenerateTests.ts
+++ b/src/goGenerateTests.ts
@@ -65,7 +65,7 @@ export function toggleTestFile(): void {
 	vscode.commands.executeCommand('vscode.open', vscode.Uri.file(targetFilePath));
 }
 
-export function generateTestCurrentPackage(): Thenable<boolean> {
+export function generateTestCurrentPackage(): Promise<boolean> {
 	let editor = checkActiveEditor();
 	if (!editor) {
 		return;
@@ -74,7 +74,7 @@ export function generateTestCurrentPackage(): Thenable<boolean> {
 		vscode.workspace.getConfiguration('go', editor.document.uri));
 }
 
-export function generateTestCurrentFile(): Thenable<boolean> {
+export function generateTestCurrentFile(): Promise<boolean> {
 	let editor = checkActiveEditor();
 	if (!editor) {
 		return;
@@ -83,32 +83,30 @@ export function generateTestCurrentFile(): Thenable<boolean> {
 		vscode.workspace.getConfiguration('go', editor.document.uri));
 }
 
-export function generateTestCurrentFunction(): Thenable<boolean> {
-	let editor = checkActiveEditor();
+export async function generateTestCurrentFunction(): Promise<boolean> {
+	const editor = checkActiveEditor();
 	if (!editor) {
 		return;
 	}
 
-	return getFunctions(editor.document).then(functions => {
-		let currentFunction: vscode.DocumentSymbol;
-		for (let func of functions) {
-			let selection = editor.selection;
-			if (selection && func.range.contains(selection.start)) {
-				currentFunction = func;
-				break;
-			}
+	const functions = await getFunctions(editor.document);
+	const selection = editor.selection;
+	const currentFunction: vscode.DocumentSymbol = functions.find(func => {
+		if (selection && func.range.contains(selection.start)) {
+			return true;
 		}
-		if (!currentFunction) {
-			vscode.window.showInformationMessage('No function found at cursor.');
-			return Promise.resolve(false);
-		}
-		let funcName = currentFunction.name;
-		if (funcName.includes('.')) {
-			funcName = funcName.split('.')[1];
-		}
-		return generateTests({ dir: editor.document.uri.fsPath, func: funcName },
-			vscode.workspace.getConfiguration('go', editor.document.uri));
+		return false;
 	});
+
+	if (!currentFunction) {
+		vscode.window.showInformationMessage('No function found at cursor.');
+		return Promise.resolve(false);
+	}
+	let funcName = currentFunction.name;
+	if (funcName.includes('.')) {
+		funcName = funcName.split('.')[1];
+	}
+	return generateTests({ dir: editor.document.uri.fsPath, func: funcName }, vscode.workspace.getConfiguration('go', editor.document.uri));
 }
 
 /**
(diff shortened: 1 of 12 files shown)


vladbarosan:
@JacksonKearl  I hope you don't mind I updated the PR to the latest master and refactored it a little

vladbarosan:
@JacksonKearl , @ramya-rao-a , @jhendrixMSFT  if you can have a look since its a change that touches multiple parts. Still testing all the functionality touched. will update when im done

src/goGenerateTests.ts
```diff
@@ -83,32 +83,30 @@ export function generateTestCurrentFile(): Thenable<boolean> {
...
-			funcName = funcName.split('.')[1];
+	const functions = await getFunctions(editor.document);
+	const selection = editor.selection;
+	const currentFunction: vscode.DocumentSymbol = functions.find(func => {
+		if (selection && func.range.contains(selection.start)) {
+			return true;
 		}
-		return generateTests({ dir: editor.document.uri.fsPath, func: funcName },
-			vscode.workspace.getConfiguration('go', editor.document.uri));
+		return false;
```
ramya-rao-a: The above can be simplified to `return func.range.contains(selection.start)`
Active editors will always have a selection

src/goOutline.ts
```diff
@@ -51,31 +51,20 @@ export interface GoOutlineOptions {
 	 */
 	document?: vscode.TextDocument;
 
-	/**
-	 * Skips range information in the output.
-	 * Calculating ranges is slightly expensive for large files, therefore skip it when not required.
-	 */
-	skipRanges?: boolean;
```
ramya-rao-a: Why have we deleted the `skipRanges` option?

src/goOutline.ts
```diff
@@ -151,44 +139,49 @@ function convertToCodeSymbols(
...
+		if (decl.type === 'type') {
+			let line = document.lineAt(document.positionAt(start));
+			let regex = new RegExp(`^\\s*type\\s+${decl.label}\\s+struct\\b`);
+			decl.type = regex.test(line.text) ? 'struct' : 'type';
 		}
 
-		let symbolInfo = new vscode.SymbolInformation(
+		const symbolInfo = new vscode.DocumentSymbol(
 			label,
+			decl.type,
```
ramya-rao-a: Is this the "details" that is shown next to the symbol in the outline feature?

src/goOutline.ts
vladbarosan: @ramya-rao-a  the `DocumentSymbol` it wont accept null paramters for ranges ( i.e in needs ranges).

vladbarosan: was this something we needed ? from what I could tell we were not actually skipping ranges in the provider

src/goOutline.ts
vladbarosan: yes ( so in this case would look like the name and next to it the type ( e.g. package, function etc)

vladbarosan: also looking in go-outline, everything seems fine, but we stop only at the function level. do we own https://github.com/lukehoban/go-outline ? seems we could get more from the AST

src/goOutline.ts
ramya-rao-a: Currently that option is being used in only in the `Go: Add Imports` command which doesnt need the range.

The only benefit is we avoid the work involved in calculating the ranges.

src/goOutline.ts
ramya-rao-a: We own the fork of it https://github.com/ramya-rao-a/go-outline
Feel free to look into that for improvements

src/goGenerateTests.ts
```diff
@@ -193,12 +186,9 @@ function generateTests(conf: Config, goConfig: vscode.WorkspaceConfiguration): T
...
-		.provideDocumentSymbols(doc, null)
-		.then(symbols =>
-			symbols.filter(sym =>
-				sym.kind === vscode.SymbolKind.Function)
-		);
+async function getFunctions(doc: vscode.TextDocument): Promise<vscode.DocumentSymbol[]> {
+	const documentSymbolProvider = new GoDocumentSymbolProvider();
+	const symbols = await documentSymbolProvider.provideDocumentSymbols(doc, null);
+	const symbols_1 = symbols[0].children;
+	return symbols_1.filter(sym => sym.kind === vscode.SymbolKind.Function);
```
ramya-rao-a: Pardon my Go ignorance here...
We are picking up on only the functions that are the first level children of the `package` as per the data returned from the `go-outline` tool.
Does Go have any concept where there could functions at a deeper level?

vladbarosan: @ramya-rao-a plain(named) functions can only be declared on the top level (last time I checked, I doubt anything changed in 1.12 ). 

You can have anonymous functions declared and assigned to variables though
```go
funcVar := func(arg string) {}
```
but this is not exactly the same thing.

src/goOutline.ts
vladbarosan: well it seems we might be able to get more info from the AST, of things inside the functions, @jhendrixMSFT has more experience with the Go AST, he can confirm.

src/goOutline.ts
vladbarosan: doesn't seem to be a way to avoid the ranges as it throws if those are null.

vladbarosan:
@ramya-rao-a @jhendrixMSFT @JacksonKearl everything touched seemed fine on testing, so if you want to double check, otherwise from my side I think its good to go.

src/goOutline.ts
jhendrixMSFT: Absolutely.  I've opened https://github.com/Microsoft/vscode-go/issues/2370 to track enhancing the outline view.

Issue #1780, opened 2018-07-09
launch.json "mode": "auto"


gertcuykens:
`auto` mode please that switches between "test" and "debug" depending the selected file is *_test.go
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch",
            "type": "go",
            "request": "launch",
            "mode": "test", <= dynamiclly adjust depending on selected file *_test
            "remotePath": "",
            "port": 2345,
            "host": "127.0.0.1",
            "program": "${fileDirname}",
            "env": {},
            "args": [],
            "showLog": true
        }
    ]
}
```

ramya-rao-a:
That would be cool. But instead of introducing a new mode, how about updating the current `debug` mode to be smart enough to use `test` when debugging a test file?

PRs are welcome. The change should be pretty simple.

Code pointers: 
- Wiki for [Running the Go extension from source](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code) and [debugging the debug adapter](https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter)
- https://github.com/Microsoft/vscode-go/blob/0.6.84/src/goDebugConfiguration.ts#L26 is where the change should be made

gertcuykens:
How do you get current selected file path in vscode?

ramya-rao-a:
vscode.window.activeTextEditor

ramya-rao-a:
@gertcuykens @qzyse2017 on second thoughts, having a separate mode called `auto` is better. This way we wont be breaking the expectation of users who might have been just looking at their test code but then want to debug the actual package.

@qzyse2017 See https://github.com/Microsoft/vscode-go/commit/2cc995250e450b8418065188a8f16064d2595cf2

PR #1800, opened 2018-07-22
Add auto `test` debug mode for launch.json (#1780)

qzyse2017:
I tested it manually and it worked correctly for #1780 , I don' t know how to add a unit test myself, I run the tests already in the repo and it performed just as before I made a change.

### DIFF
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -29,16 +29,22 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			if (!activeEditor || activeEditor.document.languageId !== 'go') {
 				return;
 			}
+			let debugMode;
+			let testFileRegExp = RegExp('.*_test\.go$', 'g');
+			if (testFileRegExp.test(activeEditor.document.fileName) === true) {
+				debugMode = 'test';
+			} else {
+				debugMode = 'debug';
+			}
 
 			debugConfiguration = {
 				'name': 'Launch',
 				'type': 'go',
 				'request': 'launch',
-				'mode': 'debug',
+				'mode': debugMode,
 				'program': activeEditor.document.fileName
 			};
 		}
-
 		const gopath = getCurrentGoPath(folder ? folder.uri : null);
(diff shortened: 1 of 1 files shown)


src/goDebugConfiguration.ts
```diff
@@ -29,16 +29,22 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
...
+			} else {
+				debugMode = 'debug';
+			}
 
 			debugConfiguration = {
 				'name': 'Launch',
 				'type': 'go',
 				'request': 'launch',
-				'mode': 'debug',
+				'mode': debugMode,
```
ramya-rao-a: When there is a debug configuration in the launch.json file, this `if` block (starting at line 27) will not be entered. So this fix will only get applied to cases when there is no debug configuration in the launch.json file.

I would suggest to update the mode after this `if` block

ramya-rao-a: When there is a debug configuration in the launch.json file, this `if` block (starting at line 27) will not be entered. So this fix will only get applied to cases when there is no debug configuration in the launch.json file.

I would suggest to update the mode after this `if` block

src/goDebugConfiguration.ts
```diff
@@ -29,16 +29,22 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			if (!activeEditor || activeEditor.document.languageId !== 'go') {
 				return;
 			}
+			let debugMode;
+			let testFileRegExp = RegExp('.*_test\.go$', 'g');
+			if (testFileRegExp.test(activeEditor.document.fileName) === true) {
```
ramya-rao-a: isnt it simpler with `if (activeEditor.document.fileName.endsWith('_test.go'))` ?

qzyse2017:
sorry I had deleted the branch which I used to change the code before you reviewed😥.
It seems there is no way I can continue to commit to this pull request.
Created a new pull request [here](https://github.com/Microsoft/vscode-go/pull/1832)

ramya-rao-a:
Fixes #1780

PR #1801, opened 2018-07-22
launch.json env option to display numbers as hex while debugging

valerino:
wonder why this was missing.......


ramya-rao-a:
Can you elaborate on some examples where this will be useful?
Is the need for this Golang specific or a generic requirement?
Is this a common enough requirement or a very special case?

valerino:
hi ramya,
well..... in a debugger, you should be able to display int values as hex. 
for example, i'm developing a computer emulator in golang (commodore 64 emulator) and in this situation having hex values displayed while debugging is vital.
but well...... this is not the only case, i mean ..... just to display a buffer in memory while debugging, without relying to printf, and so on .....

src/debugAdapter/goDebug.ts
```diff
@@ -519,6 +517,19 @@ class GoDebugSession extends DebugSession {
 		if (/^(\w:[\\/]|\\\\)/.test(path)) return '\\';
 		return path.includes('/') ? '/' : '\\';
 	}
+	protected convertToHex(v) {
+		let s = v;
+		if (this.numAsHex === 'hex') {
+			s = '0x' + parseInt(v).toString(16);
```
jhendrixMSFT: This will truncate floating-point numbers.  Perhaps this functionality should be limited to integers?

ramya-rao-a:
@jhendrixMSFT Does https://github.com/Microsoft/vscode/issues/28025#issuecomment-414013391 work for Go?

jhendrixMSFT:
@ramya-rao-a no it doesn't work, would be great to fix it.

marco-m-pix4d:
Hello, is this PR still valid?

ramya-rao-a:
Closing this PR as it has not seen any activity after the review by @jhendrixMSFT last year https://github.com/microsoft/vscode-go/pull/1801#pullrequestreview-191430426

If anyone is interested in reviving this PR, please create a new one in https://github.com/golang/vscode-go as [we are moving](https://github.com/microsoft/vscode-go#we-are-moving---) the repo for this extension.

PR #1826, opened 2018-08-05
The name of the extension should be lowercase

yokoe:
According to this page, name of the extension should be all lowercase with no spaces.
https://code.visualstudio.com/docs/extensionAPI/extension-manifest

Other extensions are obeying this rule. (https://github.com/Microsoft/vscode-azureappservice/blob/master/package.json, https://github.com/Microsoft/vsts-vscode/blob/master/package.json)

Changing the name seems dangerous though...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "Go",
+  "name": "go",
   "version": "0.6.86-beta.3",
   "publisher": "ms-vscode",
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
> Changing the name seems dangerous though...

Well... Changing the name, changes the extension id which can cause a lot of issues, so I dont believe we should be changing this.

Thanks for the effort though.

PR #1828, opened 2018-08-06
Fix error handling for when err is not a string

fstanis:
Currently, I'm getting a `spawn EACCES` error, but the console is reporting it as `TypeError: err.startsWith is not a function` because the `err` variable is an object and doesn't have the `startsWith` method.


ramya-rao-a:
This fix is now out in the latest update (0.6.86)

PR #1832, opened 2018-08-09
Add auto `test` debug mode for launch.json (#1780)

qzyse2017:
use regex to check if the current file is for test

### DIFF
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -54,10 +54,8 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['dlvLoadConfig'] = dlvConfig['dlvLoadConfig'];
 		}
 
-		if (activeEditor.document.fileName.endsWith('_test.go')) {
+		if (activeEditor && activeEditor.document.fileName.endsWith('_test.go')) {
 			debugConfiguration['mode'] = 'test';
-		} else {
-			debugConfiguration['mode'] = 'debug';
 		}
 
 		return debugConfiguration;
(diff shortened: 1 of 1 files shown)


src/goDebugConfiguration.ts
```diff
@@ -54,6 +54,15 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['dlvLoadConfig'] = dlvConfig['dlvLoadConfig'];
 		}
 
+		let debugMode;
+		let testFileRegExp = RegExp('.*_test\.go$', 'g');
```
ramya-rao-a: Isnt it simpler with `if (activeEditor.document.fileName.endsWith('_test.go'))` ?
I have a tendency to avoid regex when possible.

src/goDebugConfiguration.ts
```diff
@@ -54,6 +54,15 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['dlvLoadConfig'] = dlvConfig['dlvLoadConfig'];
 		}
 
+		let debugMode;
+		let testFileRegExp = RegExp('.*_test\.go$', 'g');
+		const currentFileName = vscode.window.activeTextEditor.document.fileName;
+		if (testFileRegExp.test(currentFileName)) {
```
ramya-rao-a: Isnt it simpler with `if (activeEditor.document.fileName.endsWith('_test.go'))` ?
I have a tendency to avoid regex when possible.

ramya-rao-a:
Fixes #1780

Issue #1829, opened 2018-08-08
Using vscode-go from multiple user accounts clashes on /tmp/go-code-check


lmb:
Steps to Reproduce:

1. On a system where TMPDIR is shared for the whole machine (e.g. Ubuntu)
2. Create two user accounts A and B
3. Work on Go code with A
4. Switch to user B, work on Go code
5. Observe error message

```
go build package/name: open /tmp/go-code-check: permission denied
```

```
$ ls -lh /tmp/go-code-check
-rw-rw-r-- 1 private private 173K Aug  7 22:54 /tmp/go-code-check
```

I believe this is due to https://github.com/Microsoft/vscode-go/blob/a1b8648a0b2ef5858669fa648a438f79c668f256/src/goBuild.ts#L59 assuming that TMP is exclusive to a user, which isn't the case on a default Ubuntu install.

lmb:
Option 1: generate a random subdirectory, see 
https://golang.org/src/io/ioutil/tempfile.go?s=2138:2195#L66

Option 2: include the UID in the temp file name.

ramya-rao-a:
Thanks for logging this issue @lmb and Thanks for the PR @rikkuness! Your fix is now available in the latest update (0.6.86) to the Go extension

lmb:
Thank you both for the quick fix!

PR #1835, opened 2018-08-09
Make temporary build filename user specific to prevent conflict on systems with shared temp folders

rikkuness:
As per #1829

### DIFF
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -5,6 +5,7 @@ import { outputChannel } from './goStatus';
 import os = require('os');
 import { getNonVendorPackages } from './goPackages';
 import { getTestFlags } from './testUtils';
+import { getUsernameHash } from './util';
 import { getCurrentGoWorkspaceFromGOPATH } from './goPath';
 import { diagnosticsStatusBarItem } from './goStatus';
 /**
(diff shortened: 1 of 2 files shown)


src/goBuild.ts
```diff
@@ -56,7 +56,7 @@ export function goBuild(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfigura
 	}
 
 	const buildEnv = Object.assign({}, getToolsEnvVars());
-	const tmpPath = path.normalize(path.join(os.tmpdir(), 'go-code-check'));
+	const tmpPath = path.normalize(path.join(os.tmpdir(), 'go-code-check.' + os.userInfo().username));
```
ramya-rao-a: Can we hash that name? There are cases where when trying to investigate an issue I request users to paste a screenshot of the output channel. User name is clearly PII 100% of the time and I want to stay clear of it.

Also, if we hash lets hash it once and re-use it.

rikkuness: It doesn't really even need be a username, the filename could be anything providing it didn't cause a collision with another user.

I guess a hash based off username would be okay though, saves randomizing the filename and filling up /tmp.

I'll take another look at it 👍

PR #1849, opened 2018-08-15
Fix  `build worksapce fails when editing _test file`#(1789)

qzyse2017:
Remove `isTestFile` check when `buildWorkspace` is true


qzyse2017:
There seems no need to check whether the current file is a '_test' file when choosing 'build Workspace' command

Issue #1823, opened 2018-08-01
Enhance the Generate Interface Stubs command to include user interface to browse packages and select the interface


covrom:
Proposal. Please, add function "Implement interface" by right click on type name and select interface from current package, current project, current scope, core golang packages, package by name. This function inserts empty functions definitions for implement selected interface.

ramya-rao-a:
We use the [impl](https://github.com/josharian/impl) tool to provide this feature via the command `Go: Generate Interface Stubs`. What is missing is the user interface to select the interface from packages which should be straight forward to implement. PRs are welcome.

Code pointers:
- User interface for browsing packages: https://github.com/Microsoft/vscode-go/blob/master/src/goBrowsePackage.ts
- Once the package is selected, you can use the `go list` command on it to get all the files
- Once a file is selected you can use the `go-outline` tool to get the interfaces

stamblerre:
Future work in this space will be done in `gopls`, rather than in this extension itself. I'm going to close this issue in favor of the upstream one: https://github.com/golang/go/issues/37537.

Issue #1741, opened 2018-06-19
Feature Request: Better interface stubbing


matthewmueller:
It can be hard to get the input right with the current implementation of `Generate Interface Stubs`. 

For example, if your path is `github.com/matthewmueller/graphql` and you have a `Query` interface and you want your `query` struct to implement it:

```go
package graphql

// interface to implement
type Query interface {
  CreateTeam(ctx context.Context, teamInput TeamInput) (Team, error)
  // ... many more
}

// this struct needs to implement the Query interface
type query struct {}

// interface enforcement
var _ Query = (*query)(nil)
```

With the `Generate Interface Stubs`, you'd need to write in the following:

```
q *query github.com/matthewmueller/graphql.Query
```

This is quite hard to get right. I think it'd be a lot more useful if it did the following:

1. Highlight over the interface enforcement with your cursor: `var _ Query = (*query)(nil)`
2. Right click, and click on `Generate Interface Stubs`
3. It calls `impl` under the hood with the correct paths. The receiver variable could just be the first letter of the struct (lowercased)
4. The output is written below your cursor selection

---

**Update:** Even better would be how typescript does it with the little lightbulb next to any interface enforcement that is currently erroring out.

qzyse2017:
@matthewmueller could you describe your demand more specific with your directory structure and the path of your project? I would like to take this issue but it seems you just need to type in `graphql.Query` instead of `github.com/matthewmueller/graphql.Query` , the current implemantation in vscode -go extension which make use of `impl` seems to perform the same with `github.com/matthewmueller/graphql.Query` and  `graphql.Query`

qzyse2017:
If you put your project under `$GOPATH` directory, `impl` will automatically search it and you just need to type in the package name without its previous path if it is not ambiguous with another package

matthewmueller:
@qzyse2017 awesome! I'm not able to reproduce needing the full path, maybe I was typing in the command wrong. That's definitely easier

Either way, I think it would be awesome if it could auto implement: `var _ Query = (*query)(nil)`

qzyse2017:
@matthewmueller I am doing it in a similar way to what you need.
It  just needs to 
1. select the reciever type name and use right click to choose the command `Generate interface stub `, the inputbox will have the `t *typeReciever`(something like it) auto completed 
2. type in the last arg: `packageName.interfaceName`(packageName here do not need its previous path) 

[...]

PR #1863, opened 2018-08-21
workaround for #1741 and #1823

qzyse2017:
1. add autocomplete for package name and allow the user to type in arguments of "Generate interface stubs" just as what they will do in a file with the package already imported.
e.g.You need to type in "github.com/hyperledger/fabric/core/shim.Chaincode"(before) => "shim.Chaincode"(after)

2. allow the user to selected the type name and right click to choose "Generate interface stubs" and the type name and reciever name will be autocompleted. 

3. allow the user to omitted the third arg(inteface name) and choose it manually from all packages(a little slow it you just start vscode and have not used `goBrowsePackages`) and it will occupy much of your cpu and memery resources for the first time(I just make use of the current `getAllPackages` to implement manual search for interface).


qzyse2017:
The use is like below 
test
test1

PR #1865, opened 2018-08-22
Use ListPackageVars to show global variables

### LINKED ISSUE #1854: How to see global variables in visual studio for golang while debugging?
While debugging in GoLang using delve only the local variables are shown. Cant see all the global variables. Is there any settings change to see the list of all variables
screen shot 2018-08-17 at 11 12 09 am
Here a is not showing Even when I hover mouse over it its value is not coming.

I also added this setting :- 
"go.delveConfig": {
        "useApiV1": false,
        "dlvLoadConfig": {
            "followPointers": true,
            "maxVariableRecurse": 2,
            "maxStringLen": 64,
            "maxArrayValues": 64,
            "maxStructFields": -1
        }
    }
Still not coming

ramya-rao-a:
This PR enables use to have the global variables in the current package show up in the variables pane.
Fixes #1854
Doesnt work for remote debugging at the moment.

@lggomez Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Go",
-  "version": "0.6.88",
+  "version": "0.6.89-beta.1",
   "publisher": "ms-vscode",
   "description": "Rich Go language support for Visual Studio Code",
   "author": {
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -521,6 +521,8 @@ class GoDebugSession extends DebugSession {
 	private packageInfo = new Map<string, string>();
 	private launchArgs: LaunchRequestArguments;
 
+	private readonly initdone = 'initdone·';
+
 	public constructor(debuggerLinesStartAt1: boolean, isServer: boolean = false) {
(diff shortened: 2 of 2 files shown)


lggomez:
Neat

Are you sure about remote not working @ramya-rao-a ? I could see the globals just fine locally, at least with delve `Version 1.0.0 - Build: $Id: c98a142125d0b17bb11ec0513bde346229b5f533`

Also a minor detail, but we could filter the [initdone variable](https://github.com/golang/go/blob/ae2a2d12f6d8cde35637a13f384f6de524112768/src/cmd/compile/internal/gc/init.go#L58-L75). It seems to be a global inyected by internal package initialization code, and it should be != 2 during debugging (in a sane scenario)

ramya-rao-a:
The remote scenario would need us to map the remote file to local file and then use the local file to determine the package name/import path. I didn't want to include that in the first go as I didn't have the time to test it out.

Good point about initdone. I'll update accordingly.

ramya-rao-a:
Ok, I've updated the PR to support remote paths as well.
@lggomez Could you give it a try?

PR #1866, opened 2018-08-22
Language Server in Windows

ramya-rao-a:
Now that there is considerable progress in the Sourcegraph language server for Windows, its time we integrate the same in Windows.


doxxx:
I just tested this out and with regards to your [comment](https://github.com/sourcegraph/go-langserver/issues/113#issuecomment-414876125), I get the same results except:

* Signature help works
* Workspace Symbol Search works
* Find References works but suffers from the same problem as `guru` on Windows: searching for references of a symbol at the definition site of the symbol only finds references in the same package but searching for references from a usage site of the symbol outside of its package finds all references in all packages.

Issue #1842, opened 2018-08-13
Separate build-tags and test-tags


johan-lejdung:
Hi!

I have come across a use-case that is not covered by this extension, a use-case that I cannot imagine being alone with.

For many of my Go projects I have integration tests, these are tests that spin up docker containers and take longer to run. I also like to test my code on save.. To let me have the best of two worlds, fast running tests on save and longer integration tests on demand, I tag my integration tests with:
```
// +build integration
```

This stops the file from building unless I supply the tag with the build or test command.

Unfortunately this also disables the linting and syntax checks on the file, making it really hard to verify the code is working.

**Atom** solves this by having an option for build tags for build/install commands and a command for build tags for tests. I am not finding it possible to do a similar configuration with Visual Studio Code. All i find is `go.buildTags` which affects both build and save.

Are there any plans to fix that?

ramya-rao-a:
When you say you'd like to test your code on save, are you referring to the linter and syntax checks or actual tests that are run?

johan-lejdung:
Well both of it, but I am referring to actual tests being run.

As long as the tests are simple unit tests they run fast enough, but of course they only run if the linter and syntax checks pass :)

ramya-rao-a:
I would suggest the below:
- Add the tag `!fastRunning` on the files that have the integration tests
- Add `"go.testFlags": ["-tags", "fastRunning"]` to your settings
At this point, linter and syntax checks will run on all files, test on save will not run on the integration tests.

Set up a test task to run just all tests. Tasks have no knowledge of the go extension and so no knowldge of its settings.
- Run command `Tasks: Configure Task` -> `Create tasks.json file from template` -> Choose "Others" from the template options
- Update the command to `go test <your package name>`. 
- Run command `Tasks: Run Task` and then choose your task to run all the tests on demand.

See https://code.visualstudio.com/Docs/editor/tasks to learn more about tasks

ramya-rao-a:
Oh wait, my bad. The `go.testFlags` get used when compiling so you wont get the syntax checks. This is because `go test` is used to compile test files and `go build` is used to compile non test files.

You can use a small variation of the above.

- Add the `integration` tag like you first described
- Have the task run the build and/or run tests using the tag

You can now run the task on demand every once in a while to get feedback on errors or to run the tests themselves

johan-lejdung:
But then we are back full-circle. I still wont have linting and syntax checks on my test files unless I run it manually, which I already do today - task or not -.

[...]

PR #1877, opened 2018-08-26
feat: added testFlags option

johan-lejdung:
I tested it out locally and it worked fine and all tests except rungodoc passed but it looked like I had to do some setup for that to pass to that is most likely the reason.


johan-lejdung:
Solves: https://github.com/Microsoft/vscode-go/issues/1842

src/testUtils.ts
```diff
@@ -166,7 +166,7 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 			outputChannel.show(true);
 		}
 
-		let buildTags: string = testconfig.goConfig['buildTags'];
+		let buildTags: string = testconfig.goConfig['testTags'];
```
ramya-rao-a: I believe we should fallback to using `buildTags` if no `testTags` are provided to ensure we are not breaking anybody else's use case.

package.json
```diff
@@ -515,9 +515,15 @@
         "go.buildTags": {
           "type": "string",
           "default": "",
-          "description": "The Go build tags to use for all commands that support a `-tags '...'` argument",
+          "description": "The Go build tags to use for all commands, excluding `go test` (see testTags), that support a `-tags '...'` argument",
```
ramya-rao-a: `buildtags` will still be used for compiling the test files which will be using `go test`
So we should probably change this to `excluding when tests are run` or something to that effect

src/testUtils.ts
johan-lejdung: Of course!

src/testUtils.ts
```diff
@@ -166,7 +166,7 @@ export function goTest(testconfig: TestConfig): Thenable<boolean> {
 			outputChannel.show(true);
 		}
 
-		let buildTags: string = testconfig.goConfig['buildTags'];
+		let testTags: string = (testconfig.goConfig.has("testTags")) ? testconfig.goConfig['testTags'] : testconfig.goConfig['buildTags'];
```
ramya-rao-a: We need to use single quotes, else the linter breaks the build.

package.json
```diff
@@ -515,9 +515,15 @@
         "go.buildTags": {
           "type": "string",
           "default": "",
-          "description": "The Go build tags to use for all commands that support a `-tags '...'` argument",
+          "description": "The Go build tags to use for all commands, excluding when running tests (see testTags), that support a `-tags '...'` argument",
```
ramya-rao-a: The description now is a little confusing. Can we move the exclude statement to the end? How about the below?

`The Go build tags to use for all commands that support a `-tags '...'` argument. When running tests, `go.testTags` will be used instead if it was set.`

src/testUtils.ts
johan-lejdung: oops!

package.json
johan-lejdung: That does sound better :)

ramya-rao-a:
The `testTags` are to be used when running the tests. The unit tests are using the `check` function that is used to only compile the files and not run tests. Therefore, the `check` method will only use the build tags and will never use the test tags.

That's the reason for the failures

johan-lejdung:
Ah, thanks! Will get it sorted out today. Haven't had a second look yet, sorry for the delay :)

johan-lejdung:
I removed the test for fallback on `buildTags`. The reason was that [this line](https://github.com/Microsoft/vscode-go/pull/1877/commits/c29017cac6cc0edc830cd4c6bba588032d703108#diff-869d156f8d53e34e379625eb2aab2d2cR169) would evaluate to true even though 'testTags' was equal to `''`.

First I chose to modify the check to 
```
let testTags: string = testconfig.goConfig['testTags'].length > 1 ? testconfig.goConfig['testTags'] : testconfig.goConfig['buildTags'];
```

But that would create a weird dependency on `buildTags`, and remove the useCase we are looking to create; where you wanted tests to run without any tags but let it build the files with some tags.

Not sure how to solve the backwards compatibility problem actually since an unset config evaluated to an empty string and not undefined.

johan-lejdung:
Ehm, getting weird errors `Expected 3 arguments, but got 2.`

The function definition is as follows:
```
export function testCurrentFile(goConfig: vscode.WorkspaceConfiguration, args: string[]): Thenable<boolean>
```

Any ideas?

ramya-rao-a:
> Ehm, getting weird errors Expected 3 arguments, but got 2.

Merge from the `master` branch to fix that.

ramya-rao-a:
> and remove the useCase we are looking to create; where you wanted tests to run without any tags but let it build the files with some tags.

We had a similar case with the `go.testFlags` setting. We ended up supported `null` as a valid value. See below:

screen shot 2018-09-02 at 2 50 04 am

johan-lejdung:
That should do it!

johan-lejdung:
When can I expect this to get released? :)

johan-lejdung:
Super, thank you for the help!

Rhionin:
Bless you for building this! I've been dying for this feature!

johan-lejdung:
That's really nice to hear ☺️

PR #1893, opened 2018-08-30
Do not show blank identifiers in file outline

### LINKED ISSUE #1889: Outline should exclude `_` "blank identifier"
We use `var _` in the top-level of some of our Go files for various reasons, most notably the ["interface checks pattern"](https://golang.org/doc/effective_go.html#blank_implements) documented in Effective Go.

These should be excluded from the Outline view since they don't represent a real variable that can be referenced and they clutter the output.

For example this code will produce 3 entries for `_` in the outline view:
```go
package foo

var (
	_ io.Closer = Foo(nil)
	_ io.Closer = Bar(nil)
	_ io.Closer = Baz(nil)
)
```

ragurney:
Resolves #1889

### DIFF
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -110,7 +110,7 @@ export class GoDocumentSymbolProvider implements vscode.DocumentSymbolProvider {
 		symbols: vscode.SymbolInformation[],
 		containerName: string,
 		byteOffsetToDocumentOffset: (byteOffset: number) => number): void {
-		const ignoredSymbols = ['_'];
+
 		let gotoSymbolConfig = vscode.workspace.getConfiguration('go', document.uri)['gotoSymbol'];
 		let includeImports = gotoSymbolConfig ? gotoSymbolConfig['includeImports'] : false;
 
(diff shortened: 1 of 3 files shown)


ragurney:
Needs a test, having trouble getting the `test.go` file to work.

ramya-rao-a:
The `gogetdoc` related failures are because you don't have `gogetdoc` installed. Someone is working on updating the tests to not fail in such cases.

You can comment them and the gometalinter one while running your tests. Your changes are unrelated to them so you can safely ignore them.

> When trying to add a line such as var _ string = "foo", I get two new failures as well,

That is mostly because there are other tests using the same file and now all the line numbers are changing. Try this
- In the `fixtures` table, create a new folder for `outline` and add a go file there as your test file. Update all the outline related tests to use this file

ramya-rao-a:
@ragurney If you merge from master branch now, you will not get those failures

ragurney:
@ramya-rao-a sounds good, thanks! I think this PR is ready for another review.

ramya-rao-a:
Looks great, Thanks!

PR #1898, opened 2018-08-30
Add command: "Go: Benchmark package."

biancarosa:
This PR adds a new command called "go.benchmark.package" that runs benchmark tests for the current package.
I did this adding a new isBenchmark flag to the testCurrentPackage file. 
Also, according to [this](https://golang.org/cmd/go/#hdr-Testing_flags) docs, the `-bench` flag doesn't have a default value and by default doesn't run any tests. I've edited the testConfig to send `.` when it is running for a package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,11 @@
         "description": "Runs all unit tests in the package of the current file."
       },
       {
+        "command": "go.benchmark.package",
+        "title": "Go: Benchmark Package",
+        "description": "Runs all unit tests in the package of the current file."
+      },
+      {
         "command": "go.test.workspace",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -113,6 +113,11 @@
         "title": "Go: Test Package",
         "description": "Runs all unit tests in the package of the current file."
       },
+      {
+        "command": "go.benchmark.package",
+        "title": "Go: Benchmark Package",
+        "description": "Runs all unit tests in the package of the current file."
```
ramya-rao-a: Runs all benchmarks? Otherwise, things look great!

PR #1900, opened 2018-08-31
Added completions for symbols from unimported pkg if there are multiple matches for pkg name

### LINKED ISSUE #1884: No completions for symbols from unimported pkg if there are multiple matches for pkg name
`fmt.` gives completions for all the symbols from the fmt package
`template.` doesn't do the same.

This is because we cannot determine which of the `text/template` and `html/template` should be used.

Proposed fix:
- When encountering `template.` provide completions for all the matching pkgs
- Once user chooses the package, the import should be added for that package, and completions should be triggered again.

This is possible to do because we can run commands as part of the completion. But since we can run one 1 command for a completion item, we will have to create a new command that internally runs `go.import.add` command followed by triggering suggestions again  with `editor.action.suggest` command

shreyu86:
Fixes #1884

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -117,7 +117,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 					if (suggestions.length === 0 && lineTillCurrentPosition.endsWith('.')) {
 
 						let pkgPath = this.getPackagePathFromLine(lineTillCurrentPosition);
-						if (pkgPath) {
+						if (pkgPath.length === 1) {
 							// Now that we have the package path, import it right after the "package" statement
 							let { imports, pkg } = parseFilePrelude(vscode.window.activeTextEditor.document.getText());
 							let posToAddImport = document.offsetAt(new vscode.Position(pkg.start + 1, 0));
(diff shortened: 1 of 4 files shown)


src/goSuggest.ts
```diff
@@ -131,11 +131,28 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 								// add additionalTextEdits to do the same in the actual document in the editor
 								// We use additionalTextEdits instead of command so that 'useCodeSnippetsOnFunctionSuggest' feature continues to work
 								newsuggestions.forEach(item => {
-									item.additionalTextEdits = getTextEditForAddImport(pkgPath);
+									item.additionalTextEdits = getTextEditForAddImport(pkgPath[0]);
```
ramya-rao-a: Shouldnt we use `pkgPath[0]` at line 124 as well?

shreyu86:
Thanks, @ramya-rao-a I tested the new change. How do I go about updating the unit test? Can I hit tab on the first suggestion to trigger subsequent suggestion and then check for those?

ramya-rao-a:
Or you can run the command `acceptSelectedSuggestion`

shreyu86:
@ramya-rao-a I will update the tests and have this PR up for review soon, sorry I was out of office.

ramya-rao-a:
No problem, take your time. I was out last week on vacation as well :)

shreyu86:
@ramya-rao-a I attempted to refactor the tests but was not successful in `acceptSelectedSuggestion` can you please provide some samples/pointers?

ramya-rao-a:
- After `showTextDocument`, set `editor.selection` to the right position where we want suggestions to be triggered
- Then use `vscode.commands.executeCommand('editor.action.triggerSuggest')` followed by `vscode.commands.executeCommand('acceptSelectedSuggestion')`
- Then check the document contents

Note: These are just ideas, I havent tried them out myself :)

shreyu86:
@ramya-rao-a I tried couple of options ^^ and I was not able to get the test refactored, can you please help?

ramya-rao-a:
The problem is that the suggest widget is not completely rendered before the accept suggestion command is executed.

Never mind, I think we can skip that.

I pushed a commit for some extra checks in your test case. This is good to go.

Thanks again!

shreyu86:
Thanks a lot for help on this! Really appreciate it.

shreyu86:
Thanks a lot @ramya-rao-a

ramya-rao-a:
No problem at all

PR #1912, opened 2018-09-04
Use mkdtemp for secure temp files

### LINKED ISSUE #1905: Insecure use of temporary file
There are multiple insecure uses of temporary files in the vscode-go extension. I found the following just looking at the output console:

- `$TMPDIR/go-code-cover` . Command palette: "> Go Toggle Test Coverage In Current Package".
- `$TMPDIR/go-code-check.*` . See #1829.
- `$TMPDIR/vscode-go-debug.txt` . Created when debugging with `trace: "true"` in the launch config.

On a multi-user or shared system such clashing or predictable use of temporary files is a hassle at best (two users cannot use vscode without errors) to a serious security issue at worst.

For example, on some system configurations user Bob can overwrite arbitrary files belonging to user Alice (or root) using a symlink attack:

```
bob@box$ ln -s /home/alice/thesis.tex /tmp/go-code-cover
bob@box$ ln -s /etc/passwd /tmp/vscode-go-debug.txt 
```
(This example is prevented on modern Linux systems if /tmp has the sticky bit set.)

Recommendation: use secure library functions for temporary files and private files (e.g., in the home directory or workspace) for non-temporary files.

Charliekenney23:
Closes #1905.

This closely mirrors the [official TypeScript extension's implementation](https://github.com/Microsoft/vscode/blob/ad68ff316c472aea4acfa31b3a5bab3b101d968b/extensions/typescript-language-features/src/utils/electron.ts#L17), which does not persist the directory path for reuse between activations.

@ramya-rao-a, making the directory path persistent is pretty hacky and not the convention for any of the official extensions I've seen, so I've left it out (let me know your thoughts on this).

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -12,7 +12,7 @@ import { basename, dirname, extname } from 'path';
 import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, resolveHomeDir, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
-import { getTempFile } from '../util'
+import { getTempFile } from '../util';
 import * as logger from 'vscode-debug-logger';
 
 require('console-stamp')(console);
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -12,6 +12,7 @@ import { basename, dirname, extname } from 'path';
 import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, resolveHomeDir, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
+import { getTempFile } from '../util'
 import * as logger from 'vscode-debug-logger';
 
 require('console-stamp')(console);
@@ -531,7 +532,7 @@ class GoDebugSession extends DebugSession {
 		this.delve = null;
 		this.breakpoints = new Map<string, DebugBreakpoint[]>();
 
-		const logPath = path.join(os.tmpdir(), 'vscode-go-debug.txt');
+		const logPath = getTempFile('vscode-go-debug.txt');
 		logger.init(e => this.sendEvent(e), logPath, isServer);
 	}
 
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -12,7 +12,6 @@ import { basename, dirname, extname } from 'path';
 import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, resolveHomeDir, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
-import { getTempFile } from '../util';
 import * as logger from 'vscode-debug-logger';
 
 require('console-stamp')(console);
@@ -532,7 +531,7 @@ class GoDebugSession extends DebugSession {
 		this.delve = null;
 		this.breakpoints = new Map<string, DebugBreakpoint[]>();
 
-		const logPath = getTempFile('vscode-go-debug.txt');
+		const logPath = path.join(os.tmpdir(), 'vscode-go-debug.txt');
 		logger.init(e => this.sendEvent(e), logPath, isServer);
 	}
 
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -1,6 +1,6 @@
 import path = require('path');
 import vscode = require('vscode');
-import { getToolsEnvVars, runTool, ICheckResult, handleDiagnosticErrors, getWorkspaceFolderPath, getCurrentGoPath, TempFileProvider } from './util';
+import { getToolsEnvVars, runTool, ICheckResult, handleDiagnosticErrors, getWorkspaceFolderPath, getCurrentGoPath, getTempFilePath } from './util';
 import { outputChannel } from './goStatus';
 import os = require('os');
 import { getNonVendorPackages } from './goPackages';
@@ -70,7 +70,7 @@ export function goBuild(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfigura
 	}
 
 	const buildEnv = Object.assign({}, getToolsEnvVars());
-	const tmpPath = TempFileProvider.getFilePath('go-code-check');
+	const tmpPath = getTempFilePath('go-code-check');
 	const isTestFile = fileUri && fileUri.fsPath.endsWith('_test.go');
 	const buildFlags: string[] = isTestFile ? getTestFlags(goConfig, null) : (Array.isArray(goConfig['buildFlags']) ? [...goConfig['buildFlags']] : []);
 	const buildArgs: string[] = isTestFile ? ['test', '-c'] : ['build'];
--- a/src/goBuild.ts
+++ b/src/goBuild.ts
@@ -1,6 +1,6 @@
 import path = require('path');
 import vscode = require('vscode');
-import { getToolsEnvVars, runTool, ICheckResult, handleDiagnosticErrors, getWorkspaceFolderPath, getCurrentGoPath, getUserNameHash } from './util';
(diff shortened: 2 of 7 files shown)


src/goBuild.ts
```diff
@@ -70,7 +70,7 @@ export function goBuild(fileUri: vscode.Uri, goConfig: vscode.WorkspaceConfigura
 	}
 
 	const buildEnv = Object.assign({}, getToolsEnvVars());
-	const tmpPath = path.normalize(path.join(os.tmpdir(), 'go-code-check.' + getUserNameHash()));
+	const tmpPath = getTempFile('go-code-check.' + getUserNameHash());
```
ramya-rao-a: We can skip the use of user name hash here which was introduced to fix #1829
The current solution of using `fs.mkdtempSync` should fix #1829 as well.

src/util.ts
```diff
@@ -844,3 +844,19 @@ export function makeMemoizedByteOffsetConverter(buffer: Buffer): (byteOffset: nu
 	};
 }
 
+export const getTempDir = (() => {
+	let dir: string | undefined;
+	return (): string => {
+		if (!dir) {
+			dir = fs.mkdtempSync(os.tmpdir() + '/vscode-go');
```
ramya-rao-a: Use `path.sep` to create the path here instead of `/`

Charliekenney23:
@ramya-rao-a I've updated the code to persist the temp directory path between vscode sessions. Please let me know if there's anything else that needs to be updated!

src/debugAdapter/goDebug.ts
```diff
@@ -12,6 +12,7 @@ import { basename, dirname, extname } from 'path';
 import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, resolveHomeDir, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
+import { TempFileProvider } from '../util';
```
ramya-rao-a: The debug adapter runs in a separate process and does not have access to anything from the `vscode` module. Therefore a dependency here on `util` will fail as util imports from `vscode`.

The tmp dir is used in the debug adapter to create the trace files. Since this is not done by default and the user sets explicitly to debug an issue with the debugger, I believe we are ok with creating a new file under `os.tmpdir()`

src/util.ts
```diff
@@ -844,3 +844,45 @@ export function makeMemoizedByteOffsetConverter(buffer: Buffer): (byteOffset: nu
 	};
 }
 
+export class TempFileProvider {
+	private constructor() {}
+	private static _instance: TempFileProvider;
+
+	static get instance() {
```
ramya-rao-a: Why not make this a static class? Then, we can skip the `instance()` method here.

src/util.ts
```diff
@@ -844,3 +844,45 @@ export function makeMemoizedByteOffsetConverter(buffer: Buffer): (byteOffset: nu
...
+		if (!TempFileProvider._instance) {
+			TempFileProvider._instance = new TempFileProvider();
+		}
+		return TempFileProvider._instance;
+	}
+
+	private store: vscode.Memento;
```
ramya-rao-a: let's call this the `globalState` so that when code reading we know exactly where the data is stored

src/util.ts
```diff
@@ -844,3 +844,45 @@ export function makeMemoizedByteOffsetConverter(buffer: Buffer): (byteOffset: nu
...
+
+	/**
+	 * register store to provider for persistance
+	 */
+	registerStore(store: vscode.Memento) {
+		this.store = store;
+	}
+
+	get dir(): string {
```
ramya-rao-a: Since `getFilePath` is the only place this is used, can we merge this code into `getFilePath`?

src/util.ts
Charliekenney23: Very good point. I'll make this a static class 😃

src/debugAdapter/goDebug.ts
Charliekenney23: 👍

Charliekenney23:
I've updated the code with the suggested changes 😃

src/goMain.ts
```diff
@@ -83,6 +83,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 			});
 		}
 		ctx.globalState.update('goroot', currentGoroot);
+		TempFileProvider.registerStore(ctx.globalState);
```
ramya-rao-a: Move this to the very top. Just to make sure that the globalState is available in the TempFileProvider right from the start

src/util.ts
```diff
@@ -844,3 +844,32 @@ export function makeMemoizedByteOffsetConverter(buffer: Buffer): (byteOffset: nu
...
+	 */
+	static registerStore(globalState: vscode.Memento) {
+		TempFileProvider.globalState = globalState;
+	}
+
+	/**
+	 * returns path to temp file with name
+	 */
+	static getFilePath(name: string): string {
+		let tempDir = TempFileProvider.globalState.get<string>('tempDir');
```
ramya-rao-a: One corner case is that `getFilePath` gets called when the globalState is not yet set. This may not happen now, but when the code that registers the store gets moved around in the future, it can happen.

Charliekenney23: I've added a comment to above `TempFileProvider.registerStore must be called be for this`.

One work around could be creating a new temp dir if globalState is undefined (and storing it locally on the object). Thoughts?

antong:
Many systems clean the tmp directory every boot or on other schedules or even use a RAM based file system that is not persisted to disk. Will the persisted tmpdir in this PR work correctly if the tmpdir disappears?

Charliekenney23:
@antong yes. If you have a look [here](https://github.com/Charliekenney23/vscode-go/blob/secure-tmp-files/src/util.ts#L869), I'm checking if the persisted directory exists and creating it if not present.

antong:
Usually, mkdtemp and friends are used so that the library safely creates and picks a directory and then you use whatever dir you got. I haven't seen it used so that you recreate one with the same name you got earlier. It may be ok, but I'm not sure.

Now, you may call me overly paranoid, but what about the following attack scenario: Alice uses the vscode extension, which correctly makes a tmpdir /tmp/vscode.hMF460T1CB and it is persisted in vscode's config. Bob on the same machine monitors /tmp and sees Alice's tmpdir. It is inaccessible to him, so he can't do anything about it. However, Bob knows that /tmp is wiped on boot, and he crafts a cron job (a scheduled program execution) that as soon as possible (when the dir disappears) creates a directory with Alice's tmpdir name and sets the permissions to be accessible by Alice (e.g., world readable and writable) and performs the symlink attack I suggested earlier, e.g., `ln -s /home/alice/thesis.tex /tmp/vscode.hMF460T1CB/go-code-cover`. Alice then opens up vscode, the correctly named dir exists, so the vscode extension reuses it and her `thesis.tex` is overwritten.

Personally I'd err on the side of caution and not persist the tmpdir name, but recreate it every time and remove it when done with. If the concern is that junk may be left behind, we could remove all /tmp/vscode.* owned by the user at some convenient time, perhaps before creating a new tmpdir. If we want a persistent dir, then tmpdir might not be the right thing and we should put it under the users home dir or maybe the workspace dir.

Sorry for trying to figure out how to break things, I've my security goggles on this week :-)

Charliekenney23:
So I think the only benefit of persisting the temp directory path is to limit VSCode from polluting the `/tmp` directory (like with Microsoft/vscode#53423). While the scenario that you walked through would be extreme, I agree that creating a directory at launch and cleaning up in `deactivate` would solve the problem and without any extra risk in the first place. Thanks for weighing in. 

Thoughts, @ramya-rao-a?

ramya-rao-a:
Those are some great points @antong, thanks for bringing them up.

I have pushed a [change](https://github.com/Microsoft/vscode-go/pull/1912/commits/f9ab38824a9236635f6b4f7821627051ef10ce4f) to not re-use a persisted directory if it does not exist. That should take care of most of your concerns

@Charliekenney23 The comment is helpful, but I added a check regardless. Do take a look.

> creating a directory at launch and cleaning up in deactivate would solve the problem and without any extra risk in the first place.

Yes, lets do this if @antong finds any other risk with the change I made :)

antong:
What if the directory does exist and is writable, but is owned by somebody else (was autoremoved, but recreated by evil Bob)?

ramya-rao-a:
On a side note: @antong you are awesome

Charliekenney23:
Agreed! I've just updated the code.

Charliekenney23:
Just a note: There will be some edge cases where the tmp dir will not be deleted (for instance, when the user force quits vscode, the extension does not have time to call `deactivate` and delete the temp dir). Though, any stray temp dirs will not be reused.

In effort to minimize these stray directories, I'm only creating the temp directory as needed (when the `TempFileProvider.getFilePath` method is first called).

ramya-rao-a:
@Charliekenney23 I've refactored the code a bit to avoid using classes as the rest of the util.ts file doesn't use any. This is just to be consistent with the coding style in the file and to make it easier for future maintenance. Can you take a look?

Charliekenney23:
@ramya-rao-a  LGTM!

PR #1922, opened 2018-09-11
fix: statusbar Analysis Tools Missing text

svipben:
image


src/goStatus.ts
```diff
@@ -37,9 +37,8 @@ export function hideGoStatus() {
 
 export function showGoStatus(message: string, command: string, tooltip?: string) {
 	statusBarEntry = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, Number.MIN_VALUE);
-	statusBarEntry.text = message;
+	statusBarEntry.text = `$(alert) ${message}`;
```
ramya-rao-a: Cool

svipben: Thanks 😄

PR #1932, opened 2018-09-17
add support for the experimental version of gocode running on go/packages

stamblerre:
This PR adds support for installing and switching between the regular version of certain Go tools and the versions that run on-top of [go/packages](https://godoc.org/golang.org/x/tools/go/packages) (which works with Modules). 

If the user is in a directory with a go.mod file, `gocode-gomod` will be executed, whereas if the user is in a normal `$GOPATH` workspace, the normal version will be executed. This allows users to test out their usual Go tools with Modules, while avoiding any bugs that may still be present, since these tools are still quite new.

For now, only gocode switches between `gocode-gomod` and `gocode`. `godef-gomod` can be downloaded, but it's not yet added in this change. Ideally, the version of the tool to run should be decided in a function like `getBinPath`, but I wanted to test one tool before using that approach.


src/goInstallTools.ts
```diff
@@ -344,16 +357,22 @@ function installTools(missing: string[]) {
 					resolve([...sofar, null]);
 					return;
 				}
-				cp.execFile(goRuntimePath, ['get', '-u', '-v', allTools[tool]], { env: envForTools }, (err, stdout, stderr) => {
+				let args = ['get', '-u', '-v'];
+				if (tool.endsWith("-gomod")) {
+					args.push('-d');
```
ramya-rao-a: Why are we choosing to not install and instead run a `go build` later?

src/goSuggest.ts
```diff
@@ -145,159 +168,166 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
-			}
-
-			// Unset GOOS and GOARCH for the `gocode` process to ensure that GOHOSTOS and GOHOSTARCH
-			// are used as the target operating system and architecture. `gocode` is unable to provide
-			// autocompletion when the Go environment is configured for cross compilation.
-			let env = Object.assign({}, getToolsEnvVars(), { GOOS: '', GOARCH: '' });
-			let stdout = '';
-			let stderr = '';
+			let gocodeName = 'gocode';
+			hasModFile().then(goMod => {
```
ramya-rao-a: `ensureGoCodeConfigured` would be a better place for this. Have a variable `isGoMod` like `setGocodeOptions`, update it from inside `ensureGoCodeConfigured` and use it in `runGoCode`

ramya-rao-a: Also, the linter is failing

stamblerre:
It seems like the broken build is a result of an error in "fillStruct", which I don't think is related to this change.

stamblerre:
@ramya-rao-a - just wanted to follow up, does this change seem ready to you?

fatih:
I'm currently working on adding `gocode-gomod` as well, however, it only returns the candidates 20% of the time. I'm not sure what's going. I don't have much time this week (took a week off), but here is a WIP PR for vim-go: https://github.com/fatih/vim-go/pull/1988 

I'll check it again once I'm back in town.

stamblerre:
Thanks for merging @ramya-rao-a. I'll probably have a similar PR soon for you for godef. Let me know if there are any problems with it.

ramya-rao-a:
@stamblerre 

I am seeing similar things as @fatih.

I just have a simple file with just 1 non std lib package dependency. It worked well for the past hour, but now I get "null" as the output from `gocode-gomod`.

You can use the instructions at https://github.com/Microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension to install the beta version and test it out

ramya-rao-a:
@stamblerre 

I am seeing similar things as @fatih.

I just have a simple file with just 1 non std lib package dependency. It worked well for the past hour, but now I get "null" as the output from `gocode-gomod` all the time. It doesn't return results for std lib packages either at the moment

To test it out in VS Code:
- Download https://github.com/Microsoft/vscode-go/blob/master/Go-gocode-mod.vsix
- Reload VS Code

stamblerre:
I'm not sure if this will help fully, but I just pushed https://github.com/stamblerre/gocode/commit/b960ec24cf795e8590198eb337c6286d5488e399, which should allow regular gocode and gocode-gomod run simultaneously. I'll try to keep using it though to see if I can find more problems.

stamblerre:
I haven't been able to reproduce that yet, but I am using gocode-gomod right now and trying to see if I can get it to happen. If you run "gocode-gomod close" and then "gocode-gomod -s -debug", and then try a completion and see what it prints out, that might give some useful information. Thanks!

ramya-rao-a:
I think I know whats happening!!

In the above example, I am using the `quote` package. 

1. Say I used `quote.Hello()` and save the file. Subsequent `quote.` will give me completions.
2. Remove all `quote` related code and the import, save the file.
3. Add the `quote` package back to the import block. No completions for `quote.`

On every file save, we run `go build` which in turn updates the module cache.
In step 1 above, `quote` is now part of the module, and so completion works.
In step 2, it is no longer part of the module, and so no completions.

@fatih Do you see something similar?

stamblerre:
Oh, so is the only problem you're seeing related to completing with unsaved imports?

ramya-rao-a:
Yes, looks like that

stamblerre:
Ok, I will try to investigate this error and see if I can resolve it. Though this doesn't seem like a huge deal to me--the pretty simple solution is to add the import back.

src/goInstallTools.ts
```diff
@@ -344,16 +359,22 @@ function installTools(missing: string[]) {
...
+				if (tool.endsWith('-gomod')) {
+					args.push('-d');
+				}
+				args.push(allTools[tool]);
+				cp.execFile(goRuntimePath, args, { env: envForTools }, (err, stdout, stderr) => {
 					if (stderr.indexOf('unexpected directory layout:') > -1) {
 						outputChannel.appendLine(`Installing ${tool} failed with error "unexpected directory layout". Retrying...`);
-						cp.execFile(goRuntimePath, ['get', '-u', '-v', allTools[tool]], { env: envForTools }, callback);
+						cp.execFile(goRuntimePath, args, { env: envForTools }, callback);
+					} else if (tool.endsWith('-gomod')) {
```
Dexus: I'll never pass this in my tests. see #2024

ramya-rao-a: @Dexus Did you try the latest from master?

You need to update either of `gocode-gomod` or `godef-gomod` via the `Go: Install/Update Tools` command to hit this `else` block

Dexus: As in #2024 already pointed out I fix it myself with a copy. Would be nice if the current fix will be published asap. That works for me.

PR #1937, opened 2018-09-23
Add module support via gopkgs

uudashr:
Add module support. All the module dependencies handled by `gopkgs`.

### DIFF
--- a/src/goBrowsePackage.ts
+++ b/src/goBrowsePackage.ts
@@ -10,7 +10,6 @@ import cp = require('child_process');
 import path = require('path');
 import { getAllPackages } from './goPackages';
 import { getImportPath, getCurrentGoPath, getBinPath } from './util';
-import { isModSupported } from './goModules';
 
 export function browsePackages() {
 	let workDir = '';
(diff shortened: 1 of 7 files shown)


uudashr:
@ramya-rao-a not sure why it always failed on `gogetdoc` related, any idea?

uudashr:
Why do we need `isMod`? `gopkgs` already provide the module dependencies

ramya-rao-a:
For auto-completion support when using modules, we will be testing a fork of `gocode`. Therefore I need to know whether current file being edited is using modules or not. See the commented line https://github.com/Microsoft/vscode-go/blob/master/src/goSuggest.ts#L169

ramya-rao-a:
On a side note, I was not aware that you were working on getting module support from gopkgs itself. That's why I went ahead and did 
https://github.com/Microsoft/vscode-go/commit/0a9001f8fa2d3d69f5946f05cd23549abea458eb.

Am very glad to see the support from gopkgs directly.

Also, I wanted to invite you to join https://groups.google.com/forum/#!forum/golang-tools
Post Gophercon this year, this mailing list was created as a place to discuss the current modules/packages related work in the Go tooling system and how it affects editors.

uudashr:
Cool... all comes together

PR #1947, opened 2018-09-27
goMain.ts add commands to Go: Show All Commands

### LINKED ISSUE #1822: Add Go to implementation cmd to the Go: Show all Commands for better discoverability
Proposal. Please, implement a function that shows which interfaces from the current scope are satisfied by the selected function.

vimalk78:
commands added :
 * Go to definition
 * Go to implementation
 * Go to symbol in file
 * Go to symbol in workspace

fixes #1822

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -367,8 +367,29 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	}));
 
 	ctx.subscriptions.push(vscode.commands.registerCommand('go.show.commands', () => {
-		vscode.window.showQuickPick(getExtensionCommands().map(x => x.title)).then(cmd => {
-			let selectedCmd = getExtensionCommands().find(x => x.title === cmd);
+		let extCommands = getExtensionCommands()
+		extCommands.push({
+			command : 'editor.action.goToTypeDefinition',
+			title : 'Go : Go to Type Definition',
+			description : 'Go : invokes the GoDefinitionProvider'
+		})
+		extCommands.push({
+			command : 'editor.action.goToImplementation',
+			title : 'Go : Go to Implementation',
+			description : 'Go : invokes the GoImplementationProvider'
+		})
+		extCommands.push({
+			command : 'workbench.action.gotoSymbol',
+			title : 'Go : Go to Symbol in Document',
+			description : 'Go : invokes the GoDocumentSymbolProvider'
+		})
+		extCommands.push({
+			command : 'workbench.action.showAllSymbols',
+			title : 'Go : Go to Symbol in Workspace',
+			description : 'Go : invokes the GoWorkspaceSymbolProvider'
+		})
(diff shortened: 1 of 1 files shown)


src/goMain.ts
```diff
@@ -367,8 +367,29 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	}));
 
 	ctx.subscriptions.push(vscode.commands.registerCommand('go.show.commands', () => {
-		vscode.window.showQuickPick(getExtensionCommands().map(x => x.title)).then(cmd => {
-			let selectedCmd = getExtensionCommands().find(x => x.title === cmd);
+		let extCommands = getExtensionCommands()
+		extCommands.push({
+			command : 'editor.action.goToTypeDefinition',
+			title : 'Go : Go to Type Definition',
+			description : 'Go : invokes the GoDefinitionProvider'
```
ramya-rao-a: - This should be `editor.action.goToDeclaration` with title `Go to Definition`.
- We can skip the description. It doesnt get used in the `Show all commands` feature
- Let's not include the prefix `Go:` in the title. 
The prefix is for commands explicitly registered and implemented by the Go extension.
Where as these 4 commands that we are adding is registered in core VS Code an is only implemented by the Go extension

ramya-rao-a:
By the way, if you are interested in winning a t-shirt in the hactoberfest, close this PR and open a new one in October. See https://open.microsoft.com/2018/09/18/hacktoberfest-2018-microsoft/?WT.mc_id=hacktoberfest-twitter-beverst for more details

src/goMain.ts
```diff
@@ -367,8 +367,29 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+		})
+		extCommands.push({
+			command : 'editor.action.goToImplementation',
+			title : 'Go : Go to Implementation',
+			description : 'Go : invokes the GoImplementationProvider'
+		})
```
ramya-rao-a: Missing semicolon for each of the `push` is causing linting errors in the build

src/goMain.ts
vimalk78: 1. will change to `editor.action.goToDeclaration` . 
2. description i kept for sake of documentation. it was hard for me to see the connection between the various `Providers` in the vscode-go extension and their corresponding internal commands. thats why initially i started with creating a new command
3. ok will remove the 'Go:' prefix, else it may be thought of a separate command.
in fact i was looking for some api to get the complete command object for these four commands, so that title, description etc can be reused directly. i couldn't find so created the literals.

src/goMain.ts
vimalk78: thanks. i am new to type script.
is there a way to run such checks locally? before sending the PR

ramya-rao-a: Yes, you can run `npm run lint` in the terminal

PR #1961, opened 2018-10-03
Use more icons in autocomplete box

cezarsa:
This PR changes the autocomplete dialog to use more icons, distinguishing structs, interfaces and other types.

Before:
screen shot 2018-10-02 at 22 09 45

After:
screen shot 2018-10-02 at 22 08 13

The following code is used in the autocomplete example above:
```
type BoolType bool
const Const = 8
func Func() {}
type Interface interface{}
type StringType string
type Struct struct{}
var Var = 9
```

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -15,14 +15,20 @@ import { getTextEditForAddImport } from './goImport';
 import { getImportablePackages } from './goPackages';
 import { isModSupported } from './goModules';
 
-function vscodeKindFromGoCodeClass(kind: string): vscode.CompletionItemKind {
+function vscodeKindFromGoCodeClass(kind: string, type: string): vscode.CompletionItemKind {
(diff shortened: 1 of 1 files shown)


cezarsa:
So, the reasoning for Go Structs to use the Class icon was because Go structs might look a bit more like classes in other languages since they can have methods. Also, C++ structs use the Class icon as well.

Using Struct icons for other types was arbitrary really, I didn't know exactly what icon to use on them. After taking a look at how the C extension handles this I found out that it uses the Enum icon for `typedef` declarations. Maybe I should also switch other types to the Enum icon for consistency?

ramya-rao-a:
Ah, ok. That makes sense.
I dont have a preference for Enum icon over Struct icon, so let's go with what you have at the moment.
If we get any feedback around this, then we can act on it.

Issue #1973, opened 2018-10-07
File paths in the test output are not clickable when using Go modules


ramya-rao-a:
Steps to Reproduce:

1. In a project using Go modules, write a test that will fail
2. Run the test using any means: codelens or the test commands

The output contains only the file name and is not clickable:

![image](https://user-images.githubusercontent.com/16890566/46577040-3d8e2700-c990-11e8-991f-7af022d10658.png)

When not using Go modules, the output contains the absolute path of the file and is clickable

![image](https://user-images.githubusercontent.com/16890566/46577041-4f6fca00-c990-11e8-9b78-c3b0a845a054.png)

ramya-rao-a:
This is fixed in the latest update to the Go extension (0.6.91)

PR #1975, opened 2018-10-07
make file paths clickable in test output when using go modules

jrwren:


jrwren:
References: https://github.com/Microsoft/vscode-go/issues/1973

jrwren:
err... oops, I just noticed this isn't correct. I'll attempt to fix.

ramya-rao-a:
@jrwren I've already pushed a fix for this issue a few days ago and the fix is available in the latest update to the Go extension. Sorry about that.

In case, you are interested in the fix, see https://github.com/Microsoft/vscode-go/commit/ddd26356d535fe8653c0cf6619975b167b34f41e

Also, I dont believe the hacktoberfest label is needed. This will already be counted automatically as far as I know.

Regardless, if you want to pick up any other bug, you are most welcome.
Look for issues with the label "help-wanted" or "good-first-issue"

Issue #1966, opened 2018-10-04
gocode-gomod and godef-gomod fails to install on multi path gopath


AlexRouSg:
Steps to Reproduce:

1. Have a multi path gopath, e.g. `C:\folder1;C:\folder2; `
2. Run command Go: Install/Update Tools
3. Select all

Looks like the script is assuming gopath is a single path and blindly passing gopath around.

```
2 tools failed to install.

gocode-gomod:
Error: Command failed: C:\Go\bin\go.exe build -o C:\folder1;C:\folder2;\bin\gocode-gomod.exe github.com/stamblerre/gocode
go build github.com/stamblerre/gocode: mkdir C:\folder1;C:: The filename, directory name, or volume label syntax is incorrect.
go build github.com/stamblerre/gocode: mkdir C:\folder1;C:: The filename, directory name, or volume label syntax is incorrect.

godef-gomod:
Error: Command failed: C:\Go\bin\go.exe build -o C:\folder1;C:\folder2;\bin\godef-gomod.exe github.com/ianthehat/godef
go build github.com/ianthehat/godef: mkdir C:\folder1;C:: The filename, directory name, or volume label syntax is incorrect.
go build github.com/ianthehat/godef: mkdir C:\folder1;C:: The filename, directory name, or volume label syntax is incorrect.
```

Scapal:
This commit in `golang.org/x/tools/go/packages` introduces a breaking change:
https://github.com/golang/tools/commit/2f1727f1b37d7d257665debc72d4ae0b8e449748#diff-04792145c70ebd708928684da4875fe9R108

Both `github.com/ianthehat/godef` and  `github.com/stamblerre/gocode` depend on this.

AlexRouSg:
@godblesshugh That issue is unrelated

Delphier:
same issue +1

AlexRouSg:
Thought I'd tell everyone the workaround just incase no one figured it out yet. Set `go.toolsGopath` in the settings so it will then only have one gopath location.

margusha:
installation on mac also fails and setting go.toolsGopath did not help either. Seems the issue for me is on this line: 

https://github.com/Microsoft/vscode-go/blob/4abdf8f604c720a346bf7c0910588568a6a23807/src/goInstallTools.ts#L358

[...]

PR #1988, opened 2018-10-11
use last gopath element when installing tools

jrwren:
https://github.com/Microsoft/vscode-go/issues/1966

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -301,6 +301,8 @@ export function installTools(missing: string[]) {
 	// Else use the Current Gopath
 	let toolsGopath = getToolsGopath() || getCurrentGoPath();
 	if (toolsGopath) {
+		let paths = toolsGopath.split(path.delimiter);
+		toolsGopath = paths[0];
 		envForTools['GOPATH'] = toolsGopath;
 	} else {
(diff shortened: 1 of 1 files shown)


AlexRouSg:
Why use last? The `go get` behavior is to use the first.

jrwren:
> Why use last? The `go get` behavior is to use the first.

I thought so too, but that is not what I observed. Can you double check that you observe it installing to first?

AlexRouSg:
Yes, that is how I observe it to work on Linux and Windows. This is even documented behavior. 

https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies

> When checking out a new package, get creates the target directory GOPATH/src/<import-path>. If the GOPATH contains multiple entries, get uses the first one. For more details see: 'go help gopath'.

jrwren:
That is pretty definite documentation. I wonder why I think I observed what I observed. I'll change this.

I also had a question re: calling `go install`? Why use -o rather than calling `go install` with `GOPATH` set?

jrwren:
I realize now that the same code paths in vscode-go are made to install the tools even when there is no GOPATH, e.g. the GOPATH is infered per project or the go project uses go modules. I'm going to change the index to 0 and hopefully that is acceptable.

PR #1996, opened 2018-10-12
Code coverage removal fix for single line comments

### LINKED ISSUE #1852: Coverage UI not updating upon writing comments in the source code
This is both an issue and a feature request: 

- It's an issue because comments are inert by the code compiler, yet simply adding comments to the source code under test breaks the coverage highlighting feature. 
- It's a feature request, because, by [this logged issue](https://github.com/Microsoft/vscode-go/issues/1104#issuecomment-324063874), that feature breaking happens <q> because we can't guarantee that the change hasn't changed the test results </q>. Also, because my job here, and use case, is that I'm responsible for both documenting and testing my team's code. 

All I'm doing, for documentation of the source code, is **inserting line comments** before the function declarations.

karthikraobr:
This pull request partially fixes #1852.

### DIFF
--- a/src/goCover.ts
+++ b/src/goCover.ts
@@ -257,13 +257,15 @@ export function applyCodeCoverage(editor: vscode.TextEditor) {
  * @param e TextDocumentChangeEvent
  */
 export function removeCodeCoverageOnFileChange(e: vscode.TextDocumentChangeEvent) {
-	if (e.document.languageId !== 'go' || e.contentChanges.length < 1) {
+	if (e.document.languageId !== 'go' || !e.contentChanges.length || !isCoverageApplied) {
 		return;
 	}
-	if (isPartOfComment(e)) {
-		return;
-	}
+	
 	if (vscode.window.visibleTextEditors.every(editor => editor.document !== e.document)) {
+		return;
+	}
+
+	if (isPartOfComment(e)) {
 		return;
 	}
 
@@ -312,13 +314,14 @@ export function toggleCoverageCurrentPackage() {
 
 export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
 	return e.contentChanges.every(change => {
-		let line = e.document.lineAt(change.range.start);
-		let text = e.document.getText(line.range);
-		let idx = text.search('//');
-		let changeIdx = change.range.start.character;
-		if (idx === -1 || idx > changeIdx || !change.range.isSingleLine || change.text.includes('\n')) {
+		// We cannot be sure with using just regex on individual lines whether a multi line change is part of a comment or not
+		// So play it safe and treat it as not a comment
+		if (!change.range.isSingleLine || change.text.includes('\n')) {
 			return false;
 		}
-		return true;
+
+		const text = e.document.lineAt(change.range.start).text;
+		const idx = text.search('//');
+		return (idx > -1 && idx <= change.range.start.character);
 	});
 }
--- a/src/goCover.ts
+++ b/src/goCover.ts
@@ -314,11 +314,10 @@ export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
 	let result = false
 	e.contentChanges.every(function(change, index){
 		let line = e.document.lineAt(change.range.start)
-		let text = e.document.getText(line.rangeIncludingLineBreak)
+		let text = e.document.getText(line.range)
 		let idx = text.search('//')
 		let changeIdx = change.range.start.character
-		// Exit the loop as soon as one of the changes is not part of a comment.
-		if (idx != -1 && idx > changeIdx){
+		if (idx == -1 || idx > changeIdx || !change.range.isSingleLine || change.text.includes("\n") ){
 			result = false
 			return false;
 		}else{
@@ -326,14 +325,4 @@ export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
 		}
 	})
 	return result
-}
-
-export function isPartOfMultiLineComment(lineNumber: number, doc: vscode.TextDocument): boolean{
-	do{
-		let text = doc.lineAt(lineNumber).text
-		if (text.search("*/")){
-			return true
-		}
-	}while(lineNumber<doc.lineCount)
-	return false
 }
--- a/src/goCover.ts
+++ b/src/goCover.ts
@@ -260,13 +260,13 @@ export function removeCodeCoverageOnFileChange(e: vscode.TextDocumentChangeEvent
 	if (e.document.languageId !== 'go' || e.contentChanges.length < 1) {
 		return;
 	}
-	if(isPartOfComment(e)){
+	if (isPartOfComment(e)) {
 		return
 	}
 	if (vscode.window.visibleTextEditors.every(editor => editor.document !== e.document)) {
 		return;
 	}
-	
+
 	clearCoverage();
 }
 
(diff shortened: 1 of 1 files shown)


src/goCover.ts
```diff
@@ -305,3 +309,19 @@ export function toggleCoverageCurrentPackage() {
 		});
 	});
 }
+
+export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
+	let result = false
+	e.contentChanges.every(function(change, index){
+		let line = e.document.lineAt(change.range.start)
+		let text = e.document.getText(line.rangeIncludingLineBreak)
```
ramya-rao-a: What about the case when the range spans multiple lines and only the first line was part of a comment? In that case, `isPartOfComment` would return `true`

src/goCover.ts
```diff
@@ -305,3 +309,19 @@ export function toggleCoverageCurrentPackage() {
 		});
 	});
 }
+
+export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
+	let result = false
+	e.contentChanges.every(function(change, index){
+		let line = e.document.lineAt(change.range.start)
```
ramya-rao-a: What about the case when the range spans multiple lines and only the first line was part of a comment? In that case, `isPartOfComment` would return `true` instead of `false`

src/goCover.ts
```diff
@@ -305,3 +309,19 @@ export function toggleCoverageCurrentPackage() {
...
+		let text = e.document.getText(line.rangeIncludingLineBreak)
+		// Exit the loop as soon as one of the changes is not part of a comment.
+		if (!text.startsWith('//')){
```
ramya-rao-a: What about a case like the below?

`var a = 1 // this is a comment`

Here when the comment is edited, `isPartOfComment` returns `false` instead of `true`

src/goCover.ts
```diff
@@ -305,3 +309,19 @@ export function toggleCoverageCurrentPackage() {
...
+			result = false
```
ramya-rao-a: We dont need to use the `result` variable here. Instead, we can do the below

```
return e.contentChanges.every(change => {
// Just return true or false based on your logic here
}
```
The `.every` on an array stops looping over the array items the moment one of them returns `false` in the callback

src/goCover.ts
karthikraobr: For block comments of the form /* .... *,. I was thinking of the following solution

- Scan backwards from current change until start of the document to find /*. This would be necessary in cases where the block comment starts on a line after the code and changes are made to the code. In this case we would have to clear the coverage.  Example

`return 0, err /*
			one
			two
			three.
		*/`


- Scan forwards until */ is found or end of the document. 
- Check if change is within the above range.

Seems very inefficient to me.  Is there a better way of doing it? Perhaps a vscode API to identify block comments?

src/goCover.ts
karthikraobr: Fixed. Will update the PR once I have fixed the block comment issue.

src/goCover.ts
karthikraobr: Fixed. Will update the PR once I have fixed the block comment issue.

src/goCover.ts
ramya-rao-a: Here I wasnt referring to block comments. Take the below case
```
// I am a comment
var a = 1
var b = 2
```
Now if I start selecting from the word `comment` and continue till the next line and delete the selected text, then the `line` variable in your code will have `// I am a ` and `isPartOfComment` will return `true` even though some code was deleted.

karthikraobr: I have updated the PR to resolve this issue.

karthikraobr:
Apologies for the multiple lint commits. Go dev here :) Hence the missing semi-colons.

src/goCover.ts
```diff
@@ -305,3 +309,20 @@ export function toggleCoverageCurrentPackage() {
 		});
 	});
 }
+
+export function isPartOfComment(e: vscode.TextDocumentChangeEvent): boolean {
+	let result = false;
+	e.contentChanges.every(function (change, index) {
+		let line = e.document.lineAt(change.range.start);
```
ramya-rao-a:
```
return e.contentChanges.every(change => {
// Just return true or false based on your logic here
}
```

> I guess it would be a good idea not to return as soon as we find that the change is a part of a comment. If another change which is not a part of a comment is found in a later iteration, then we would have to anyway remove the highlighting right?

A `return` statement in the callback passed to the `.every` doesnt cause a return from `isPartOfComment`. 
A `return true` statement in the callback passed to the `.every` cause the iteration to continue;
A `return false` statement in the callback passed to the `.every` cause the iteration to stop;

ramya-rao-a: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

src/goCover.ts
```diff
@@ -305,3 +309,20 @@ export function toggleCoverageCurrentPackage() {
...
+		let text = e.document.getText(line.range);
+		let idx = text.search('//');
+		let changeIdx = change.range.start.character;
+		if (idx === -1 || idx > changeIdx || !change.range.isSingleLine || change.text.includes('\n')) {
```
ramya-rao-a: Can you add some tests to cover these cases?

karthikraobr: Sure!

src/goCover.ts
karthikraobr: Thanks for the detailed explanation.

src/goCover.ts
karthikraobr: Could you please give me some pointers on how to write tests. I ran the tests and seems like currently there are no tests for code coverage.  I am assuming these would be the steps I would have to follow : 

- Create a go file with code and comments to cover all scenarios. 
- Write some tests for the above file and run them. (use [testCurrentFile](https://github.com/Microsoft/vscode-go/blob/d789de0f9561a1c9aa979a145cb64bcbfbe185cf/src/goTest.ts#L172) function).
- Select some part of the code and delete/add (based on the scenarios in the if condition using [applyUsingTextEditorEdit](https://github.com/Microsoft/vscode-go/blob/d789de0f9561a1c9aa979a145cb64bcbfbe185cf/src/diffUtils.ts#L49) ).
- And check if coverage highlights have been removed or retained. (How do I achieve this?)

karthikraobr:
In case of a deletion, the changed text is always "".  Is there some way to get the deleted text, rather than an empty string? This is necessary in cases where multiple lines are deleted. How would I differenciate between deleting line 1&2 and deleting line 2&3.  
```
1.// hello world
2.// hello again
3.func sayHello()
```

ramya-rao-a:
This will need some more time from my side to look into, I'll get back to this next week.

karthikraobr:
> This will need some more time from my side to look into, I'll get back to this next week.

@ramya-rao-a  could you please let me know if there has been an update on this?

ramya-rao-a:
@karthikraobr Thanks for your patience, I'll get right back to this as soon as I release the current update which should be sometime this week.

karthikraobr:
hi @ramya-rao-a , any update on this? Thanks and sorry for the repeated reminders :)

karthikraobr:
@ramya-rao-a any update on this and I apologise in advance :)

ramya-rao-a:
@karthikraobr No apologies needed :) You have every right to give the reminders.

I have been a little busy at work from my end, and so couldnt get back sooner. I'll definitely take a look soon

ramya-rao-a:
I have refactored the code a bit to avoid checking for comments when there is no coverage applied in the first place and also the code for `isPartOfComment` as well

PR #2015, opened 2018-10-16
Remove env mask on GOOS and GOARCH

eternal-flame-AD:
Removed to allow suggestions on platform-specific syscall like `syscall/js`

<s>WORK IN PROGRESS! Still undecided on how to resolve packages without precompiled objects.</s>

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -187,7 +187,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				return reject();
 			}
 
-			let env = Object.assign({}, getToolsEnvVars());
+			let env = getToolsEnvVars();
 			let stdout = '';
 			let stderr = '';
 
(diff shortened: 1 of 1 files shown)


stamblerre:
If the user hasn't built their dependencies, then we really can't do better than using the `-source` flag. I've actually recently added a change to gocode that causes it to fall back to using `-source` if it can't find dependencies (https://github.com/mdempsky/gocode/commit/a07fefd98fdcd9d2c16aa3f61d8e89eda396bf99), so perhaps if you rebuild the latest version, that will work for you.

ramya-rao-a:
@eternal-flame-AD Can you run `Go: Install/Update Tools`, select `gocode` and press `Ok` to update it? Then try vscode with your changes and see if it works without needing to specifically installing the imports with the other GOOS?

eternal-flame-AD:
Yes, I think the problem is resolved now with the updated `gocode`:)

ramya-rao-a:
@eternal-flame-AD Last question: The changes in this PR are also needed for the problem to be fixed, correct? Or only updating `gocode` was enough?

eternal-flame-AD:
Yes, the change is needed because otherwise there would be of no way `gocode` could get the build context.

Issue #1990, opened 2018-10-11
Can't copy value from the Debugger Variables list (Go/Delve)


altimac:
Issue Type: Bug
I'm using the Go delve debugger (remote mode) inside VSCode and right clicking any of the local variables in the debugger (displaying the correct variable values!) only displays a disabled contextual menu where I can't copy (or set) value.

Copying a variable value from the debugger is pretty basic debugging workflow.

Thanks
<img width="445" alt="screen shot 2018-10-11 at 11 26 17" src="https://user-images.githubusercontent.com/778986/46803165-b827bf80-cd5f-11e8-88d0-35af1242ef1b.png">

ramya-rao-a:
Thanks for reporting @altimac 

@roblourens Do you know what needs to be done to enable the context menu entries here?

roblourens:
@isidorn

altimac:
I reported the bug at the wrong place first. @isidorn replied:

> Please file this agains the go and delve debugger and tell them they have to support evaluateName property for this actions to be enabled
>

isidorn:
@ramya-rao-a yeah, please support the [`evaluateName`](https://github.com/Microsoft/vscode/blob/5d645696cc669e4bc9c8824749395c7112e8c845/src/vs/workbench/parts/debug/common/debugProtocol.d.ts#L1387) property.

Example: in the casses in the picture the `evaluateName` is equal to `name`, however inside the `zone` object if it had the `id` property the evaluate name of that would be `zone.id`
So `evaluateName` is the full path to get to the variable

ramya-rao-a:
Thanks @isidorn 

Marking this as `help-wanted`. PRs are most welcome.

Code Pointers:
- Get started with a [local set up of the Go extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- [Debug the debugger](https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter)
- Update [variableRequest](https://github.com/Microsoft/vscode-go/blob/0.6.91/src/debugAdapter/goDebug.ts#L958) to add the `evaluateName` property on the variables

ramya-rao-a:
This feature is now implemented with https://github.com/Microsoft/vscode-go/pull/2020

[...]

PR #2020, opened 2018-10-17
Fix for #1990

karthikraobr:
Fix for #1990

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -809,14 +809,7 @@ class GoDebugSession extends DebugSession {
 			}
 			const locals = this.delve.isApiV1 ? <DebugVariable[]>out : (<ListVarsOut>out).Variables;
 			verbose('locals', locals);
-			locals.every(local => {
-				local.fqn = local.name;
-				local.children.every(child => {
-					child.fqn = local.name;
-					return true;
-				});
-				return true;
-			})
+			this.addFullyQualifiedName(locals);
 			let listLocalFunctionArgsIn = { goroutineID: this.debugState.currentGoroutine.id, frame: args.frameId };
 			this.delve.call<DebugVariable[] | ListFunctionArgsOut>('ListFunctionArgs', this.delve.isApiV1 ? [listLocalFunctionArgsIn] : [{ scope: listLocalFunctionArgsIn, cfg: this.delve.loadConfig }], (err, outArgs) => {
 				if (err) {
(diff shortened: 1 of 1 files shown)


karthikraobr:
@ramya-rao-a Could you have a look at this when you get some time. Thanks!

karthikraobr:
@ramya-rao-a  Fixed it.

src/debugAdapter/goDebug.ts
```diff
@@ -816,8 +825,17 @@ class GoDebugSession extends DebugSession {
 				}
 				const args = this.delve.isApiV1 ? <DebugVariable[]>outArgs : (<ListFunctionArgsOut>outArgs).Args;
 				verbose('functionArgs', args);
+				args.every(local => {
```
karthikraobr: I repeated this instead of creating a function because of the const and my understanding is that const variables can't be reassigned right? js noob here.

src/debugAdapter/goDebug.ts
```diff
@@ -922,6 +943,13 @@ class GoDebugSession extends DebugSession {
 					variablesReference: 0
 				};
 			} else {
+				if(v.children[0].children.length > 0){
```
karthikraobr: This is also a repetition since there is a ternary operator doing the same. Is there a cleaner way to do this?

src/debugAdapter/goDebug.ts
ramya-rao-a: Because of which `const`? The one on `args`?
The const only means that you cannot re-assign the array object. But it doesnt stop you from iterating over the contents and changing them.

src/debugAdapter/goDebug.ts
```diff
@@ -979,15 +1008,18 @@ class GoDebugSession extends DebugSession {
...
 				});
 			}
 		} else {
 			variables = vari.children.map((v, i) => {
 				let { result, variablesReference } = this.convertDebugVariableToProtocolVariable(v, i);
+				v.fqn = v.fqn == undefined ? vari.fqn + '.' + v.name : v.fqn;
 				return {
 					name: v.name,
 					value: result,
+					evaluateName: v.fqn == undefined ? vari.fqn + '.' + v.name : v.fqn,
```
ramya-rao-a: Cant we just use `v.fqn` here as it already has the right value?

src/debugAdapter/goDebug.ts
```diff
@@ -157,6 +157,7 @@ interface DebugVariable {
 	cap: number;
 	children: DebugVariable[];
 	unreadable: string;
+	fqn: string;
```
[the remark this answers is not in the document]
karthikraobr: Fully Qualified Name.Should I change the field name to Fully Qualified Name?

karthikraobr:
@ramya-rao-a thanks for the review comments. I have incorporated most of your comments. Lemme know if I can do something better?

karthikraobr:
@ramya-rao-a could you have a look please :)

PR #2025, opened 2018-10-18
Installation of godoc fails due to missing key/value in allTools

### LINKED ISSUE #2024: try to install gocode-gomod and godef-gomod but output panel stacking there
```
Installing 2 tools at /home/xxxx/go/bin
  gocode-gomod
  godef-gomod
```

karthikraobr:
Partially fixes #2024


karthikraobr:
@ramya-rao-a I see that godoc has been removed in [this commit](https://github.com/Microsoft/vscode-go/commit/a3f43e9e1575f740e885832119fcfdfd0f2076af#diff-3f57b095642b0ee445d6cdbc4f60d751). I guess we would have to remove it from [installAllTools](https://github.com/Microsoft/vscode-go/blob/master/src/goInstallTools.ts#L166). I will revert my previous commit.

PR #2027, opened 2018-10-19
Fix for issue 1836

### LINKED ISSUE #1836: Attempt to get absolute path on every line of test output causes wrongful paths
Plugin version: 0.6.85
Go version: 1.10.3

Just a simple file, example_test.go:
```go
// ...
logger := log.New(os.Stdout, "", log.Lshortfile)
logger.Println("Hello!")
// output: example_test.go:10: Hello
// I miss a "!" so that it will show FAIL when running test
```

This is the output by clicking `run package tests` from plugin:
log_shortfile_1

And this is the output from terminal directly, also what I expected:
log_shortfile_2

Any idea?

MagnumOpus21:
Fixes #1836.
@ramya-rao-a 
These are the images I get after making the modifications.
This image is that of a passing test case. 
pass_golang
This image is for a failing test case.
fails_golang
This is for one passing case and one failing case. (Package wide)
onepass.
I have tested this with all the options from the command palette. 
Do tell me if I need to add more to the PR, to make it even better :)


MagnumOpus21:
Ah my bad. I thought the "prepended" by --- FAIL : was a mistake. I shall rectify this in the following commit.

src/testUtils.ts
```diff
@@ -340,7 +340,7 @@ export function cancelRunningTests(): Thenable<boolean> {
 function expandFilePathInOutput(output: string, cwd: string): string {
 	let lines = output.split('\n');
 	for (let i = 0; i < lines.length; i++) {
-		let matches = lines[i].match(/^\s*(.+.go):(\d+):/);
+		let matches = lines[i].match(/^\s*(\s*\w+.go):(\d+):/);
```
ramya-rao-a: Why do we want to capture the spaces here? Is this to cater to the case when the file name contains a space?

MagnumOpus21: I thought spaces might be present at the beginning of the file. This has another bug, where if we have spaces in the filename like `file name.go` the regex will fail to pick this. Another commit inc.

ramya-rao-a:
@MagnumOpus21 Are you sure the last [commit ](https://github.com/Microsoft/vscode-go/pull/2027/commits/9f00d7bb194f40d3c8ead61c365aa45107fe130e) makes a difference?

File names with spaces would have worked even without that commit as far as I understand.

Since that is a very corner case, I am ok not supporting the scenario where file name has spaces.
I am more concerned about https://github.com/Microsoft/vscode-go/pull/2027#discussion_r226883288

MagnumOpus21:
The GitHub outage prevented me from viewing your previous message with the regex for the build errors. And I couldn't see the entire thread until now. Still am not sure if I can see all the messages on this PR 😅.

MagnumOpus21:
@ramya-rao-a Now it should capture all the failures on non-Windows platforms as well. I was just testing the extension on Windows. :sweat:

src/testUtils.ts
```diff
@@ -340,7 +340,7 @@ export function cancelRunningTests(): Thenable<boolean> {
 function expandFilePathInOutput(output: string, cwd: string): string {
 	let lines = output.split('\n');
 	for (let i = 0; i < lines.length; i++) {
-		let matches = lines[i].match(/^\s*(.+.go):(\d+):/);
+		let matches = lines[i].match(/^\s*([(\.\/)?\w+\s*]+.go):(\d+):/);
```
ramya-rao-a: Things inside the `[]` dont need the quantifiers like `?`, `+` or `*`. `[]` defines the allowed characters not their quantity or order.

Also, this will not work for Windows.

Can you make this an exported function and add some tests to cover the different scenarios?

You can add the tests in https://github.com/Microsoft/vscode-go/blob/master/test/unit/utils.test.ts

MagnumOpus21:
Sorry for the no progress during the past weeks. Hectic school session in progress. I will finish the PR during Thanksgiving week. Thanks for your patience.

ramya-rao-a:
No problem @MagnumOpus21, take your time.

peergynt:
@MagnumOpus21 did you see my comment in the original issue (#1836) by any chance?
It seems like checking that the line starts with at least a tab or four spaces could help.

MagnumOpus21:
I'm sorry 😔. I was busy with school. I'll take a look at this and finish this fast. It's been far too long. Apologize for the lateness @peergynt

MagnumOpus21:
@ramya-rao-a Do I have to do anything to get the Go Extension tests to pass for go1.11. It passes for the other versions of golang on Travis.

ramya-rao-a:
Hey @MagnumOpus21 

My apologies for not getting to this sooner 

I tried a  few more cases with the code in this PR
- `some random things C:/my/folder/fine.go:21:` fails as expected because the path is already expanded
- `some random things /my/folder/fine.go:21:` does not fail even though the path is an absolute path in non Windows machines.

I am starting to feel that we are getting into risky areas by trying to cater to extract the right relative path that needs to be expanded.

We know what the expanded path needs to look like because we have access to `cwd`, so I am leaning towards checking the existence of `cwd` in the output line should be taken as a sign as to not to bother expanding the file path.

Thoughts?

ramya-rao-a:
Hey @MagnumOpus21 

Since working with regexes still left some corner cases not covered, I've decided to take a different approach here. Please see https://github.com/microsoft/vscode-go/issues/1836#issuecomment-604250317

Thanks for all the work here & Happy Coding!

PR #2032, opened 2018-10-20
Drop support for go 1.8 from TravisCI builds

cristim:
The Travis builds are failing on Go 1.8, and this is shown on the Github badge as a build failure. 

The failures are because one of the dependencies seems to need type aliasing support available only as of Go 1.9: https://travis-ci.org/Microsoft/vscode-go/builds/444157465


ramya-rao-a:
@stamblerre brings up a good point in https://github.com/mdempsky/gocode/issues/73#issuecomment-431616650. We can use `nsf/gocode` for Go versions < 1.10

@cristim Would you be interested in looking into that?

Code pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension) to have a local set up of this extension up and running
- Instead of directly accessing `allTools` in `goInstallTools.ts`, have a function that accepts `tool` and `goVersion`. This way we can use the nsf/gocode for older versions of Go

ramya-rao-a:
I've made the changes to fallback to nsf/gocode for Go 1.8

Issue #2039, opened 2018-10-23
Add option to run long-running linters manually


Xjs:
I tried using gometalinter with many linters enabled in my day-to-day, but I found linting takes very long to complete. It would be great if I could trigger a full linting run (e. g. using gometalinter with many slow linters) manually, but have a fast linter (e. g. golint) set for lint-on-save.

ramya-rao-a:
Manual linting can be triggered using the command `Go: Lint Current Package` or `Go: Lint Workspace`.
These commands will use the settings from `go.lintFlags` just like when the linter is run on file save.

One way to achieve what you want is to update the `go.lintFlags`, run the above command. Then change the flags back.

Another way is to use [tasks](https://code.visualstudio.com/Docs/editor/tasks) which you can set up any way you want to use at any time you want

Xjs:
Reading https://github.com/Microsoft/vscode/issues/1435, I'm not sure if the support for global tasks is really that high on the vscode agenda. I put together a small patch that essentially duplicates the lint command to provide a global »Slow Lint« that can e. g. be used for gometalinter, and will send a PR. Feel free to discard it if you don't like it, but I'm going to use it for the time being. :)

Issue #1451, opened 2018-01-10
Manually run metalinter


wayneashleyberry:
I've found `gometalinter` to be too slow and sometimes too annoying to be set as my default linter, but I do like running it occasionally to see what it reports. Would it be possible to register a command in the command palette to run `gometalinter` manually?

canercidam:
I've been having a similar problem lately with `gometalinter`. Every time I hit on save after I code a bit, all of my 4 cores reach 100% and memory usage jumps by 1-1.5 GB for like 5 seconds. In #1373, a useful suggestion was setting `go.lintOnSave` to `false` and using `Go: Lint Current Package` command. So my workaround to this issue turned out to be as following:

`keybindings.json`:
```json
...
{
    "key": "ctrl+l",
    "command": "go.lint.package"
}
...
```

However, after doing this, `Output` window annoyingly started showing up every time I hit `Ctrl+L` so I also had to toggle it afterwards. I ended up combining the command with toggle by using `macros` extension which can be found [here](https://marketplace.visualstudio.com/items?itemName=geddski.macros):

`settings.json`:
```json
...
"macros": {
    "goLint": [
        "go.lint.package",
        "workbench.action.output.toggleOutput"
    ]
}
...
```

`keybindings.json`:
```json
...
{
    "key": "ctrl+l",
    "command": "macros.goLint"
}
...
```

ramya-rao-a:
@wayneashleyberry Like @canercidam has suggested you can use the command `Go: Lint Current Package` to lint manually and disable the lint on save feature in the settings.

@canercidam I have logged #1456 for the issue of the output pane popping up

wayneashleyberry:
Hi @ramya-rao-a, although that is a possible workaround I might have explained my request poorly. I want to run `golint` on save, that default behaviour is perfect for me. That works 99% of the time, but every now and then I want to run the metalinter just to see what it finds. That's why I wanted it as a separate command, so I could keep the default behaviour and just invoke the metalinter when needed. I'll carry on doing it from the command-line though, not a big deal :) Thanks for all your great work on this plugin!

ramya-rao-a:
@wayneashleyberry You can have both. You can keep the default behavior of linting on save and run the command `Go: Lint Current Package` from the command pallet whenever you want.

canercidam:
@wayneashleyberry Hmm, so it should use `golint` on save and command should be able to use `gometalinter`? Which means different values for `go.lintTool` in `settings.json`.

[...]

PR #2048, opened 2018-10-25
Duplicate Lint command into 'Slow Lint' (e. g. for manual gometalinter)

Xjs:
As discussed in issues like #2039, #1451: Some users have the need to occasionally and manually run a different linter than the `lintOnSave` linter. This patch essentially doubles the Lint command with a separate config setting `slowLintTool` that can be used as a manual override, using the same problemMatcher and thereby displaying results from the slow lint tool in the Problems pane as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -524,7 +524,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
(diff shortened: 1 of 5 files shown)


ramya-rao-a:
Closing this PR as it hasnt seen any activity in over half a year

Issue #1229, opened 2017-09-19
Allow debugging of different files with main function from the same directory


MZhoume:
Thanks again for this amazing extension. 

Regarding the debugging adapter, it would be nice to be able for us to debug a single `.go` file.

```typescript
let dlvArgs = [mode || 'debug'];
if (mode === 'exec') {
	dlvArgs = dlvArgs.concat([program]);
}
dlvArgs = dlvArgs.concat(['--headless=true', '--listen=' + host + ':' + port.toString()]);
if (launchArgs.showLog) {
	dlvArgs = dlvArgs.concat(['--log=' + launchArgs.showLog.toString()]);
}
if (launchArgs.cwd) {
	dlvArgs = dlvArgs.concat(['--wd=' + launchArgs.cwd]);
}
if (launchArgs.buildFlags) {
	dlvArgs = dlvArgs.concat(['--build-flags=' + launchArgs.buildFlags]);
}
if (launchArgs.init) {
	dlvArgs = dlvArgs.concat(['--init=' + launchArgs.init]);
}
if (launchArgs.backend) {
	dlvArgs = dlvArgs.concat(['--backend=' + launchArgs.backend]);
}
if (launchArgs.args) {
	dlvArgs = dlvArgs.concat(['--', ...launchArgs.args]);
}
```

The `program` is only evaluated if the mode is set to `exec`, but not `debug`. However, `delve` supports debug specific file: `del debug single.go`. 

Please let me know if that can be implemented. Thanks in advance!

ramya-rao-a:
When mode is not exec and the program points to a file and not directory, `dlv debug` is run on the parent directory of the file. Wont that work for your single file case?

MZhoume:
@ramya-rao-a Thanks for your reply.

Unfortunately no. Let's think of a situation where we have a `scripts` folder for all the automation scripts. Each `.go` file should be able to run individually so they all are in the `main` package and with a `main` function. And you can just use `go run script.go` to run a single file.

Since they are all in the same folder, and the current `delve` debugger adapter does not support debugging single file. When debugging is needed, you will have to move the file to a different folder without any other go script in order for the debugger to work. 

I am also taking this class where they put all the executable `go` files in one `main` folder (I assume it has something to do with the folder structure restriction `go` is enforcing). Not being able to debug only a single file make it impossible to work with the class assignments as well.

ramya-rao-a:
good point

ramya-rao-a:
[...]

PR #2051, opened 2018-10-25
Add support for debugging single files

chrislewisdev:
Address #1229 

In cases where you might have multiple files defining a `main` function in the same directory, this change allows the debugger to detect if you are trying to debug just a single file directly, rather than having to debug an entire directory (which fails because there are multiple `main` functions).

Should work in all situations that the `launch.json` config for the workspace has the `program` property set to `${file}`.

I'm open to any and all suggestions on how this could be improved if required :)

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -262,7 +262,7 @@ function fileContainsEntryPoint(filePath: string) {
 		);
 	}
 	catch (e) {
-		//If go-outline failed, we are probably not in a position to debug this file.
+		//If go-outline failed or it's not installed, we can't assume this file is executable.
 		return false;
 	}
 }
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -9,7 +9,7 @@ import { DebugProtocol } from 'vscode-debugprotocol';
 import { DebugSession, InitializedEvent, TerminatedEvent, ThreadEvent, StoppedEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles } from 'vscode-debugadapter';
 import { existsSync, lstatSync } from 'fs';
 import { basename, dirname, extname } from 'path';
-import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
+import { spawn, ChildProcess, execSync, spawnSync, execFile, execFileSync } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, resolveHomeDir, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
 import * as logger from 'vscode-debug-logger';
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -245,6 +245,28 @@ function normalizePath(filePath: string) {
 	return filePath;
 }
 
+function fileContainsEntryPoint(filePath: string) {
+	let goOutline = getBinPathWithPreferredGopath('go-outline', []);
+	let goOutlineFlags = ['-f', filePath];
+
+	try {
+		let goOutlineOutput = execFileSync(goOutline, goOutlineFlags);
+		let declarations = JSON.parse(goOutlineOutput) as { label, type, children }[];
```
chrislewisdev: I opted for an anonymous type here since we don't need the full `GoOutlineDeclaration` information as defined in `goOutline`. An alternative if we wanted to keep the type information in sync for future updates could be moving `GoOutlineDeclaration` and `GoOutlineRange` into a separate file to share with `goDebug`.

ramya-rao-a:
Thanks for the PR @chrislewisdev and my apologies for not getting to this PR sooner.
It somehow got lost in the sea of PRs that we received during Hacktoberfest.

@jhendrixMSFT I recall we recently discussed the problem this PR is trying to resolve and had come up with a solution that would avoid parsing the file and look for the `main` function. Do you recall what we landed on?

chrislewisdev:
That's ok @ramya-rao-a :) I noticed the builds failed on this PR but decided not to worry about it until I received some feedback. I'll keep an eye on this PR to see what conclusions you come to before making any further changes!

jhendrixMSFT:
@ramya-rao-a the context was a bit different, the solution was to pass the file name to dlv, e.g. `dlv debug main2.go`

ramya-rao-a:
> the solution was to pass the file name to dlv, e.g. dlv debug main2.go

And if the file being passed does not have the `main` function, then delve would throw an appropriate error?

jhendrixMSFT:
That's correct, dlv will give the following error and return a non-zero exit code.
```
Can not debug non-main package
```

ramya-rao-a:
Thanks @jhendrixMSFT 

@chrislewisdev Here is the way we want to go forward

- When the `program` attribute points to a file, pass this file to the `delve debug` command instead of passing the import path to the package like we do today
- When the `program` attribute points to a directory, we should do what we do today i.e pass the import path to the package to `dlv debug`
- Update the default value for `program` attribute in `goDebugConfiguration.ts` file to be `{$file}` instead of `${fileDirname}`

chrislewisdev:
Thanks @ramya-rao-a for the update - unfortunately I haven't had much time to revisit this recently but I hope to be able to do so in the coming weeks. In the meantime, if anyone would fancy having a crack at it themselves, feel free.

ramya-rao-a:
Thanks for letting us know about that @chrislewisdev 

In that case, I'll close this PR.

I have also updated the original issue with our latest approach. See 
https://github.com/Microsoft/vscode-go/issues/1229#issuecomment-473731132

PR #2054, opened 2018-10-26
Add item documentation to completions

### LINKED ISSUE #194: Godoc not appearing in code-completion
Hi Luke.

System:
- vscode-go 0.6.26
- Visual Studio Code 0.10.6.
- Windows 10 (64-bit)

Godoc does not appear when using code-completion.

In this animation we have some code written on the right pane which calls a function from the code on the left pane. You see that the func `GetGroupIDsOfUser` has a comment, however, it does not appear in the code-completion:
animation

Godoc should appear when:
- the cursor is inside the parentheses of a function call
- the auto-completion list is open with suggestions of functions

The godocs do not show up on my own code, however, when using the `fmt` package as an example, it only shows itself sometimes:
animation2

segevfiner:
Using `go doc` instead of `godoc` due to https://github.com/golang/go/issues/25443 (plus `godoc` doesn't show doc for unexported identifiers nor for anything in a main package)

`go doc` doesn't seem to support giving the code of an unsaved file via stdin like `gocode`, so it won't display documentation for unsaved changes.

Fixes #194 

cc @ramya-rao-a

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -90,7 +90,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 		const env = getToolsEnvVars();
 		const cwd = path.dirname(item.fileName);
 
-		return new Promise<vscode.CompletionItem>((resolve, reject) => {
+		return new Promise<vscode.CompletionItem>(resolve => {
 			const args = ['doc', '-c', '-cmd', '-u'];
 			if (item.package) {
 				args.push(item.package, item.label);
@@ -99,9 +99,10 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 				args.push(item.label);
 			}
 
-			cp.execFile(goRuntimePath, args, { cwd, env }, (err, stdout, stderr) => {
+			cp.execFile(goRuntimePath, args, { cwd, env }, (err, stdout) => {
 				if (err) {
-					reject(err);
+					console.log(err);
+					return resolve(item);
 				}
 
 				let doc = '';
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -41,8 +41,14 @@ function vscodeKindFromGoCodeClass(kind: string, type: string): vscode.Completio
 
 interface GoCodeSuggestion {
 	class: string;
+	package: string;
 	name: string;
 	type: string;
+}
+
+class ExtendedCompletionItem extends vscode.CompletionItem {
+	package: string;
+	document: vscode.TextDocument;
 }
 
 const lineCommentRegex = /^\s*\/\/\s+/;
@@ -73,6 +79,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 			}
 			return result;
 		});
+	}
+
+	public resolveCompletionItem(item: vscode.CompletionItem, token: vscode.CancellationToken): vscode.ProviderResult<vscode.CompletionItem> {
+		if (item instanceof ExtendedCompletionItem) {
+			const goRuntimePath = getBinPath('go');
+			if (!goRuntimePath) {
+				vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
+				return;
+			}
+
+			const env = getToolsEnvVars();
+			const cwd = path.dirname(item.document.fileName);
+
+			return new Promise<vscode.CompletionItem>((resolve, reject) => {
+				let args = ['doc', '-c', '-cmd', '-u'];
+				if (item.package) {
+					args.push(item.package, item.label);
+				} else {
+					args.push(item.label);
+				}
+
+				cp.execFile(goRuntimePath, args, { cwd }, (err, stdout, stderr) => {
+					if (err) {
+						reject(err);
+					}
+
+					let doc = '';
+					const goDocLines = stdout.toString().split('\n');
+					// i = 1 to skip the func signature line
+					for (let i = 1; goDocLines[i].startsWith('    '); i++) {
+						doc += goDocLines[i].substring(4) + '\n';
+					}
+
+					item.documentation = doc;
+					resolve(item);
+				});
+			});
+		}
 	}
 
 	public provideCompletionItemsInternal(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken, config: vscode.WorkspaceConfiguration): Thenable<vscode.CompletionItem[] | vscode.CompletionList> {
@@ -228,9 +272,11 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 					if (results && results[1]) {
 						for (let suggest of results[1]) {
 							if (inString && suggest.class !== 'import') continue;
-							let item = new vscode.CompletionItem(suggest.name);
+							let item = new ExtendedCompletionItem(suggest.name);
 							item.kind = vscodeKindFromGoCodeClass(suggest.class, suggest.type);
 							item.detail = suggest.type;
+							item.package = suggest.package;
+							item.document = document;
 							if (inString && suggest.class === 'import') {
 								item.textEdit = new vscode.TextEdit(
 									new vscode.Range(
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -82,41 +82,40 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 	}
 
 	public resolveCompletionItem(item: vscode.CompletionItem, token: vscode.CancellationToken): vscode.ProviderResult<vscode.CompletionItem> {
-		if (item instanceof ExtendedCompletionItem) {
-			const goRuntimePath = getBinPath('go');
-			if (!goRuntimePath) {
-				vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
-				return;
+		const goRuntimePath = getBinPath('go');
+		if (!goRuntimePath || !(item instanceof ExtendedCompletionItem)) {
+			return;
+		}
+
+		const env = getToolsEnvVars();
+		const cwd = path.dirname(item.fileName);
+
+		return new Promise<vscode.CompletionItem>((resolve, reject) => {
+			const args = ['doc', '-c', '-cmd', '-u'];
+			if (item.package) {
+				args.push(item.package, item.label);
+			} else {
+				// item.package would be empty for symbols from the current package
+				args.push(item.label);
 			}
 
-			const env = getToolsEnvVars();
-			const cwd = path.dirname(item.fileName);
-
-			return new Promise<vscode.CompletionItem>((resolve, reject) => {
-				let args = ['doc', '-c', '-cmd', '-u'];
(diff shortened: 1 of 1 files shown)


src/goSuggest.ts
```diff
@@ -41,10 +41,16 @@ function vscodeKindFromGoCodeClass(kind: string, type: string): vscode.Completio
...
 interface GoCodeSuggestion {
 	class: string;
+	package: string;
 	name: string;
 	type: string;
 }
 
+class ExtendedCompletionItem extends vscode.CompletionItem {
+	package: string;
+	document: vscode.TextDocument;
```
ramya-rao-a: Is passing the entire document necessary? Since this is only used for getting the filename, why not store the filename directly in the `ExtendedCompletionItem`

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+	public resolveCompletionItem(item: vscode.CompletionItem, token: vscode.CancellationToken): vscode.ProviderResult<vscode.CompletionItem> {
+		if (item instanceof ExtendedCompletionItem) {
+			const goRuntimePath = getBinPath('go');
+			if (!goRuntimePath) {
+				vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
+				return;
+			}
+
+			const env = getToolsEnvVars();
+			const cwd = path.dirname(item.document.fileName);
```
ramya-rao-a: Does `cwd` matter to `go doc`?

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
 
+	public resolveCompletionItem(item: vscode.CompletionItem, token: vscode.CancellationToken): vscode.ProviderResult<vscode.CompletionItem> {
+		if (item instanceof ExtendedCompletionItem) {
+			const goRuntimePath = getBinPath('go');
+			if (!goRuntimePath) {
+				vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
+				return;
+			}
+
+			const env = getToolsEnvVars();
```
ramya-rao-a: `env` is created but not used

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+			const cwd = path.dirname(item.document.fileName);
+
+			return new Promise<vscode.CompletionItem>((resolve, reject) => {
+				let args = ['doc', '-c', '-cmd', '-u'];
+				if (item.package) {
```
ramya-rao-a: In what scenario do we expect package to be absent?

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+
+				cp.execFile(goRuntimePath, args, { cwd }, (err, stdout, stderr) => {
+					if (err) {
+						reject(err);
+					}
+
+					let doc = '';
+					const goDocLines = stdout.toString().split('\n');
+					// i = 1 to skip the func signature line
+					for (let i = 1; goDocLines[i].startsWith('    '); i++) {
```
ramya-rao-a: I would add the `i < goDocLines.length` check here as well.

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
+				let args = ['doc', '-c', '-cmd', '-u'];
+				if (item.package) {
+					args.push(item.package, item.label);
+				} else {
+					args.push(item.label);
+				}
+
+				cp.execFile(goRuntimePath, args, { cwd }, (err, stdout, stderr) => {
+					if (err) {
+						reject(err);
```
ramya-rao-a: In case of error, we do not want the completion feature to fail. So lets `return resolve(item)` here.

src/goSuggest.ts
```diff
@@ -75,6 +81,44 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 		});
 	}
 
+	public resolveCompletionItem(item: vscode.CompletionItem, token: vscode.CancellationToken): vscode.ProviderResult<vscode.CompletionItem> {
+		if (item instanceof ExtendedCompletionItem) {
+			const goRuntimePath = getBinPath('go');
+			if (!goRuntimePath) {
+				vscode.window.showInformationMessage('Cannot find "go" binary. Update PATH or GOROOT appropriately');
+				return;
```
ramya-rao-a: I am not sure if the completion feature fails in some way when we return this way. Can you test this scenario? One way to be sure is to always return the `item` as is when we are not able to get the docs.

src/goSuggest.ts
segevfiner: How VS Code behaves in regards to this callback failing seems poorly documented. I will have to check this.

src/goSuggest.ts
segevfiner: If you return with null/undefined, it simply uses the original item. It's documented here: [CompletionItemProvider](https://code.visualstudio.com/docs/extensionAPI/vscode-api#CompletionItemProvider.resolveCompletionItem)

src/goSuggest.ts
segevfiner: I think that's the code handling this: [vscode/src/vs/editor/contrib/suggest/suggest widget.ts:609-631](https://github.com/Microsoft/vscode/blob/f13c8d1eab216fb84472a68c3ba8c3d9d76b1a30/src/vs/editor/contrib/suggest/suggestWidget.ts#L609-631).

It looks like it simply carries on without the extra information and doesn't abort the completion. Probably logging the error into one of the output streams. That's how I remember it behaving too, but please verify me. If you feel more comfortable with just `resolve(item)` than I can change it as such. But in that case we should log the error ourself somewhere. Otherwise it would be impossible to debug errors here.

ramya-rao-a: https://github.com/Microsoft/vscode/blob/f13c8d1eab216fb84472a68c3ba8c3d9d76b1a30/src/vs/editor/contrib/suggest/suggest.ts#L157 is where the `resolveCompletionItem` gets called. When it doesnt return anything, `value` is `undefined`.

The `assign` function then tries to run `Object.keys()` on it which would give a `Cannot convert undefined or null to object` error.

Error messages can be logged using console.log such that if anyone wants to investigate why docs are not present, the logs can help.

src/goSuggest.ts
ramya-rao-a: Yes, lets just pass the minimum information.

ramya-rao-a:
> In older versions of `gocode`, `package` will probably end up as `undefined`. With the current code, it will end up looking the symbol in the current package. This might not be desirable... So maybe it should be prevented with a prompt to update `gocode`?

And in the newer `gocode`, is `package` an empty string or `undefined` for symbols from current package? If we can differentiate between the 2 cases, then yes, we should skip the doc finding for the older gocode and prompt to update `gocode`

segevfiner:
> And in the newer gocode, is package an empty string or undefined for symbols from current package? If we can differentiate between the 2 cases, then yes, we should skip the doc finding for the older gocode and prompt to update gocode

It's an empty string, the property always seems to exist in the newer version ([gocode/formatters.go:160-161](https://github.com/nsf/gocode/blob/7b1d4e18cdc58a74dc1bd4c2d45b3f1b2ca227c3/formatters.go#L160-L161)). I pushed a commit that should implement this.

ramya-rao-a:
Thanks for all your work!

I have pushed https://github.com/Microsoft/vscode-go/commit/3ba34e0ae5c02079af8da5892389c12aa7647c2f that converts the doc string to markdown. It looks better now.

segevfiner:
> Thanks for all your work!
> 
> I have pushed [3ba34e0](https://github.com/Microsoft/vscode-go/commit/3ba34e0ae5c02079af8da5892389c12aa7647c2f) that converts the doc string to markdown. It looks better now.

But aren't Go documentation strings not Markdown? They don't render as Markdown in `godoc`, and friends. Wouldn't this cause random formatting issues when stuff that wasn't formatted as Markdown to begin with gets treated as such? 

godoc comments do have some very minimal formatting: paragraph handling (Similar to Markdown) indented preformatted text, hyperlinks, and sections (If I'm not missing any...) But no things like bold/italic and such. 

So I think the right way to handle this is to write a function which will convert any godoc formatting not already handled by `godoc` to Markdown but escaping any other construct so that it will appear plain as it does in `godoc`.

ramya-rao-a:
> godoc comments do have some very minimal formatting: paragraph handling (Similar to Markdown) indented preformatted text, hyperlinks, and sections (If I'm not missing any...) But no things like bold/italic and such.

Can you point me to some examples?

segevfiner:
> Can you point me to some examples?

Some references:
https://golang.org/pkg/go/doc/#ToHTML
https://blog.golang.org/godoc-documenting-go-code

This shows the difference in rendering:

```go
package docexample

// Hello **World**!
func Frob() {
	
}

```

VS Code:
image

godoc:
image

Granted, in this example it doesn't matter much, but there are likely plenty of examples where this can really mess up the formatting, I just don't have one on the top of my head.

I remember vscode-python hitting a similar issue where Python comments are often formatted as reStructuredText, which isn't the same as Markdown. Treating it like it's Markdown lead to formatting errors in some libraries.

ramya-rao-a:
I am aware of the markdown syntax elements like ** or ~~ causing issues. We already have this problem in the hover widget. 

But since that is a corner case, I am ok with both the hover and completion widgets having this problem.

When we do find a solution, we can apply it to both widgets.

ramya-rao-a:
@segevfiner You mentioned the below when I asked about the need for `cwd`

> Yeah, I got bitten by that when testing. It uses it to figure out which package we are in when `gocode` returns `package===""`

I tested the case of symbol from same package, I don't get any docs...

screen shot 2018-11-04 at 7 31 55 pm

segevfiner:
> I tested the case of symbol from same package, I don't get any docs...

I do get them locally, but you have to save the file first as `go doc` has no support for unsaved files.

ramya-rao-a:
In the below case, I am triggering suggestions on a saved file

screen shot 2018-11-05 at 8 38 48 am

segevfiner:
> 
> 
> In the below case, I am triggering suggestions on a saved file
> 
> screen shot 2018-11-05 at 8 38 48 am

You found a bug. The problem is that the package import path likely ends with `hello`, and the function is called `hello`, so the invocation `go doc -c -cmd -u hello` in the `hello` package directory prefers to print the documentation for the *package* `hello` rather than the documentation for the *function* `hello`. See [cmd/go - Show documentation for package or symbol](https://golang.org/cmd/go/#hdr-Show_documentation_for_package_or_symbol).

I guess we should handle the current package (`package = ""`) case differently. If we can figure out the package for the current file, we can pass that instead. (Instead of setting `cwd`). I think we can get that by `go list` on the file's `dirname`. But I'm not sure how efficient that is.

PR #2059, opened 2018-10-28
Add debugTestAtCursor command

### LINKED ISSUE #1088: Put run/debug test links in the context menu instead
The blank non-line line the run/debug test links are in looks weird and makes it hard to tell at a glance what I'm looking at.

screen shot 2017-07-13 at 10 43 06 pm

Can they be moved to context menu items instead?

screen shot 2017-07-13 at 10 43 19 pm

Ashiroq:
Closes #1088

Adds command for debugging test at the current position of cursor similar to "debug test" codeLens.
By default, the command is hidden in the context menu, if codeLens is enabled just like with testAtCursor command.

Let me know if I can improve this :)


vladbarosan:
@Ashiroq I have some fixes at https://github.com/Microsoft/vscode-go/pull/2281 , can you give me permission to push to your branch ?

Ashiroq:
I sent you an invitation to collaborators.

vladbarosan:
@Ashiroq I managed to update the branch. @ramya-rao-a i tested the the suite scenario for both the codelens and the command and they work fine.

src/testUtils.ts
```diff
@@ -126,6 +127,23 @@ export function extractInstanceTestName(symbolName: string): string {
...
+/**
+ * Gets the appropriate debug arguments for a debug session on a test function.
+ * @param document  The document containing the tests
+ * @param testFunctionName The test function to get the debug args
+ * @param testFunctions The test functions found in the document
+ */
+export function getTestFunctionDebugArgs(document: vscode.TextDocument, testFunctionName: string, testFunctions: vscode.SymbolInformation[]): string[] {
+	const instanceMethod = extractInstanceTestName(testFunctionName);
+	if (instanceMethod) {
+		const testFns = findAllTestSuiteRuns(document, testFunctions);
```
ramya-rao-a: Would there ever be a case where `testFns` would be an empty array? i.e we weren't able to find the function corresponding to the suite.run()?

vladbarosan: yes it can ( e.g. if there is no suite method. but that will work as expected ( i.e. the regex wont match any suite since there is none). and the output will be a warning that no tests to run. 
image

vladbarosan: only thing that we might want to consider is on the CodeLens behavior in the following case:
- User writes a test as part of a test suite but doesn't add the test which runs the suite.  Which would be more helpful, not to show anything in codelens or to show them and when they run it to have the warning pasted above ( current behavior ) ?

vladbarosan: or we could try to return a custom error that no test suite exists ?

ramya-rao-a: > yes it can ( e.g. if there is no suite method. but that will work as expected ( i.e. the regex wont match any suite since there is none). and the output will be a warning that no tests to run.

In this case, we would be running `go test -test.run -testify.m CurrentTestFunctionName`
We know before hand this will not work. 

What would happen if we run `go test -test.run CurrentTestFunctionName` in such case?

vladbarosan: where `CurrentTestFunctionName` is a test suite test  same warning. Also FYI currently the same behavior is for running the tests as it was behavior.

ramya-rao-a: >  On that either way we will need to decide if want the warning when creating the codelens or when someone actually clicks it.

Warnings can be done either via a notification (prompts) or just printing it to the Dev Console.

The former is too intrusive for a task like creating codelens. Codelens are silent helpers and should not interfere with daily tasks of a user. 

The latter (Dev Console) is not too useful for the current case. A user will not be driven to open that up and read through unless there they feel that something is wrong.  In our case, the debug console would have given necessary information after they click on the codelens.

We can keep this as is.

Just one last piece of feedback.

The function `debugTestAtCursor` shares a lot of code with the function `testAtCursor`. Instead of having a new file for the former, let's move it to the same file as the latter. The common code between the 2 can be moved to `testUtils.ts` file

vladbarosan: @ramya-rao-a updated

src/goTest.ts
```diff
@@ -34,66 +45,80 @@ export function testAtCursor(goConfig: vscode.WorkspaceConfiguration, isBenchmar
...
-					if (success && tmpCoverPath) {
-						return applyCodeCoverageToAllEditors(tmpCoverPath, testConfig.dir);
-					}
-				});
-			});
-		});
-	}).then(null, err => {
-		console.error(err);
-	});
+	const isMod = await isModSupported(editor.document.uri);
```
vladbarosan: here I have a question. previously the testConfig was saved without the `isMod` property. was that intentional or by accident ?

vladbarosan: i presume it was intended to be saved since the call to running previous test used it. if so then its fine since i just made it immutable but still is saved

vladbarosan:
@jhendrixMSFT , @ramya-rao-a the metalinter test seems to be failing although from looking at it doesn't seem related to anything i've changed. Any idea why ?

jhendrixMSFT:
@vladbarosan There was a new release of the metalinter [v3.0.0](https://github.com/alecthomas/gometalinter/releases/tag/v3.0.0) which has breaking changes.  How about you update travis to pin to v2.0.12 until we have time to react to these changes?

vladbarosan:
@jhendrixMSFT ok figured it out, it seems not only that the new version broke us but also the new install script version ( probably because it seems to be removing some checkers ). Locked to tne previous script version for now and tracking the issue here https://github.com/Microsoft/vscode-go/issues/2292

PR #2060, opened 2018-10-28
Remove go 1.8 version for travis pipeline since it is not longer support

angulito:
See 
[issue](https://github.com/golang/go/issues/28291): here someone opens an issue to the go project explaining the current pipeline error for go#1.8:
image

 
[nel-collector project](https://github.com/google/nel-collector/pull/35): here you can see here they have removed the tests for go 1.8 since it's not longer support and now the go tools in the go project contains new syntax which go#1.8 does not support.

I've ran the pipeline [here](https://travis-ci.org/angulito/vscode-go/builds/447535413) and now it works


ramya-rao-a:
Thanks for the PR @angulito 

#2032 also is a PR to do the same. Like I have mentioned [there](https://github.com/Microsoft/vscode-go/pull/2032#issuecomment-431615055), I dont want to stop support for Go 1.8

Feel free to follow the steps in https://github.com/Microsoft/vscode-go/pull/2032#issuecomment-431619982 instead to fallback to the older gocode for Go versions less than 1.10

angulito:
Yep, I understand you @ramya-rao-a, although maybe it will be difficult to develop with open source libraries that maybe does not have support for Go 1.8.

- I've created a new PR fixing a minor typo in the pipeline: https://github.com/Microsoft/vscode-go/pull/2065
- I've checked there is a new library which seems not supporting the Go 1.8 and I've seen you have opened an issue - so I guess we cannot fix that pipeline error right now.

Thanks to you.

angulito:
@ramya-rao-a also you can check here golint project has removed 1.8 version because it's not longer support too - https://travis-ci.org/golang/lint/jobs/444888944

ramya-rao-a:
I've updated the travis file to skip golint and gometalinter for 1.8
Also updated the tests to not test these 2 tools for 1.8

PR #2070, opened 2018-11-01
#2063 only suggest exported member name at start of comment

### LINKED ISSUE #2063: IntelliSense activates on single-line comments for exported symbols.
Steps to Reproduce:

Add to a comment on an exported name, using the period charater `.` will cause IntelliSense to activate.

Linked to https://github.com/Microsoft/vscode-go/issues/2006

kagof:
Resolves #2063 - only suggest the exported member name if the current comment has no content.

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -51,7 +51,7 @@ class ExtendedCompletionItem extends vscode.CompletionItem {
 	fileName: string;
 }
 
-const lineCommentRegex = /^\s*\/\/\s+/;
+const lineCommentWithNoContentRegex = /^\s*\/\/\s+$/;
 const exportedMemberRegex = /(const|func|type|var)(\s+\(.*\))?\s+([A-Z]\w*)/;
 const gocodeNoSupportForgbMsgKey = 'dontshowNoSupportForgb';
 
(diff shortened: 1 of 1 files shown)


rubenzinho07:
😈

Em Dom, 4 de nov de 2018 04:21, Karl Goffin <notifications@github.com>
escreveu:

src/goSuggest.ts
```diff
@@ -133,15 +133,15 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
-				if (lineCommentRegex.test(lineTillCurrentPosition) && position.line + 1 < document.lineCount) {
+				if (lineCommentFirstWordRegex.test(lineTillCurrentPosition) && position.line + 1 < document.lineCount) {
 					let nextLine = document.lineAt(position.line + 1).text.trim();
 					let memberType = nextLine.match(exportedMemberRegex);
 					let suggestionItem: vscode.CompletionItem;
 					if (memberType && memberType.length === 4) {
-						suggestionItem = new vscode.CompletionItem(memberType[3], vscodeKindFromGoCodeClass(memberType[1], ''));
+							suggestionItem = new vscode.CompletionItem(memberType[3], vscodeKindFromGoCodeClass(memberType[1], ''));
+						}
+						return resolve(suggestionItem ? [suggestionItem] : []);
```
ramya-rao-a: Why was this return statement moved? With this change, when typing any text in the comments, we would end up calling gocode instead of exiting early

ramya-rao-a: Also can you update the existing test case to reflect the bug fix? See https://github.com/Microsoft/vscode-go/blob/0.7.0/test/go.test.ts#L1014

Issue #1989, opened 2018-10-11
The debugger doesn't show pointer values (like 0xf800000040)


willfaught:
So it seems like you can't tell if two pointers point to the same thing.

<img width="621" alt="screen shot 2018-10-10 at 6 25 33 pm" src="https://user-images.githubusercontent.com/63830/46774970-f36fb300-ccb9-11e8-944c-64c26a611fdd.png">

Latest version of macOS High Sierra (10.13.6 (17G65)), VSCode (1.28.0), and vscode-go (0.6.91).

eternal-flame-AD:
I usually watch `uintptr(x)` to obtain the pointer value. Probably a hack but I think that helps in your case:)

ramya-rao-a:
@willfaught Can you provide a small sample code and elaborate on what you would expect vs what actually is being shown?

willfaught:
@ramya-rao-a Sure:

```go
func TestX(t *testing.T) {
	var i, j int
	var p = &i
	var q = &j
	t.Logf("%#v %#v", p, q)
}
```

When run, you see output like:

```
=== RUN   TestX
--- PASS: TestX (0.00s)
    foofile_test.go:9: (*int)(0xc000018150) (*int)(0xc000018158)
PASS
ok  	foopackage	0.010s
```

When debugging TestX, you see these local variables:

<img width="212" alt="screen shot 2018-10-29 at 7 50 20 pm" src="https://user-images.githubusercontent.com/63830/47692770-2cfa5680-dbb4-11e8-98a5-cea6c884a9e1.png">

The pointer addresses for p and q aren't shown there, so you can't tell if they're equal and thus point to the same thing:

```
p: <*int>
q: <*int>
```

Instead, you should see something like:

```
p: 0xc000018150 <*int>
q: 0xc000018158 <*int>
```

There should also be a way to drill down into them to view the values they point to, like you can for the *testing.T in that same screen shot.

ramya-rao-a:
This feature is now available in the latest version of the Go extension (0.8.0)

PR #2075, opened 2018-11-02
Improve debugging of pointer values

brycekahle:
See #1989 

This displays the address of the variable the pointer is pointing to (for pointer comparison). It also allows expansion in the debugger.

Example code:
```
package main

import "testing"

func TestX(t *testing.T) {
	var i, j int
	i = 0
	j = 1
	k := []string{"asdf"}
	var p = &i
	var q = &j
	var r = &k
	t.Logf("%#v %#v %#v", p, q, r)
}
```

Shown in debugger:
screen shot 2018-11-02 at 2 35 52 pm

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -935,8 +935,8 @@ class GoDebugSession extends DebugSession {
 					});
 				}
 				return {
-					variablesReference: v.children[0].children.length > 0 ? this._variableHandles.create(v.children[0]) : 0
 					result: `<${v.type}>(0x${v.children[0].addr.toString(16)})`,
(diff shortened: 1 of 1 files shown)


brycekahle:
This does add an intermediate level to the expansion. Alternatively, I could attempt to rework the code to display it like so:
```
p: 0 <*int>(0xc000078080)
q: 1 <*int>(0xc000078088)
r: <*[]string>(0xc000086080)
  [0]: "asdf"
```

ramya-rao-a:
Re-working this to look like 
https://github.com/Microsoft/vscode-go/pull/2075#issuecomment-435472584 would be simple for simple types like int, but would get complicated for arrays and maps I believe.

The current solution should be good enough.

Thanks!

PR #2080, opened 2018-11-03
Add stacktrace paging support to debug adapter

### LINKED ISSUE #946: Call Stack limitied to 20
If you run this code

```
package main

func main() {
	d := 30
	recursiveCall(d)
}

func recursiveCall(depth int) {
	if depth <= 0 {
		return
	}
	recursiveCall(depth - 1)
}
```

with a breakpoint on line 10, you can't access the value of d in the main function because it's too far down the stack. How can this be extended?

brycekahle:
Fixes #946

This still has a maximum depth of 50. Delve has no way to specify an infinite max depth, or query the length of the stacktrace.

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,7 +500,7 @@
               "stackTraceDepth": {
                 "type": "number",
                 "description": "Maximum depth of stack trace collected from Delve",
-                "default": 50
+                "default": 20
               }
             }
           }
--- a/package.json
+++ b/package.json
@@ -496,6 +496,11 @@
                 ],
                 "description": "Delve Api Version to use. Default value is 2.",
                 "default": 2
+              },
+              "stackTraceDepth": {
+                "type": "number",
+                "description": "Maximum depth of stack trace collected from Delve",
+                "default": 50
               }
             }
           }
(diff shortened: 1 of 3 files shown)


brycekahle:
AFAICT the CI failures are not related to my changes. Perhaps someone can re-run them?

ramya-rao-a:
@brycekahle Like you said, this just increases the depth from 20 to 50

How about adding a setting to the debug configuration so that the user can set the depth instead?

brycekahle:
@ramya-rao-a I've added a setting for the maximum stack trace depth.

brycekahle:
@ramya-rao-a FYI changing this to 20 by default, will not allow any paging unless the setting is changed. VS Code uses a page size of 20, and by default we never indicate there are more than 20 frames (via `totalFrames`).

Perhaps it is worth filing an issue on delve to have better support for stack frame paging?

ramya-rao-a:
Good point, I'll change the default to 30.

For delve, looks like something similar has already been discussed. See https://github.com/derekparker/delve/issues/989 due to which the default depth for delve is now 50

Issue #2078, opened 2018-11-03
`The "godoc" command is not available` upon hover


kroppt:
Version of VS Code: `1.28.2`
Version of Go plugin: `0.6.93`
Version of Go: `1.11.1`

OS/Platform you are working on:
`Linux 4.18.3-gentoo`

- sample code which can be used to reproduce the issue:
```golang
package main

func main() {
}
```

- any Go related settings you might have added/changed:
None or valid `go.gopath`

Steps to Reproduce:

1. Update tools
2. Open a Go file with extension enabled
3. Hover over function
4. Minimal information shows up in tooltip, and a notification says, `The "godoc" command is not available.` No Go extension output. Warnings show up in dev tools upon hover:

`[Violation] Added non-passive event listener to a scroll-blocking 'mousewheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`

`[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`

Workaround:
- Change `go.docsTool` to `gogetdoc` after go-get'ing the binary.

ramya-rao-a:
We use the `godoc` binary that is shipped with the Go distribution. You must be missing that. Check your $GOROOT/bin

See https://bugs.archlinux.org/task/60537

We should fallback to `go doc` in such cases though..

kroppt:
The [portage ebuild](https://packages.gentoo.org/packages/dev-lang/go) just runs `make.bash` from source and copies the folders over (including bin). So, I think this is an issue with building from source.

ramya-rao-a:
In the latest update to this extension (0.7.0), I have replaced the use of `godoc` with `go doc`. So this should be fixed now.

PR #2096, opened 2018-11-05
Use go doc instead of godoc

ramya-rao-a:
This PR covers the below:
- Replace the use of `godoc` with `go doc` for the hover widget. This will solve issues like #2078 since godoc is no longer go gettable
- Make both the hover widget and the completion widget share the code around the use of `go doc`
- Use `go list` to get package import path given the absolute path

@segevfiner, Can you take a look and let me know your thoughts, since this is close to your work in #2054

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -8,7 +8,7 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
-import { byteOffsetAt, getBinPath } from './util';
+import { byteOffsetAt, getBinPath, runGodoc } from './util';
 import { promptForMissingTool } from './goInstallTools';
 import { getGoVersion, SemVersion, goKeywords, isPositionInString, getToolsEnvVars, getFileArchive, killProcess } from './util';
 import { isModSupported, promptToUpdateToolForModules } from './goModules';
@@ -28,6 +28,7 @@ export interface GoDefinitionInformation {
 interface GoDefinitionInput {
 	document: vscode.TextDocument;
 	position: vscode.Position;
+	word: string;
 	includeDocs: boolean;
 	isMod: boolean;
 }
@@ -51,6 +52,7 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 			const input: GoDefinitionInput = {
 				document,
 				position,
+				word,
 				includeDocs,
 				isMod
 			};
@@ -96,8 +98,6 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
 					return resolve(null);
 				}
 				let [_, file, line, col] = match;
-				let signature = lines[1];
-				let godoc = getBinPath('godoc');
 				let pkgPath = path.dirname(file);
 				let definitionInformation: GoDefinitionInformation = {
 					file: file,
@@ -111,27 +111,13 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
 				if (!input.includeDocs) {
 					return resolve(definitionInformation);
 				}
-				p = cp.execFile(godoc, [pkgPath], { env }, (err, stdout, stderr) => {
-					if (err && (<any>err).code === 'ENOENT') {
-						vscode.window.showInformationMessage('The "godoc" command is not available.');
-					}
-					let godocLines = stdout.toString().split('\n');
-					let doc = '';
-					let sigName = signature.substring(0, signature.indexOf(' '));
-					let sigParams = signature.substring(signature.indexOf(' func') + 5);
-					let searchSignature = 'func ' + sigName + sigParams;
-					for (let i = 0; i < godocLines.length; i++) {
-						if (godocLines[i] === searchSignature) {
-							while (godocLines[++i].startsWith('    ')) {
-								doc += godocLines[i].substring(4) + '\n';
-							}
-							break;
-						}
-					}
-					if (doc !== '') {
+				runGodoc(pkgPath, input.word, token).then(doc => {
+					if (doc) {
 						definitionInformation.doc = doc;
 					}
-					return resolve(definitionInformation);
+					resolve(definitionInformation);
+				}, () => {
+					resolve(definitionInformation);
 				});
 			} catch (e) {
 				reject(e);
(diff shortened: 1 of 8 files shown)


oneslash (approved):

oneslash:
I love this change

oneslash:
I would also make `go fmt` default formatter

src/goDeclaration.ts
```diff
@@ -111,27 +111,13 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
...
-						}
-					}
-					if (doc !== '') {
+				runGodoc(pkgPath, input.word, token).then(doc => {
+					if (doc) {
 						definitionInformation.doc = doc;
 					}
-					return resolve(definitionInformation);
+					resolve(definitionInformation);
+				}, () => {
```
segevfiner: Shouldn't this be `.catch()` so that it also handles errors in the `then` callback? Are you sure you want to squelch any errors here and not log them anywhere?

src/goSuggest.ts
```diff
@@ -92,36 +91,13 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
...
-					doc += goDocLines[i].substring(4) + '\n';
-				}
-
-				item.documentation = new vscode.MarkdownString(doc);
-				resolve(item);
-			});
+		return runGodoc(item.package || path.dirname(item.fileName), item.label, token).then(doc => {
+			item.documentation = new vscode.MarkdownString(doc);
+			return item;
+		}, err => {
```
segevfiner: The same about `.catch()` as in `goDeclaration.ts`.

src/util.ts
```diff
@@ -896,3 +897,48 @@ export function cleanupTempDir() {
...
+	if (!symbol) {
+		return Promise.reject(new Error('Symbol not provided'));
+	}
+
+	const goRuntimePath = getBinPath('go');
+	if (!goRuntimePath) {
+		return Promise.reject(new Error('Cannot find "go" binary. Update PATH or GOROOT appropriately'));
+	}
+
+	const cwd = packagePath;
```
segevfiner: Unnecessary variable. Probably a left over from the previous code.

src/util.ts
```diff
@@ -896,3 +897,48 @@ export function cleanupTempDir() {
...
+	const getCurrentPackagePromise = 1 !== 1 ? getCurrentPackage(cwd) : Promise.resolve(cwd);
```
segevfiner: `1 !== 1`...? So `getCurrentPackage` is never really called...

I think we only need `getCurrentPackage` in the `package === ""` case from `gocode`. We can simply call it there so that this function doesn't have to call `go list` unnecessarily. Leaving this function to only handle correct package paths.

ramya-rao-a: lol... I was testing the file path case internally, and checked it in.

The `1 !== 1` was meant to be `path.isAbsolute(cwd)`

Yes, and this was for the `godef` case which returns file path always

ramya-rao-a:
Also added tests for the `resolveCompletionItem`

ramya-rao-a:
@segevfiner FYI: Turns out the `package` field from `gocode` is empty when the suggested item is an imported package.

This resulted in the `go doc` failing for such suggestions. Fixed this with https://github.com/Microsoft/vscode-go/blob/5a6b9a99a2562354c693dea99b692ded43111e2a/src/goSuggest.ts#L85

These items were also missing the `detail` part. So fixed that with https://github.com/Microsoft/vscode-go/blob/5a6b9a99a2562354c693dea99b692ded43111e2a/src/goSuggest.ts#L261-L266

src/util.ts
```diff
@@ -917,28 +917,41 @@ export function runGodoc(packagePath: string, symbol: string, token: vscode.Canc
...
+				// Skip trailing empty lines
+				let lastLine = godocLines.length - 1;
+				for (; lastLine > 1; lastLine--) {
+					if (godocLines[lastLine].trim()) {
+						break;
+					}
+				}
+
+				let doc = '';
+				for (let i = 1; i <= lastLine; i++) {
```
segevfiner: With `<=`, aren't you also entering this loop with the first trailing empty line? Is this intended?

segevfiner:
> @segevfiner I found another case which wasnt handled before which is multi line docs with an empty line in between. I've added 

Grrr. I wish they could have just printed it in a machine parse-able format... after all it's probably already available in the parser AST like that...

> @segevfiner FYI: Turns out the `package` field from `gocode` is empty when the suggested item is an imported package.
> 
> This resulted in the `go doc` failing for such suggestions. Fixed this with
> 
> [vscode-go/src/goSuggest.ts](https://github.com/Microsoft/vscode-go/blob/5a6b9a99a2562354c693dea99b692ded43111e2a/src/goSuggest.ts#L85)
> 
> Line 85 in [5a6b9a9](/Microsoft/vscode-go/commit/5a6b9a99a2562354c693dea99b692ded43111e2a)
>  if (!(item instanceof ExtendedCompletionItem) || item.kind === vscode.CompletionItemKind.Module) { 

`go doc` will show the package documentation if we call it with only the package (Maybe an empty second argument also works). It would be nice to show that. But it seems to use a different output format for package documentation which also includes an exported item list at the end which will be annoying to strip (Nasty regex time?)

> These items were also missing the `detail` part. So fixed that with
> 
> [vscode-go/src/goSuggest.ts](https://github.com/Microsoft/vscode-go/blob/5a6b9a99a2562354c693dea99b692ded43111e2a/src/goSuggest.ts#L261-L266)
> 
> Lines 261 to 266 in [5a6b9a9](/Microsoft/vscode-go/commit/5a6b9a99a2562354c693dea99b692ded43111e2a)
>  if (suggest.class === 'package') { 
>  	const possiblePackageImportPaths = this.getPackageImportPath(item.label); 
>  	if (possiblePackageImportPaths.length === 1) { 
>  		item.detail = possiblePackageImportPaths[0]; 
>  	} 
>  }

So you will offer the package import path if you can figure out which package is it. e.g. No package with the same name. I guess `gocode` can be improved later to output the full package import path for package completions so as to make this more accurate.

src/util.ts
ramya-rao-a: `lastLine` points to last non empty line not the first empty line. So we should be good.

segevfiner: Sorry I misread the code. 😓 
You are right.

PR #2102, opened 2018-11-05
Fix completion documentation for current package symbols

segevfiner:
`go doc` prefers packages when given only a name without a package. So use `go list` to figure out the current package name instead of relying on the working directory.

See https://github.com/Microsoft/vscode-go/pull/2054#issuecomment-435988584.

Not sure if that's the best place for `getPackageFromDir`, or whether that's the best way to do it performance wise.

@ramya-rao-a

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -81,6 +81,25 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
@segevfiner Closing this PR in favor of #2096

I already had some code that runs `go list` to get package import path given an absolute path and then caches it for the given VS Code session.

Can you take a look at #2096?

PR #2103, opened 2018-11-05
add a gocodeFlags field to goConfig

stamblerre:
As more flags are being added to gocode, it is useful to have a field that allows a user to manually customize the flags being passed to gocode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -800,6 +800,19 @@
           "description": "Environment variables that will passed to the processes that run the Go tools (e.g. CGO_CFLAGS)",
           "scope": "resource"
         },
+        "go.gocodeFlags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [
+            "-builtin",
+            "-ignore-case",
+            "-unimported-packages"
+          ],
+          "description": "Additional flags to pass to gocode.",
+          "scope": "resource"
+        },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -800,6 +800,19 @@
...
         },
+        "go.gocodeFlags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [
+            "-builtin",
+            "-ignore-case",
+            "-unimported-packages"
```
ramya-rao-a: `unimported-packages` only works for standard packages, correct?

stamblerre: Yep, but it's an open to add support for other packages. Hopefully I'll have a chance to add that soon.

ramya-rao-a: When 2 packages have the same name (think of text/template, html/template), does `template.` give results from both packages?

stamblerre: One package is prioritized and the results for it are returned - so in this case `html/template` is used instead of `text/template` (https://github.com/mdempsky/gocode/blob/master/internal/suggest/stdlib.go).

marwan-at-work:
@stamblerre @ramya-rao-a I think this PR breaks auto-complete for module-enabled auto-completion. Because `-unimported-packages` flag is not recognized by stamblerre/gocode

PR #2114, opened 2018-11-10
Differentiate between interfaces and structs in outline

karthikraobr:
This PR fixes the code outline from erroneously displaying structs as interfaces.

Before 
screenshot 2018-11-11 at 00 29 59

After
screenshot 2018-11-11 at 00 31 02

### DIFF
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -158,7 +158,7 @@ function convertToCodeSymbols(
 			range = new vscode.Range(document.positionAt(start), document.positionAt(end));
 			if (decl.type === 'type') {
 				let line = document.lineAt(document.positionAt(start));
-				let regex = new RegExp('type .* struct');
+				let regex = new RegExp(`^\\s*type\\s*${decl.label}\\s*struct\\b`);
 				decl.type = regex.test(line.text) ? 'struct' : 'type';
 			}
 		}
--- a/src/goOutline.ts
+++ b/src/goOutline.ts
@@ -159,7 +159,7 @@ function convertToCodeSymbols(
 			if (decl.type === 'type') {
 				let line = document.lineAt(document.positionAt(start));
 				let regex = new RegExp('\\bstruct\\b');
-				decl.type = line.text.match(regex) ? 'struct' : 'type';
+				decl.type = regex.test(line.text) ? 'struct' : 'type';
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


karthikraobr:
@ramya-rao-a No idea why the test failed :(

src/goOutline.ts
```diff
@@ -156,10 +156,13 @@ function convertToCodeSymbols(
 			let start = byteOffsetToDocumentOffset(decl.start - 1);
 			let end = byteOffsetToDocumentOffset(decl.end - 1);
 			range = new vscode.Range(document.positionAt(start), document.positionAt(end));
+			if (decl.type === 'type') {
+				let line = document.lineAt(document.positionAt(start));
+				let regex = new RegExp('\\bstruct\\b');
```
ramya-rao-a: Can you make this regex stricter? Something like `RegExp('^\s*type\s+' + label + '\s+struct\s+ {')` ?

Also, update the test for the outline feature.

karthikraobr: Thanks. I guess you are referring to an older version of my code. But, as you pointed out even the newer exp is not very strict. Will fix it and fix the tests.

karthikraobr: For some reason, regex.test is always returning false if I use the above regex (IMO it is right). I even tried  ```let regex = new RegExp(`^\s*type\s${label}\s+struct\s*{`);```.  Still returns false.

ramya-rao-a:
@karthikraobr I think I am missing something here...

Below is the screenshot from using the latest Go extension

image

The interface and struct are appearing separately for me without the changes in this PR.

What am I missing?

karthikraobr:
@ramya-rao-a I am on the latest version of vscode and the plugin as well. But doesn't work for me. Moreover, a struct is being displayed as a class which is wrong. Finally, there is no [code](https://github.com/Microsoft/vscode-go/blob/e05d03579d33b882636e11e27bb4094b3e294cee/src/goOutline.ts#L124) to detect classes right? Are you using a beta?

karthikraobr:
@ramya-rao-a Thanks a ton for fixing the regex. I have added tests for struct outline. Let me know if I need to add more tests.

karthikraobr:
This PR also fixes breadcrumbs for go files.

ramya-rao-a:
Thanks @karthikraobr!

Issue #2085, opened 2018-11-04
Debug: Start without debugging cmd should always use current file


ramya-rao-a:
`Debug: Start without Debugging` is the command to use to simply run the current file rather than debugging. 

This works just fine just fine in the below cases:
- No launch.json file or it is empty
- The currently selected debug configuration in launch.json has the `program` attribute pointing to a file.

This falls back to debugging when the `program` attribute in the currently selected debug configuration points to a directory. This is because `go run` needs a file.

In such cases, we should use the file open in the currently active editor instead of falling back to debugging the package.

PRs are most welcome to fix this.
Code Pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- Update `resolveDebugConfiguration` in `goDebugConfiguration.ts` file to return a new property called `currentFile`. This should have the path to file open in the active editor.
- [Debug the debugger](
https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter)
- Update the `go run` call at https://github.com/Microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts#L317 to use the value from the `currentFile` property if the `program` property does not point to a file itself

ramya-rao-a:
This feature is now available in the latest version of the Go extension (0.8.0)

PR #2123, opened 2018-11-13
Fix 'Debug: Start Without Debugging' to use current file and include buildFlags/args

### LINKED ISSUE #2086: Debug: Start without Debugging command doesn't use the args, buildflags from the debug configuration
`Debug: Start without Debugging` is the command to use to simply run the current file rather than debugging. 

This is expected to use the currently selected debug configuration in launch.json
It does use the `program` and `env` attributes as expected, but ignores the `args` and `buildflags` attributes.

PRs are most welcome to fix this.
Code Pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- [Debug the debugger](
https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter)
- Update the `go run` call at https://github.com/Microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts#L317 to use the args and buildflags

brycekahle:
Include buildFlags and args when running without debugging. Use current file (if there is one) when 'running without debugging'.

Fixes #2086 and #2085

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -211,6 +211,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 	apiVersion: number;
 	/** Delve maximum stack trace depth */
 	stackTraceDepth: number;
+	currentFile: string;
 }
 
 process.on('uncaughtException', (err: any) => {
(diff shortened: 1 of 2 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -209,6 +209,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 	dlvToolPath: string;
 	/** Delve Version */
 	apiVersion: number;
+	currentFile: string;
```
brycekahle: @ramya-rao-a This was the only way to pass along `currentFile`, but this is obviously not configuration. Is it OK to not have `currentFile` in the schema in `package.json`?

brycekahle:
The latest CI failure does not appear to be related.

src/debugAdapter/goDebug.ts
ramya-rao-a: Yes, its ok. In this case, this is an internal implementation. Current file can always be determined by VS Code using the `vscode.window.activeTextEditor`. It need not be provided by the user

ramya-rao-a:
Thanks @brycekahle!

PR #2124, opened 2018-11-13
Fix module enabled gocode

marwan-at-work:
A recent addition broke module-enabled gocode (gocode-gomod binary which is the github.com/stamblerre/gocode fork) 

Unrecognized flags cause errors, so this change only adds this flag to the binary that recognizes it (mdempsky/gocode)

### DIFF
--- a/src/goSuggest.ts
+++ b/src/goSuggest.ts
@@ -222,8 +222,12 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 			let stdout = '';
 			let stderr = '';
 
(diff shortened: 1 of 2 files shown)


src/goSuggest.ts
```diff
@@ -222,8 +222,17 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider {
 			let stdout = '';
 			let stderr = '';
 
+			const flags = [...this.gocodeFlags];
```
ramya-rao-a: You can work on `this.gocodeFlags` directly. No need to create a new array

Issue #1609, opened 2018-04-06
Support attaching and detaching from remote debug server without killing the process


radu-matei:
Related: https://github.com/Microsoft/vscode-go/issues/1599#issuecomment-377214763

If you have a long running process and start a `dlv attach --headless` session, then you point VS Code to that host and have a debug session, after you stop debugging from VS Code you would expect the remote process to continue running - but the extension actually sends a Halt request to the debug server.

It would be extremely helpful (for this and for other extensions using this Go debugger) to support gracefully detaching from remote processes without killing them.

lggomez:
A similar issue: if you launch a remote debug session and the headless server isn't up it seems that the debug task hangs forever waiting to connect

I thought that the client connection had a defined [timeout](https://github.com/Microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts#L400-L410), although it may be missing something

ramya-rao-a:
@radu-matei We do a halt + restart in that case and do not kill any process explicitly. I dont know why we went with halt + restart instead of detach without killing, as that was before my time. 

Sending a detach command needs halting as well, so we can look into halt + detach instead of halt + restart

Thoughts?

olegburov:
Hi @ramya-rao-a, is there any update on this issue?

It would be supper handy if I VS Code can accept input from `fmt.Scan` while debugging  such a simple code as follows:

``` go
package main

import "fmt"

func main() {
	fmt.Println("What's your name?")

	var s string
	fmt.Scanln(&s)

	fmt.Printf("Hello, %s", s)
}
```

ramya-rao-a:
@radu-matei 

> If you have a long running process and start a dlv attach --headless session, then you point VS Code to that host and have a debug session, after you stop debugging from VS Code you would expect the remote process to continue running - but the extension actually sends a Halt request to the debug server.
It would be extremely helpful (for this and for other extensions using this Go debugger) to support gracefully detaching from remote processes without killing them.

In order to send a Detach request, we need to first send a Halt request. See https://github.com/go-delve/delve/blob/4c9a72e486f1f0d0c90ecede8415a871dced8117/Documentation/api/ClientHowto.md#gracefully-ending-the-debug-session

@olegburov Your scenario is being tracked in https://github.com/Microsoft/vscode-go/issues/219. Please subscribe to that issue for further updates

ramya-rao-a:
[...]

Issue #1599, opened 2018-03-29
Debug Attach doesn't work


caibirdme:
If you have a question, please ask it on https://gitter.im/Microsoft/vscode-go or on the VSCode channel in Gophers Slack(https://invite.slack.golangbridge.org/)

If you have issues around GOPATH, read https://github.com/Microsoft/vscode-go/wiki/GOPATH-in-the-VS-Code-Go-extension first.

If you have issues debugging your Go programs
- read https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code first.
- set `"trace": "verbose"` and share the resulting logs in the debug console

If this is a bug report, please share 
- a sample code which can be used to reproduce the issue
- any Go related settings you might have added/changed

Steps to Reproduce:

1. set `launch.json`
```
        {
            "name": "Attach",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "remotePath": "${workspaceRoot}",
            "port": 2345,
            "host": "127.0.0.1",
            "program": "${workspaceRoot}",
            "env": {},
            "args": [],
            "showLog": true
        }
```
2. go build && run
3. get pid && dlv attach pid --headless --listen=:2345 --log
4. in VSCode,  Click begin debug button
5. Next button isn't available
<img width="606" alt="report" src="https://user-images.githubusercontent.com/8054803/38081699-3e150c16-3377-11e8-98a2-db9ba5d5c661.png">
6. change launch.json from "request":"attach" to "request":"launch", everything wil be OK but
7. Everytime I stop the debugging, it also stop the process which is not the same as attach

AlexRouSg:
Next is not available cause the program didn't hit a breakpoint, and it didn't hit a breakpoint cause it was already running when you attached and so it missed the first few breakpoints and so the first breakpoint it will hit from the screenshot is line 13 which will only hit when you send a http request after attatching.

caibirdme:
@AlexRouSg Yes I know it and I did send a request via `curl xxxxx` and didn't get the response. So I think it might be blocked at somewhere but In VSCode, nothing happened, and the next button was still unavailable

caibirdme:
Maybe something related to this?
My VSCode version 1.21.1, vscode-go version 0.6.77
<img width="449" alt="invalid" src="https://user-images.githubusercontent.com/8054803/38086378-aec5ed4a-3386-11e8-87b6-050c5d63084d.png">

AlexRouSg:
[...]

PR #2125, opened 2018-11-14
Supoort attach/detach remote headless server without kill process

BetaXOi:
Related: #1599 #1609 
The reason is that sending `halt` request twice when close `delve` session.

This PR add support for `"request": "attach"` in `launch.json`. You can attach a running process from `dlv --headless` remote server, and detach without kill the running process.

I'm not a `TypeScript` beginner even, so the code modified maybe ugly, but it works. It's wonderful if anyone improve the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -371,13 +371,13 @@
             }
           },
           {
-            "label": "Go: Attach to process",
+            "label": "Go: Attach to local process",
             "description": "Attach to an existing process by process ID",
             "body": {
               "name": "${1:Attach to Process}",
               "type": "go",
               "request": "attach",
-              "mode": "attach",
+              "mode": "local",
               "processId": 0
             }
           },
@@ -388,7 +388,7 @@
               "name": "${1:Connect to server}",
               "type": "go",
               "request": "attach",
-              "mode": "connect",
+              "mode": "remote",
               "remotePath": "^\"\\${workspaceFolder}\"",
               "port": 2345,
               "host": "127.0.0.1"
@@ -577,11 +577,11 @@
               },
               "mode": {
                 "enum": [
-                  "attach",
-                  "connect"
+                  "local",
+                  "remote"
                 ],
-                "description": "One of 'auto', 'debug', 'remote', 'test', 'exec'.",
-                "default": "auto"
+                "description": "Indicates local or remote debugging.  Local maps to the dlv 'attach' command, remote maps to 'connect'.",
+                "default": "local"
               },
               "stopOnEntry": {
                 "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -371,19 +371,27 @@
             }
           },
           {
+            "label": "Go: Attach to process",
+            "description": "Attach to an existing process by process ID",
+            "body": {
+              "name": "${1:Attach to Process}",
+              "type": "go",
+              "request": "attach",
+              "mode": "attach",
+              "processId": 0
+            }
+          },
+          {
             "label": "Go: Connect to server",
             "description": "Connect to a remote headless debug server",
             "body": {
-              "name": "${3:Connect to server}",
+              "name": "${1:Connect to server}",
               "type": "go",
-              "request": "launch",
-              "mode": "remote",
-              "remotePath": "^\"\\${workspaceFolder}${1:}\"",
+              "request": "attach",
+              "mode": "connect",
+              "remotePath": "^\"\\${workspaceFolder}\"",
               "port": 2345,
-              "host": "127.0.0.1",
-              "program": "^\"\\${workspaceFolder}${1:}\"",
-              "env": {},
-              "args": []
+              "host": "127.0.0.1"
             }
           }
         ],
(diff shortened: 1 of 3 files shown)


ramya-rao-a:
Thanks for the PR @BetaXOi! This is a very much needed feature.
Unfortunately, I am little tight on time, so it will take a while for me to look and test the changes. But I'll surely get to this in the next 2 weeks

jhendrixMSFT:
Rebased on top of master.

src/debugAdapter/goDebug.ts
```diff
@@ -276,162 +302,193 @@ class Delve {
...
-			}
+				const currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+				dlvArgs.push(mode || 'debug');
+				if (mode === 'exec') {
+					dlvArgs.push(program);
+				} else if (currentGOWorkspace && env['GO111MODULE'] !== 'on') {
+					dlvArgs.push(dirname.substr(currentGOWorkspace.length + 1));
+				}
+				dlvArgs.push('--headless=true', '--listen=' + launchArgs.host + ':' + launchArgs.port.toString());
+				if (!this.isApiV1) {
```
quoctruong: Since we already removed `isApiV1` in the json with `apiVersion`, should we just use that instead?

src/debugAdapter/goDebug.ts
```diff
@@ -201,6 +201,7 @@ interface DiscardedBreakpoint {
 
 // This interface should always match the schema found in `package.json`.
 interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
+	kind: 'launch';
```
quoctruong: Do we need an extra `kind` field? Can we just use the `request` field?

src/debugAdapter/goDebug.ts
```diff
@@ -276,162 +302,193 @@ class Delve {
...
-			this.loadConfig = launchArgs.dlvLoadConfig || {
-				followPointers: true,
-				maxVariableRecurse: 1,
-				maxStringLen: 64,
-				maxArrayValues: 64,
-				maxStructFields: -1
-			};
+				if (!existsSync(launchArgs.dlvToolPath)) {
+					log(`Couldn't find dlv at the Go tools path, ${process.env['GOPATH']}${env['GOPATH'] ? ', ' + env['GOPATH'] : ''} or ${envPath}`);
+					return reject(`Cannot find Delve debugger. Install from https://github.com/derekparker/delve & ensure it is in your Go tools path, "GOPATH/bin" or "PATH".`);
```
quoctruong: I think the path is now `https://github.com/go-delve/delve/`

src/debugAdapter/goDebug.ts
```diff
@@ -276,162 +302,193 @@ class Delve {
...
-				return;
-			}
+				const currentGOWorkspace = getCurrentGoWorkspaceFromGOPATH(env['GOPATH'], dirname);
+				dlvArgs.push(mode || 'debug');
+				if (mode === 'exec') {
+					dlvArgs.push(program);
+				} else if (currentGOWorkspace && env['GO111MODULE'] !== 'on') {
+					dlvArgs.push(dirname.substr(currentGOWorkspace.length + 1));
+				}
+				dlvArgs.push('--headless=true', '--listen=' + launchArgs.host + ':' + launchArgs.port.toString());
```
quoctruong: Can also use `--listen=${launchArgs.host}:${launchArgs.port}`.

src/debugAdapter/goDebug.ts
```diff
@@ -692,6 +761,25 @@ class GoDebugSession extends LoggingDebugSession {
...
+			return;
+		}
+		this.initLaunchAttachRequest(response, args);
+	}
+
+	protected attachRequest(response: DebugProtocol.AttachResponse, args: AttachRequestArguments): void {
+		args.kind = 'attach';
+		if (args.mode === 'attach' && !args.processId) {
+			this.sendErrorResponse(response, 3000, 'Failed to continue: the processId attribute is missing in the debug configuration in launch.json');
+		} else if (args.mode === 'connect' && !args.port) {
```
quoctruong: It seems like on line 686, we are assigning random ports to `args` if `args.port` is falsy. Do we need this check then?

src/debugAdapter/goDebug.ts
jhendrixMSFT: It's legacy but I was hesitant to remove it in case somebody was still relying on it.  @ramya-rao-a do you think we can finally remove this?

src/debugAdapter/goDebug.ts
jhendrixMSFT: Good point, fixed.

src/debugAdapter/goDebug.ts
jhendrixMSFT: Fixed all instances.

src/debugAdapter/goDebug.ts
jhendrixMSFT: Fixed.

src/debugAdapter/goDebug.ts
jhendrixMSFT: When attaching the user must specify the port to connect to.  The random port number is used in the `launch` scenario.

src/debugAdapter/goDebug.ts
ramya-rao-a: Its `useApiV1` that has been deprecated
`isApiV1` is a variable maintained only in this file

The support for `useApiV1` has been removed in this PR (look at the first few lines inside the constructor for the `Delve` class. I am fine with this, as there is appropriate messaging in place to tell the user what to do in case of version mismatch when remote debugging

src/debugAdapter/goDebug.ts
```diff
@@ -276,162 +302,193 @@ class Delve {
...
-				dlvArgs = dlvArgs.concat([program]);
-			} else if (currentGOWorkspace && env['GO111MODULE'] !== 'on') {
-				dlvArgs = dlvArgs.concat([dirname.substr(currentGOWorkspace.length + 1)]);
-			}
-			dlvArgs = dlvArgs.concat(['--headless=true', '--listen=' + host + ':' + port.toString()]);
-			if (!this.isApiV1) {
-				dlvArgs.push('--api-version=2');
-			}
+				if (!existsSync(launchArgs.dlvToolPath)) {
+					return reject(`Cannot find Delve debugger. Install from https://github.com/go-delve/delve & ensure it is in your Go tools path, "GOPATH/bin" or "PATH".`);
```
ramya-rao-a: We need this check for when request is launch as well

src/debugAdapter/goDebug.ts
```diff
@@ -692,6 +761,23 @@ class GoDebugSession extends LoggingDebugSession {
...
 
+	protected launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
+		if (!args.program) {
+			this.sendErrorResponse(response, 3000, 'Failed to continue: The program attribute is missing in the debug configuration in launch.json');
+			return;
+		}
+		this.initLaunchAttachRequest(response, args);
+	}
+
+	protected attachRequest(response: DebugProtocol.AttachResponse, args: AttachRequestArguments): void {
```
ramya-rao-a: When attaching, here, we are not expecting user to provide the `program` property, instead we are defaulting to `cwd` provided by the user which in turn defaults to `$workspacFolder`

The main use case for the `program` attribute  was to get the `localPath` which then is used to map to the remote path. Forcing this to be the `$workspaceFolder` can cause issues.

For example, say in the remote machine I have the code at `/Users/me/myproject/mysubfolder/mypackage`
And locally, this for me maps to `/Users/me2/myproject/mysubfolder/mypackage` and I have opened the folder "myproject" which makes `$workspaceFolder` resolve to `/Users/me2/myproject`

Now in the debugConfiguration, `remotePath` will point to `/Users/me/myproject/mysubfolder/mypackage`

Now the functions `toDebuggerPath` and `toLocalPath` won't work as expected

package.json
```diff
@@ -541,6 +549,143 @@
...
+            "required": [],
+            "properties": {
+              "processId": {
+                "type": "number",
+                "description": "The ID of the process to be debugged."
+              },
+              "mode": {
+                "enum": [
+                  "attach",
+                  "connect"
```
ramya-rao-a: Can we not re-use `debug` and `remote` here instead of introducing new modes?

src/debugAdapter/goDebug.ts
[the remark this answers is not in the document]
jhendrixMSFT: It's there on line 423.  Duplicated as there's some (questionable) logging in the other code path that's specific to launch.  If we can remove the logging I will consolidate.

package.json
jhendrixMSFT: These modes match their corresponding `dlv` commands.  `attach` is new and is to attach to a process by PID.  `connect` supersedes `launch/remote`.  We could rename `connect` to `remote` here (and fix it up in the code) to lessen the churn on the end-user though.

src/debugAdapter/goDebug.ts
jhendrixMSFT: This has been accounted for and works as expected (I've tested the remote case where Windows is the debugger and Linux is the debuggee; will test the inverse).

src/goDebugConfiguration.ts
```diff
@@ -71,7 +71,9 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 
 		const dlvConfig: { [key: string]: any } = goConfig.get('delveConfig');
 		if (!debugConfiguration.hasOwnProperty('useApiV1') && dlvConfig.hasOwnProperty('useApiV1')) {
-			debugConfiguration['useApiV1'] = dlvConfig['useApiV1'];
+			if (typeof dlvConfig['useApiV1'] === 'boolean' && dlvConfig['useApiV1'] === true) {
+				debugConfiguration['apiVersion'] = 1;
+			}
```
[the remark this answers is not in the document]
jhendrixMSFT: Good idea, fixed.

src/debugAdapter/goDebug.ts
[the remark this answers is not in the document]
ramya-rao-a: Got it

src/goDebugConfiguration.ts
```diff
@@ -70,8 +70,14 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 		});
 
 		const dlvConfig: { [key: string]: any } = goConfig.get('delveConfig');
-		if (!debugConfiguration.hasOwnProperty('useApiV1') && dlvConfig.hasOwnProperty('useApiV1')) {
-			debugConfiguration['useApiV1'] = dlvConfig['useApiV1'];
+		let useApiV1 = false;
+		if (debugConfiguration.hasOwnProperty('useApiV1')) {
+			useApiV1  = debugConfiguration['useApiV1'] === true;
```
quoctruong: nit: This and the line below can also be `useApiV1 = !!debugConfiguration['useApiV1']`.

quoctruong:
@jhendrixMSFT @ramya-rao-a Any update on this PR?

jhendrixMSFT:
Sorry for the delay, been busy with Go SDK work.  I've rebased on top of master.

minkir014:
Why not merging this PR?? It's very perfect.

jhendrixMSFT:
@ramya-rao-a how do you want to proceed?

package.json
ramya-rao-a: I would prefer renaming `connect` to `remote` so that in docs,  "remote debugging" and the `mode` being `remote` would be synonymous

ramya-rao-a: And we could print a warning in the debug console when `launch/remote` is being used, suggesting to use `attach/remote` instead

minkir014:
What's the full goal of this PR? When will it be merged?

jhendrixMSFT:
Ramya and I discussed this earlier today, I'm going to make a few small changes.  Should get merged later today/tomorrow.  Sorry for the delay.

package.json
```diff
@@ -370,20 +370,28 @@
...
+              "request": "attach",
               "mode": "remote",
-              "remotePath": "^\"\\${workspaceFolder}${1:}\"",
+              "remotePath": "^\"\\${workspaceFolder}\"",
               "port": 2345,
-              "host": "127.0.0.1",
-              "program": "^\"\\${workspaceFolder}${1:}\"",
-              "env": {},
-              "args": []
+              "host": "127.0.0.1"
```
ramya-rao-a: This wouldn't work without a process id right? If so, then we should add `processId` into this snippet

src/goDebugConfiguration.ts
```diff
@@ -82,6 +89,9 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 		if (!debugConfiguration.hasOwnProperty('showGlobalVariables') && dlvConfig.hasOwnProperty('showGlobalVariables')) {
 			debugConfiguration['showGlobalVariables'] = dlvConfig['showGlobalVariables'];
 		}
+		if (!debugConfiguration['cwd']) {
+			debugConfiguration['cwd'] = '${workspaceFolder}';
```
ramya-rao-a: this gets passed to delve for `--wd`. I am not sure if this will result in unexpected behavior for existing `launch` requests. I know we need to this to do the mapping between local and remote paths in case of attach/remote. Can we add this only for `attach` requests?

package.json
jhendrixMSFT: This is connecting to an existing dlv server so no process ID is required.

PR #2126, opened 2018-11-14
debug: fix 'pause' command

### LINKED ISSUE #978: Pause command doesn't work
After starting a debug session, if you hit pause button, nothing happens.
This is important if you want to add breakpoints after running the application.
I was debugging the plugin and saw that `pauseRequest` was never called. So I wonder whether this is a bug in this plugin or in the whole vscode.

xiphon:
Fixes #978 

**Addressed issues**
* Should maintain alive threads list and update it on 'ThreadsRequest'
* Should update `debugState` and `threads` on 'PauseRequest'
* Returning an empty array on `threads` request. VS Code expects that a process consists of at least one thread
* Unhandled exception in `handleReenterDebug` in case of `this.debugState.currentGoroutine` is unitialized
* Unhandled exception in `threadsRequest` when `this.debugState` is unitialized

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -1077,6 +1077,9 @@ class GoDebugSession extends LoggingDebugSession {
 				}
 				const goroutines = this.delve.isApiV1 ? <DebugGoroutine[]>out : (<ListGoroutinesOut>out).Goroutines;
 				this.updateThreads(goroutines);
+				if (!this.debugState.currentGoroutine && goroutines.length > 0) {
+					this.debugState.currentGoroutine = goroutines[0];
+				}
 
 				let stoppedEvent = new StoppedEvent(reason, this.debugState.currentGoroutine.id);
 				(<any>stoppedEvent.body).allThreadsStopped = true;
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -759,7 +759,7 @@ class GoDebugSession extends LoggingDebugSession {
 		}
 		log('ThreadsRequest');
 		this.delve.call<DebugGoroutine[] | ListGoroutinesOut>('ListGoroutines', [], (err, out) => {
-			if (this.debugState.exited) {
+			if (this.debugState && this.debugState.exited) {
 				// If the program exits very quickly, the initial threadsRequest will complete after it has exited.
 				// A TerminatedEvent has already been sent. Ignore the err returned in this case.
 				response.body = { threads: [] };
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -774,7 +774,7 @@ class GoDebugSession extends LoggingDebugSession {
 	protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
 		if (this.continueRequestRunning) {
 			// Thread request to delve is syncronous and will block if a previous async continue request didn't return
-			response.body = { threads: [] };
+			response.body = { threads: [new Thread(1, 'Dummy')] };
 			return this.sendResponse(response);
 		}
 		log('ThreadsRequest');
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Thanks for the PR @xiphon! This is a very much needed feature along with your work in #2128.
Unfortunately, I am little tight on time, so it will take a while for me to look and test the changes. But I'll surely get to this in the next 2 weeks

Nerzal:
@ramya-rao-a  did u have time to look at this ? :)

OneOfOne:
Any news about this?

ramya-rao-a:
@OneOfOne, @Nerzal, 

Not yet, just back from vacation and catching up on notifications.
Have either of you gotten a chance to try out the changes in this PR?

@segevfiner, @brycekahle, @lggomez Would you mind taking a look at this PR, since all 3 of you have worked in various debugging features of this extension?

OneOfOne:
I rebased it on current master and I can confirm this fixes the pause command, however I hit #2133.

jhendrixMSFT:
@xiphon can you please rebase on top of master?  Once that's done we can merge this change.

xiphon:
Rebased

src/debugAdapter/goDebug.ts
```diff
@@ -751,15 +751,35 @@ class GoDebugSession extends LoggingDebugSession {
 		});
 	}
 
+	private updateThreads(goroutines: DebugGoroutine[]): void {
```
ramya-rao-a: If I understand this correctly, what we are doing here is ensuring `this.threads` only has entries corresponding to existing goroutines. 

We should either rename this to something that communicates the above or add a comment above the function for the same

src/debugAdapter/goDebug.ts
```diff
@@ -751,15 +751,35 @@ class GoDebugSession extends LoggingDebugSession {
...
+			this.sendEvent(new ThreadEvent('exited', id));
+			this.threads.delete(id);
+		});
+	}
+
 	protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
 		if (this.continueRequestRunning) {
 			// Thread request to delve is syncronous and will block if a previous async continue request didn't return
-			response.body = { threads: [] };
+			response.body = { threads: [new Thread(1, 'Dummy')] };
```
ramya-rao-a: Can you give us some insight on why we have to create a dummy thread here?

ramya-rao-a:
@xiphon Can you answer the last 2 questions?

- We are now updating `this.threads` in the `threadRequest` and ensuring that it only has threads corresponding to current goroutines. What problem did this fix?
- In `handleReenterDebug`, if the current state doesnt have a `currentGoroutine`, we add the first goroutine to it. What problem did this fix? Why dont we need to update the `currentThread` here as well?

xiphon:
@ramya-rao-a 

> We are now updating `this.threads` in the `threadRequest` and ensuring that it only has threads corresponding to current goroutines. What problem did this fix?

Otherwise we will send to debugger a list containing non-existing/exited goroutines.

> In `handleReenterDebug`, if the current state doesnt have a `currentGoroutine`, we add the first goroutine to it. What problem did this fix?

Unhandled exception in `handleReenterDebug` in case of unitialized `currentGoroutine`, `currentGoroutine` might be unitialized.

> Why dont we need to update the `currentThread` here as well?

We always update `currentThread` before any `handleReenterDebug` call.

src/debugAdapter/goDebug.ts
xiphon: To be honest `updateThreads` sounds good to me and is exactly what this routine is doing: it updates the threads list and sends the appropriate notifications to VS Code through Events interface.

Can't be more accurate than `updateAndPropagateThreadsListToVsCode`, though sounds redundant to me.

ramya-rao-a:
> We always update currentThread before any handleReenterDebug call.

currentThread isnt being updated independently anywhere. It gets updated when they entire debug state is updated.

So now that we are specifically updating the goroutineId if the state doesnt have it, I was curious as to why we do that but not update the thread

PR #2128, opened 2018-11-15
debug: fix updating breakpoints while debuggee is running

### LINKED ISSUE #2002: Breakpoints added when debug session is not in a paused state, results in debug process not ending
Steps to Reproduce:

Every time I add a new breakpoint, the process will not be killed after restarting the debug.

macOS 10.13.6
vscode 1.28.0
vscode-go 0.6.91.

xiphon:
**Issue**
Can't set/remove breakpoints while debuggee is running

**Details**
Implemented **stop-breakpoint-start** approach to resolve the issue as described in https://github.com/Microsoft/vscode-go/issues/978#issuecomment-361015995.
Take a look at #978 for details, @muravjov's comments contain detailed description what's going wrong in that case.

Fixes #2002

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -534,7 +534,7 @@ class GoDebugSession extends LoggingDebugSession {
 
 	private _variableHandles: Handles<DebugVariable>;
 	private breakpoints: Map<string, DebugBreakpoint[]>;
-	private skipStopEventOnce: boolean; // Editing breakpoints requires halting delve, skip sending Stop Event to VS Code in such cases
+	private skipStopEventOnce: boolean;
 	private threads: Set<number>;
 	private debugState: DebuggerState;
 	private delve: Delve;
@@ -760,10 +760,10 @@ class GoDebugSession extends LoggingDebugSession {
 			this.skipStopEventOnce = true;
 			this.delve.callPromise('Command', [{ name: 'halt' }]).then(() => {
 				return this.setBreakPoints(response, args).then(() => {
-					return this.continue(true).then(null, err => {
-						logError(err);
-						this.sendErrorResponse(response, 2009, 'Failed to continue delve after halting it to set breakpoints: "{e}"', { e: err.toString() });
-					});
+					return this.continue();
+				}, err => {
+					logError(err);
+					return this.sendErrorResponse(response, 2009, 'Failed to continue delve after setting the breakpoint: "{e}"', { e: err.toString() });
 				});
 			}, err => {
 				logError(err);
@@ -1109,39 +1109,27 @@ class GoDebugSession extends LoggingDebugSession {
 
 	private continueEpoch = 0;
 	private continueRequestRunning = false;
-	private continue(calledWhenSettingBreakpoint?: boolean): Thenable<void> {
+	private continue(): void {
 		this.continueEpoch++;
 		let closureEpoch = this.continueEpoch;
 		this.continueRequestRunning = true;
-
-		const callback = (out) => {
-			if (closureEpoch === this.continueEpoch) {
-				this.continueRequestRunning = false;
-			}
-			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
-			log('continue state', state);
-			this.debugState = state;
-		};
-
-		// If called when setting breakpoint internally, we want the error to bubble up.
-		if (calledWhenSettingBreakpoint) {
-			return this.delve.callPromise('Command', [{ name: 'continue' }]).then(callback);
-		}
-		return this.delve.callPromise('Command', [{ name: 'continue' }]).then(callback, err => {
+		this.delve.call<DebuggerState | CommandOut>('Command', [{ name: 'continue' }], (err, out) => {
 			if (closureEpoch === this.continueEpoch) {
 				this.continueRequestRunning = false;
 			}
 			if (err) {
 				logError('Failed to continue - ' + err.toString());
 			}
+			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
+			log('continue state', state);
+			this.debugState = state;
+			this.handleReenterDebug('breakpoint');
 		});
 	}
 
 	protected continueRequest(response: DebugProtocol.ContinueResponse): void {
 		log('ContinueRequest');
-		this.continue().then(() => {
-			this.handleReenterDebug('breakpoint');
-		});
+		this.continue();
 		this.sendResponse(response);
 		log('ContinueResponse');
 	}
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@xiphon can you please rebase on top of master?

xiphon:
Rebased

src/debugAdapter/goDebug.ts
```diff
@@ -751,6 +752,23 @@ class GoDebugSession extends LoggingDebugSession {
...
+	protected setBreakPointsRequest(response: DebugProtocol.SetBreakpointsResponse, args: DebugProtocol.SetBreakpointsArguments): void {
+		log('SetBreakPointsRequest');
+		if (!this.continueRequestRunning) {
+			this.setBreakPoints(response, args);
+		} else {
+			this.skipStopEventOnce = true;
+			this.delve.callPromise('Command', [{ name: 'halt' }]).then(() => {
+				return this.setBreakPoints(response, args);
+			}, err => {
+				this.sendErrorResponse(response, 2002, 'Failed to set breakpoint: "{e}"', { e: err.toString() });
```
ramya-rao-a: `Failed to halt delve before attempting to set breakpoint` would be more informative

src/debugAdapter/goDebug.ts
```diff
@@ -751,6 +752,23 @@ class GoDebugSession extends LoggingDebugSession {
...
+				logError(err);
+			}).then(() => {
+				return this.continue();
```
ramya-rao-a: I believe this will get called even when the Halt request fails.
Is that what we want?
I assumed that we want to `continue` only if halting was successful

src/debugAdapter/goDebug.ts
```diff
@@ -1073,17 +1091,22 @@ class GoDebugSession extends LoggingDebugSession {
 					this.threads.delete(id);
 				});
 
+				if (this.skipStopEventOnce) {
+					this.skipStopEventOnce = false;
+					return;
```
ramya-rao-a: Can you explain why we need to do this?

xiphon: To silently skip the `pause` that we introduce just to set the breakpoint and `continue` right after.

src/debugAdapter/goDebug.ts
xiphon: Fixed

src/debugAdapter/goDebug.ts
xiphon: Updated

src/debugAdapter/goDebug.ts
```diff
@@ -743,14 +744,34 @@ class GoDebugSession extends LoggingDebugSession {
 			return breakpoints;
 		}).then(breakpoints => {
 			response.body = { breakpoints };
-			this.sendResponse(response);
 			log('SetBreakPointsResponse');
+			return this.sendResponse(response);
 		}, err => {
-			this.sendErrorResponse(response, 2002, 'Failed to set breakpoint: "{e}"', { e: err.toString() });
 			logError(err);
+			return this.sendErrorResponse(response, 2002, 'Failed to set breakpoint: "{e}"', { e: err.toString() });
```
ramya-rao-a: Any reason for the `return` here and in line 748?
Both `sendResponse` and `sendErrorResponse` return void.

Also, whats the reason for the swapping of lines?

src/debugAdapter/goDebug.ts
```diff
@@ -743,14 +744,34 @@ class GoDebugSession extends LoggingDebugSession {
...
+		if (!this.continueRequestRunning) {
+			this.setBreakPoints(response, args);
+		} else {
+			this.skipStopEventOnce = true;
+			this.delve.callPromise('Command', [{ name: 'halt' }]).then(() => {
+				return this.setBreakPoints(response, args).then(() => {
+					return this.continue();
+				}, err => {
+					logError(err);
+					return this.sendErrorResponse(response, 2009, 'Failed to continue delve after setting the breakpoint: "{e}"', { e: err.toString() });
```
ramya-rao-a: Since this error handler is for the continue operation, I would move it up like so:
```
this.delve.callPromise('Command', [{ name: 'halt' }]).then(() => {
   return this.setBreakPoints(response, args).then(() => {
     return this.continue().then(null, err => {
          logError(err);
          this.sendErrorResponse(response, 2009, 'Failed to continue delve after setting the breakpoint: "{e}"', { e: err.toString() });
        })
    }, null);
}, err => {
     logError(err)
     this.sendErrorResponse(response, 2008, 'Failed to halt delve before attempting to set breakpoint: "{e}"', { e: err.toString() });
  });
```

src/debugAdapter/goDebug.ts
xiphon: Reverted

src/debugAdapter/goDebug.ts
ramya-rao-a: Actually, on second thought, `this.continue()` doesnt return a promise.
Also, inside `this.continue()`, if the call to delve failed, it is logged and the error doesnt bubble up here.

ramya-rao-a:
@xiphon I have pushed a commit to handle the error from continue().

Please see https://github.com/Microsoft/vscode-go/pull/2128/commits/45e6b64893f9b3c4b0b090adde27ce2955b9facb.

If you are ok with this change, then I'll merge this PR.

OneOfOne:
this needs to be rebased against pull/2216

jhendrixMSFT:
Fixes https://github.com/Microsoft/vscode-go/issues/2002

xiphon:
> whether delve returns the state when it fails the "continue" request...
> I believe it doesnt

Right

Will update the PR to finally fix all the edge cases:
1. If `continue` fails, a user will end up with non-recoverable debug session state. Internally it will be paused, but VS Code UI will display it as `running`.
2. `setBreakPoints` sends response in both cases: on success and on error. But we also invoke `sendErrorResponse` if `continue` failes (https://github.com/Microsoft/vscode-go/pull/2128/files#diff-16e876ed2a6ddf6a530a935e37dddfdbR765).
If both `setBreakPoints` and `continue` fail, we send error response twice.
3. If `halt` fails, we have to revert `skipStopEventOnce` to `false`

xiphon:
Updated

ramya-rao-a:
@xiphon If the continue after the halting of delve fails, right now we are just logging the error. Shouldnt we send the error response too?

xiphon:
@ramya-rao-a 

If we reached `continue` step, we already sent `setBreakpoints` response to VS Code:, either successful one
https://github.com/Microsoft/vscode-go/pull/2128/files#diff-16e876ed2a6ddf6a530a935e37dddfdbL746
or error response
https://github.com/Microsoft/vscode-go/pull/2128/files#diff-16e876ed2a6ddf6a530a935e37dddfdbL749

ramya-rao-a:
But if the continue doesnt succeed, what will the user experience be?

Issue #2087, opened 2018-11-04
Rebuild after running `Go Get Package` cmd if current file has "package not found" errors


ramya-rao-a:
Add a package to the import block which you haven't downloaded yet. You will get the error `cannot find package...`

At this point, you can run the command `Go Get Package` with the cursor on the package in the import block, and the package will get downloaded and installed.

But the error still remains. We need to rebuild the file for the error to go away.

PRs are most welcome to fix this
Code Pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- File to update: `goGetPackage.ts`
- To build current package use the `buildCode()` function from `goBuild.ts` file.

mrtmrtmlck:
@ramya-rao-a Hi! May I pick this issue ?

ramya-rao-a:
The fix for this is now out in the latest update to the Go extension (0.8.0)
Thanks @Ashniu123!

PR #2129, opened 2018-11-15
Fix: Rebuild after running "Go Get Package"

Ashniu123:
Fixes issue #2087


Ashniu123:
Any idea why test should be failing?

Ashniu123:
On line 28, it is written that
```
// go get -v uses stderr to write output regardless of success or failure
```
From this I thought that if `go get` fails then error will remain, which is fine.
And if it is a success, build will be successful and error will go away.

Issue #1974, opened 2018-10-07
Wrong variable scope displayed when debugging


christian-boks:
The debugger shows variable content from the wrong scope: 

![image](https://user-images.githubusercontent.com/10370084/46580628-b50f9500-ca28-11e8-8bcf-c299bd42cfe3.png)

Program output:
Inner animal: cat
Outer animal: dog

Using: 
Delve Debugger
Version: 1.1.0
Build: $Id: 1990ba12450cab9425a2ae62e6ab988725023d5c $

And Go Extension v 0.6.90

ramya-rao-a:
We get the variable information from delve. 

Can you try running `dlv` from the command line to debug your code, run `locals` to get the local variables and see if `animal` points to `dog` or `cat`?

chainhelen:
![image](https://user-images.githubusercontent.com/7046329/46715363-3ddd3980-cc92-11e8-80fd-4532b539823a.png)

christian-boks:
This is the output I get when running `dlv`

```
> main.main() c:/gohome/src/play/scope_bug/main.go:9 (PC: 0x48df25)
     4:
     5: func main() {
     6:         animal := "dog"
     7:         if true {
     8:                 animal := "cat"
=>   9:                 fmt.Println("Inner animal:", animal)
    10:         }
    11:         fmt.Println("Outer animal:", animal)
    12: }
(dlv) locals
(animal) = "dog"
animal = "cat"
```

jhendrixMSFT:
I have a fix for this, one thing that comes to mind is if we should also display the shadowed value(s) like delve does in the variables pane.

ramya-rao-a:
I would stick to just showing the non-shadowed value.

jhendrixMSFT:
I tend to agree however there are cases where having both values would be useful.  IMO what would be ideal is if we could display the shadowed value when hovering over the shadowed variable.  Can we subscribe to some event when hovering over a variable?

jhendrixMSFT:
This starts to get a little unwieldy if a variable is shadowed more than once.  Perhaps it would be better if we displayed the shadowed variables as children of one another, what do you think (assuming vscode don't have problems with child vars with the same name)?

ramya-rao-a:
Shadowed variables are now showing inside parentheses in the latest update (0.9.0) to the Go extension.
Thanks @jhendrixMSFT!

PR #2130, opened 2018-11-16
debug: use unique stack frame id, fix local variables, evaluation errors

xiphon:


ramya-rao-a:
Thanks for the PR @xiphon!

@brycekahle Given that you had earlier worked in the `stackTraceRequest` area, can you please review this PR and try out the changes?

jhendrixMSFT:
@xiphon I tested this out using the repro provided in the linked issue but it doesn't appear to work, i.e. the value for `animal` is still `dog` when stopped after `animal` is shadowed with the value `cat`.  Am I missing something?

jhendrixMSFT:
It is, I thought the intent of the PR was to address it but perhaps I'm mistaken.  I had dug into it a bit, delve sends back all the locals including the shadows ones.  E.g.
```
"pet": "dog",
"pet": "cat"
```
We send all of them back to VS Code and it would appear that when there are duplicate entries the first one is displayed and the rest discarded (this is just a guess on my part but fits with my observation).
One idea is to prune out the dupes based on their line number with respect to the current line number (I actually have code for this).  Another option would be to display the shadowed variables the same way delve does which might be useful (we could also make this option configurable).

xiphon:
Yep, seems this PR is not related to #1974. Updated the description.

It still fixes the problem - we have to use unique stack frame identifiers.

Btw, there is another PR attempting to fix the same issue https://github.com/Microsoft/vscode-go/pull/2208

segevfiner:
We are not resetting `stackFrameHandles` ever. This will cause memory consumption to infinitely raise.

ramya-rao-a:
Good point, we can reset both stackFramesHandles and the variablesHandles in the disconnectRequest() call

ramya-rao-a:
@segevfiner Are you referring to resetting the handles in between 2 debugging sessions or inside a single debugging session?

segevfiner:
Both are an issue technically. During one debug session you keep accumlating handles, and handles from a previous session remain after it ends (Doesn't VS Code close the debug adapter though?)

Another point is that changing the thread ID, frame ID might cause VS Code to consider this a new thread which can make it lose focus. I didn't check whether this PR has such an issue or not.

jhendrixMSFT:
@segevfiner the debug adapter process is recycled between sessions however you're correct that the number of handles will continue to grow.
Taking a step back, can somebody articulate why this fix is needed?  Not that I'm pushing back (although we do need to address the handle growth), I don't see a referenced issue so I'm missing some context.

ramya-rao-a:
@jhendrixMSFT When investigating another issue, @segevfiner came across documentation for the debug adapter protocol that the stackframe id should be unique across all threads. See https://github.com/Microsoft/vscode-go/issues/2133#issuecomment-449671155

That's the motivation for @segevfiner's attempt to fix the problem.

I believe @xiphon encountered the same when investigating the variable shadowing problem

Though, in itself, this PR doesnt fix any user facing issue we are aware of, this bug could be the reason for some other issue in the future (or present)

jhendrixMSFT:
Digging into this a bit more we should be resetting `stackFrameHandles` and `_variableHandles` every time the debugger stops as the information is no longer valid and is essentially a memory leak; this is exactly what the node debugger does, see [here](https://github.com/Microsoft/vscode-node-debug/blob/master/src/node/nodeDebug.ts#L751).  I think the best place to do this is in `handleReenterDebug()` unless there's a better suggestion.

xiphon:
@jhendrixMSFT you are absolutely correct. Once we stop there, previous handles are no longer valid, `handleReenterDebug` is the right place. Updated.

OneOfOne:
@xiphon can you rebase against master please.

xiphon:
@OneOfOne 
Rebased

Issue #2121, opened 2018-11-13
Add support for "Go to Type Definition"/"Peek Type Definition"


segevfiner:
Implement [TypeDefinitionProvider](https://code.visualstudio.com/docs/extensionAPI/vscode-api#TypeDefinitionProvider`) which would add support for "Go to Type Definition" and "Peek Type Definition". This would be really helpful for Go, which uses type inference a lot, so you rarely have a type to click on for "Go to Definition"/"Peek Definition".

ramya-rao-a:
Doesnt the existing `Go to Implementation` command help do the same? It uses `guru implements`.

segevfiner:
> Doesnt the existing `Go to Implementation` command help do the same? It uses `guru implements`.

I think the "Go to Implementation" command as implemented will go to the implementation of an interface/method. If you use it on a variable it will just fail.

What is asked here is being able to go to the definition of the type of the symbol under the cursor. Go to Type Definition on a struct variable should go to where that struct type is defined.

ramya-rao-a:
Got it, thanks!

segevfiner:
https://github.com/Microsoft/vscode-go/pull/2136 was merged. I guess this can be closed.

ramya-rao-a:
This feature is now available in the latest version of the Go extension (0.8.0)
Thanks @hummerd!

PR #2136, opened 2018-11-20
Go to Type Definition command implementation

hummerd:
Go to Type Definition command implementation. Works with latest guru [with this commit](https://github.com/golang/tools/commit/9c8bd463e3aca40b6288cd14f7b7703206ff743b).

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -11,6 +11,7 @@ import { GoHoverProvider } from './goExtraInfo';
 import { GoDefinitionProvider } from './goDeclaration';
 import { GoReferenceProvider } from './goReferences';
 import { GoImplementationProvider } from './goImplementations';
+import { GoTypeDefinitionProvider } from './goTypeDefinition';
 import { GoDocumentFormattingEditProvider } from './goFormat';
 import { GoRenameProvider } from './goRename';
 import { GoDocumentSymbolProvider } from './goOutline';
@@ -180,6 +181,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 	ctx.subscriptions.push(vscode.languages.registerCodeActionsProvider(GO_MODE, new GoCodeActionProvider()));
 	ctx.subscriptions.push(vscode.languages.registerCodeLensProvider(GO_MODE, testCodeLensProvider));
 	ctx.subscriptions.push(vscode.languages.registerCodeLensProvider(GO_MODE, referencesCodeLensProvider));
+	ctx.subscriptions.push(vscode.languages.registerTypeDefinitionProvider(GO_MODE, new GoTypeDefinitionProvider()));
 	ctx.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('go', new GoDebugConfigurationProvider()));
 
 	buildDiagnosticCollection = vscode.languages.createDiagnosticCollection('go');
--- a/src/goTypeDefinition.ts
+++ b/src/goTypeDefinition.ts
@@ -1,0 +1,95 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------*/
+
+'use strict';
+
+import vscode = require('vscode');
+import cp = require('child_process');
+import path = require('path');
+import { byteOffsetAt, getBinPath, canonicalizeGOPATHPrefix, getFileArchive, killTree } from './util';
+import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
+import { getToolsEnvVars } from './util';
+
+interface GuruDescribeOutput {
+	desc: string;
+	pos: string;
+	detail: string;
+	value: GuruDescribeValueOutput;
+}
+
+interface GuruDescribeValueOutput {
+	type: string;
+	value: string;
+	objpos: string;
+	typespos: GuruDefinitionOutput[];
+}
+
+interface GuruDefinitionOutput {
+	objpos: string;
+	desc: string;
+}
+
+export class GoTypeDefinitionProvider implements vscode.TypeDefinitionProvider {
+	provideTypeDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): vscode.ProviderResult<vscode.Definition> {
+		return new Promise<vscode.Location[]>((resolve, reject) => {
+			let goGuru = getBinPath('guru');
+			if (!path.isAbsolute(goGuru)) {
+				promptForMissingTool('guru');
+				return reject('Cannot find tool "guru" to find type definitions.');
+			}
+
+			let filename = canonicalizeGOPATHPrefix(document.fileName);
+			let offset = byteOffsetAt(document, position);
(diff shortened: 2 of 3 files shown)


src/goMain.ts
```diff
@@ -158,6 +159,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 			registerCompletionProvider(ctx);
 			ctx.subscriptions.push(vscode.languages.registerHoverProvider(GO_MODE, new GoHoverProvider()));
 			ctx.subscriptions.push(vscode.languages.registerDefinitionProvider(GO_MODE, new GoDefinitionProvider()));
+			ctx.subscriptions.push(vscode.languages.registerTypeDefinitionProvider(GO_MODE, new GoTypeDefinitionProvider()));
```
ramya-rao-a: We need this feature even when the language server is enabled. So. move this line down near line 182

src/goTypeDefinition.ts
```diff
@@ -0,0 +1,93 @@
...
+	desc: string;
+}
+
+export class GoTypeDefinitionProvider implements vscode.TypeDefinitionProvider {
+	provideTypeDefinition(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken): vscode.ProviderResult<vscode.Definition> {
+		return new Promise<vscode.Location[]>((resolve, reject) => {
+			let goGuru = getBinPath('guru');
+			if (!path.isAbsolute(goGuru)) {
+				promptForMissingTool('guru');
+				return reject('Cannot find tool "guru" to find references.');
```
ramya-rao-a: find "type definitions"

ramya-rao-a:
Thanks for the PR @hummerd!

@segevfiner Given that you logged #2121, can you try out these changes and give feedback?

src/goMain.ts
hummerd: You right! Done.

src/goTypeDefinition.ts
hummerd: Of course :)

hummerd:
Hmm, I don't understand github says ramya-rao-a requested changes, but I see no changes. What should I ?

segevfiner:
> Hmm, I don't understand - github says ramya-rao-a requested changes, but I see no changes in last build. What should I do?

It's what GitHub says once someone submits a review for your PR until he reviews it again. If you already fixed what was requested of you in the review, than you just need to wait for the code to be reviewed again.

src/goTypeDefinition.ts
```diff
@@ -0,0 +1,96 @@
...
+					if (err) {
+						return reject(err);
+					}
+
+					let guruOutput = <GuruDescribeOutput>JSON.parse(stdout.toString());
+					let results: vscode.Location[] = [];
+
+					if (!guruOutput.value || !guruOutput.value.typespos) {
+						if (guruOutput.value) {
+							console.log('No type definitions found: probably you have not latest version of guru');
```
segevfiner: We might want something more prominent for this. Maybe `promptForUpdatingTool`? I started debugging only to see that my guru is outdated.

But this should only happen when `guru` is truly outdated not on just any failure to find the type.

src/goTypeDefinition.ts
```diff
@@ -0,0 +1,96 @@
...
+						}
+						return resolve(null);
+					}
+
+					guruOutput.value.typespos.forEach(ref => {
+						let match = /^(.*):(\d+):(\d+)/.exec(ref.objpos);
+						if (!match) return;
+						let [_, file, lineStartStr, colStartStr] = match;
+						let referenceResource = vscode.Uri.file(path.resolve(cwd, file));
+						let range = new vscode.Range(
```
segevfiner: I think we can just use a [`vscode.Position`](https://code.visualstudio.com/docs/extensionAPI/vscode-api#Position) if we don't know the full range of the type definition. See: [`vscode.Location`](https://code.visualstudio.com/docs/extensionAPI/vscode-api#Location).

src/goTypeDefinition.ts
```diff
@@ -0,0 +1,96 @@
...
+							console.log('No type definitions found: probably you have not latest version of guru');
+						}
+						return resolve(null);
+					}
+
+					guruOutput.value.typespos.forEach(ref => {
+						let match = /^(.*):(\d+):(\d+)/.exec(ref.objpos);
+						if (!match) return;
+						let [_, file, lineStartStr, colStartStr] = match;
+						let referenceResource = vscode.Uri.file(path.resolve(cwd, file));
```
segevfiner: You are considering the directory of the file to be the `cwd` which is set above to be the directory where the file we invoke this on is. But you run guru with the working directory of the extension host. If guru does output relative paths in some cases this can cause a failure to calculate the correct path. guru seems to output absolute paths for me though.

src/goTypeDefinition.ts
hummerd: I will make check if type property exist but typespos don't - this means that we have old version and should call promptForUpdatingTool.

src/goTypeDefinition.ts
hummerd: Of course. Changed range to position.

src/goTypeDefinition.ts
hummerd: Removed resolve, since I think guru will always return absolute paths.

src/goTypeDefinition.ts
```diff
@@ -0,0 +1,95 @@
...
+			}
+
+			let filename = canonicalizeGOPATHPrefix(document.fileName);
+			let offset = byteOffsetAt(document, position);
+			let env = getToolsEnvVars();
+			let buildTags = vscode.workspace.getConfiguration('go', document.uri)['buildTags'];
+			let args = buildTags ? ['-tags', buildTags] : [];
+			args.push('-json', '-modified', 'describe', `${filename}:#${offset.toString()}`);
+
+			let process = cp.execFile(goGuru, args, { env }, (err, stdout, stderr) => {
```
ramya-rao-a: To keep `guru implements` fast, we restrict the scope to current workspace in https://github.com/Microsoft/vscode-go/blob/0.7.0/src/goImplementations.ts#L32 which implements the `Go to Implementation` feature.

Would the same apply here as well?

hummerd: My measurement results says that scope doesn't affect describe's performance. I think describe looking for type definition from imports only - there is no need to analyze whole workspace and no need in scope.

hummerd: @ramya-rao-a I can try it on some big project from github and show measurements result. Should I?

ramya-rao-a: @hummerd Yes please :) If you can try on some big projects with and without the scope and then share the results, that would be great

hummerd: So I tried it on kubernetes. I tried:
```
# from go/src/k8s.io/kubernetes/pkg/kubectl
time guru -json describe apply.go:#3880
time guru -json -scope . describe apply.go:#3880
time guru -json -scope k8s.io/kubernetes/pkg/kubectl describe apply.go:#3880
```
All gives same results (on my machine mean real ~ 2.03s)

ramya-rao-a: Thanks @hummerd!

Issue #1484, opened 2018-01-26
Adopt VSCode 1.20's new CodeAction type 


mjbvz:
VS Code 1.20 adds a new [`CodeAction`](https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L1897) type that attaches additional metadata to code actions. This additional information is already being used in a few places:

* `CodeAction.kind` describes the kind of a code action (extract refactoring, quickfix, ...) Kinds are used for VS Code's new `refactor` command, and they are used to [setup keybindings for specific code actions](https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_20.md#keybindings-for-quick-fixes-and-code-actions)

* `CodeAction.diagnostics` tells which errors a quickfix addresses. This is used to order quickfixes by severity

I suggest that you migrate this extension to use the new `CodeAction` type in order to enable support for these features.

Please let me know if you have any questions about the new API

ramya-rao-a:
We now have refactor code actions which follow this model.

There is only one diagnostic for which we provide code action (the missing import), am adding some telemetry around this to check the usage as I have a hunch this is not used much because formatting in Go adds the missing imports.

BertoMejia:
Are there any plans on adding other code actions? I use this all the time in other languages. It is extremely useful! It was a bit disappointing to see that it is non included in this Go extension.

Is there a UserVoice or similar feature request system that we can vote this up on?

ramya-rao-a:
Any further work on the Code Actions will be taken up via the language server work that is going on.
No new work on this front is planned for the extension itself.

cc @stamblerre

PR #2139, opened 2018-11-22
Add command and code action to extract into function and local variable

aswinmprabhu:


Nerzal:
Thanks for the PR! This is one of my most wanted features :D 

Any Updates on this? 
As far as i see, only the OSX Build for go 1.11 failed. That shouldn't be a big issue?

aswinmprabhu:
This approach using godoctor has been tried before and was abandoned. But I guess it has gotten better and vim-go uses it. It works for simple use cases that I tried. I am waiting for the maintainer to review it or give me some feedback.

Nerzal:
@ramya-rao-a  sorry for the ping, but did u allready notice this pr? :)

src/goRefactor.ts
```diff
@@ -0,0 +1,31 @@
...
+			vscode.CodeActionKind.RefactorExtract
+		);
+		extractFunction.command = {
+			title: 'Extract to function in package scope',
+			command: 'go.godoctor.extract',
+		};
+		extractVar.command = {
+			title: 'Extract to variable in local scope',
+			command: 'go.godoctor.var',
+		};
```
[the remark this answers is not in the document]
ramya-rao-a: Also, shouldnt we set the `kind` property as well?
See https://code.visualstudio.com/docs/editor/refactoring and #1484

src/goDoctor.ts
```diff
@@ -0,0 +1,136 @@
...
+			'extract',
+			functionName,
+		];
+		let p = cp.execFile(godoctor, args, { env: getToolsEnvVars(), cwd: dirname(activeEditor.document.fileName) }, (err, stdout, stderr) => {
+			if (err && (<any>err).code === 'ENOENT') {
+				promptForMissingTool('godoctor');
+				return resolve('Could not find godoctor');
+			}
+			if (err) {
+				return resolve(`Could not extract function : \n\n${stderr}`);
```
ramya-rao-a: We should use `reject` here instead of `resolve` to deal with errors

src/goDoctor.ts
```diff
@@ -0,0 +1,136 @@
...
+		if (typeof functionName === 'undefined') {
+			return resolve('Function Name is undefined');
+		}
+		let args = [
+			'-w',
+			'-pos',
+			`${selection.start.line + 1},${selection.start.character + 1}:${selection.end.line + 1},${selection.end.character + 1}`,
+			'-file',
+			activeEditor.document.fileName,
+			'extract',
```
ramya-rao-a: I see that this `extract` is the only argument that changes when calling `godoctor` for function vs variable.

I would suggest to have a single `runGoDoctor` function instead of the two `runGoDoctorExtract` and `runGoDoctorVar`. Use an argument to differentiate between the 2 cases.

src/goDoctor.ts
```diff
@@ -0,0 +1,136 @@
...
+*/
+export function runGoDoctorExtract(functionName: string, selection: vscode.Selection, activeEditor: vscode.TextEditor): Thenable<string> {
+	let godoctor = getBinPath('godoctor');
+
+	return new Promise((resolve, reject) => {
+		if (typeof functionName === 'undefined') {
+			return resolve('Function Name is undefined');
+		}
+		let args = [
+			'-w',
```
ramya-rao-a: Is this for directly updating the file on the disk?
If yes, then we should look at working with diffs instead. Otherwise, Undo feature will not work in VS Code

ramya-rao-a: We can go with `-w` as part of this PR and have another to use the diffs

Nerzal:
Sadly i'm really bad at JavaScript, so sorry for the dumb question.
Will these features be usable through a shortcut? Something like ctrl+r+m (refactor + method - for extract method) and ctrl+r+v (refactor+variable).
If not, how could I (or someone) make these actions assignable to shortcuts?

ramya-rao-a:
@Nerzal Yes, I believe all the shortcuts as described in https://code.visualstudio.com/docs/editor/refactoring should be usable.

src/goDoctor.ts
aswinmprabhu: The type `Thenable<string>`  returned by `runGoDoctor` doesn't have a `catch` method to handle rejected promises.

src/goRefactor.ts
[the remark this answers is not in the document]
aswinmprabhu: It doesn't seem to have a `kind` property.

aswinmprabhu:
Also the refactoring seems to extremely slow in decently large files. This does not seem to be a problem with the implementation, as the godoctor command if run from the terminal with the same args, takes the same amount of time as in the editor.

helphi:
strong desire ！！！

src/goRefactor.ts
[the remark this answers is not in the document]
ramya-rao-a: Nevermind, my mistake. The `kind` is getting passed in the constructor for the `CodeAction`

aswinmprabhu:
Works good. But why do you still have the `-w` option if you are using diffs.

ramya-rao-a:
Looks like sometime in the past year, VS Code is able to do the undo feature even for the changes that were made directly to the file on disk. (It didn't do this when I was working a feature 2 years ago).

So we don't need to use the diffs after all. Reverting that commit

aswinmprabhu:
Ok. Cool. Tested it out again. Can confirm that undo works fine.

ramya-rao-a:
The latest [beta version](https://github.com/Microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) (0.9.3-beta.2) of this extension has this feature. Please see [Refactoring in VS Code](https://code.visualstudio.com/docs/editor/refactoring) on how you can use the refactor feature.

Thanks @aswinmprabhu!

helphi:
it seems that not support go mod.

ramya-rao-a:
Thanks for the observation @helphi 
I have updated https://github.com/golang/go/issues/24661 with a request to track godoctor for module support

PR #2158, opened 2018-11-28
Add support for bingo & golsp language server (#2143)

### LINKED ISSUE #2143: please support more go langserver tools such as bingo, golsp
Now, there are some new go language server tool, can you support them?

[bingo](https://github.com/saibing/bingo)

[golsp](https://github.com/golang/tools/blob/master/cmd/golsp/main.go)

tohjustin:
Hi @ramya-rao-a, here's my attempt to add support for [bingo](https://github.com/saibing/bingo) & [golsp](https://github.com/golang/tools/blob/master/cmd/golsp/main.go) language servers to the extension.

This Closes #2143.

**Additional Notes**

* When testing the extension with [bingo](https://github.com/saibing/bingo) as the language server & I would receive `<MY_PROJECT_DIR>/go.mod does not exist, please use 'go mod init' to create it` errors (since [bingo](https://github.com/saibing/bingo) only supports Go Module based projects).
* Do we want to support any language server specific flags/logic to handle cases like the example above?

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -5,13 +5,13 @@
 
 'use strict';
 
-import vscode = require('vscode');
+import cp = require('child_process');
 import fs = require('fs');
 import path = require('path');
-import cp = require('child_process');
+import vscode = require('vscode');
+import { goLiveErrorsEnabled } from './goLiveErrors';
 import { showGoStatus, hideGoStatus, outputChannel } from './goStatus';
 import { getBinPath, getToolsGopath, getGoVersion, SemVersion, isVendorSupported, getCurrentGoPath, resolvePath } from './util';
-import { goLiveErrorsEnabled } from './goLiveErrors';
 
 let updatesDeclinedTools: string[] = [];
 let installsDeclinedTools: string[] = [];
@@ -40,6 +40,8 @@ const _allTools: { [key: string]: string } = {
 	'golangci-lint': 'github.com/golangci/golangci-lint/cmd/golangci-lint',
 	'revive': 'github.com/mgechev/revive',
 	'go-langserver': 'github.com/sourcegraph/go-langserver',
+	'golsp': 'golang.org/x/tools/cmd/golsp',
+	'bingo': 'github.com/saibing/bingo',
 	'dlv': 'github.com/derekparker/delve/cmd/dlv',
 	'fillstruct': 'github.com/davidrjenni/reftools/cmd/fillstruct',
 };
@@ -85,7 +87,7 @@ function getTools(goVersion: SemVersion): string[] {
 		'dlv'
 	];
 
-	// gocode-gomod needed in go 1.11 & higher
+	// gocode-gomod needed in go1.11 & higher
 	if (!goVersion || (goVersion.major === 1 && goVersion.minor >= 11)) {
 		tools.push('gocode-gomod');
 	}
@@ -93,7 +95,7 @@ function getTools(goVersion: SemVersion): string[] {
 	// Install the doc/def tool that was chosen by the user
 	if (goConfig['docsTool'] === 'godoc') {
 		tools.push('godef');
-		// godef-gomod needed in go 1.11 & higher
+		// godef-gomod needed in go1.11 & higher
 		if (!goVersion || (goVersion.major === 1 && goVersion.minor >= 11)) {
 			tools.push('godef-gomod');
 		}
@@ -134,8 +136,16 @@ function getTools(goVersion: SemVersion): string[] {
 		tools.push('revive');
 	}
 
-	if (goConfig['useLanguageServer']) {
+	// Install Go language server selected by the user
+	if (goConfig['useLanguageServer'] === 'go-langserver') {
 		tools.push('go-langserver');
+	} else if (goConfig['useLanguageServer'] === 'golsp') {
+		tools.push('golsp');
+	} else if (
+		goConfig['useLanguageServer'] === 'bingo' &&
+		(!goVersion || (goVersion.major === 1 && goVersion.minor >= 11)) // bingo requires go1.11 & higher
+	) {
+		tools.push('bingo');
 	}
 
 	if (goLiveErrorsEnabled()) {
@@ -182,7 +192,9 @@ export function installAllTools(updateExistingToolsOnly: boolean = false) {
 		'megacheck': '\t(Linter)',
 		'golangci-lint': 'Linter)',
 		'revive': '\t\t(Linter)',
-		'go-langserver': '(Language Server)',
+		'go-langserver': '(Language Server by Sourcegraph)',
+		'golsp': '(Language Server by Google)',
+		'bingo': '(Language Server by saibing)',
 		'dlv': '\t\t\t(Debugging)',
 		'fillstruct': '\t\t(Fill structs with defaults)'
 	};
@@ -401,7 +413,11 @@ export function installTools(missing: string[], goVersion: SemVersion) {
 		outputChannel.appendLine(''); // Blank line for spacing
 		let failures = res.filter(x => x != null);
 		if (failures.length === 0) {
-			if (missing.indexOf('go-langserver') > -1) {
+			if (
+				missing.indexOf('go-langserver') > -1 ||
+				missing.indexOf('golsp') > -1 ||
+				missing.indexOf('bingo') > -1
+			) {
 				outputChannel.appendLine('Reload VS Code window to use the Go language server');
 			}
 			outputChannel.appendLine('All tools successfully installed. You\'re ready to Go :).');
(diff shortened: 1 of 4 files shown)


ramya-rao-a:
You can use the [isModSupported](https://github.com/Microsoft/vscode-go/blob/0.7.0/src/goModules.ts#L27) function to figure out if the current workspace supports modules or not. Based on this, you can then decide whether to use bingo or not.

tohjustin:
> You can use the [isModSupported](https://github.com/Microsoft/vscode-go/blob/0.7.0/src/goModules.ts#L27) function to figure out if the current workspace supports modules or not. Based on this, you can then decide whether to use bingo or not.

Just want to clarify a little more on this, if the user configured their vscode settings to `useLanguageServer: 'bingo'` and
* if `isModSupported` returns `true`, we'll use bingo as usual
* if `isModSupported` returns `false`, we display a message notifying the user that bingo is used since it only supports Go Module based projects & the current project/workspace isn't --> treat the session as `useLanguageServer: 'none'`

ramya-rao-a:
Yes, thats right.

tohjustin:
Not sure if there's a better place to check `isModSupported` when using the `bingo` language server =S

* After adding `isModSupported()` inside [`checkLanguageServer`](https://github.com/Microsoft/vscode-go/pull/2158/commits/af37c17a37e7c8e0391a16241dc0c92bec850f99#diff-3f57b095642b0ee445d6cdbc4f60d751R527), the function is now asynchronous & `async/await` statements were added.
* Another alternative I can think of is to place the `isModSupported` checks in [`provideDocumentFormattingEdits`](https://github.com/Microsoft/vscode-go/pull/2158/commits/af37c17a37e7c8e0391a16241dc0c92bec850f99#diff-9b6a4d5d6f38dfa535b1641cadfd1a80R121) & [`provideCompletionItem`](https://github.com/Microsoft/vscode-go/pull/2158/commits/af37c17a37e7c8e0391a16241dc0c92bec850f99#diff-9b6a4d5d6f38dfa535b1641cadfd1a80R127) instead (possibly affect performance?).

src/goInstallTools.ts
```diff
@@ -506,28 +524,39 @@ function getMissingTools(goVersion: SemVersion): Promise<string[]> {
...
 		vscode.window.showInformationMessage('Reload VS Code window after installing the Go language server');
+		return false;
 	}
-	return langServerAvailable;
+
+	const activeFileUri = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.document.uri : null;
+	if (activeFileUri !== null) {
+		const isModSupportedResult = await isModSupported(activeFileUri);
+		if (languageServer === 'bingo' && !isModSupportedResult) {
+			vscode.window.showInformationMessage('The Go language server (bingo) does not support non-go module projects. Falling back to default behavior.');
```
anjmao: Check this commit https://github.com/saibing/bingo/commit/f9f872adc38d795840c3c354bd3a6c540f861205
Bingo already supports non-go module projects. I tested it in GOPATH without modules and it was working fine.

src/goInstallTools.ts
```diff
@@ -40,6 +42,8 @@ const _allTools: { [key: string]: string } = {
 	'golangci-lint': 'github.com/golangci/golangci-lint/cmd/golangci-lint',
 	'revive': 'github.com/mgechev/revive',
 	'go-langserver': 'github.com/sourcegraph/go-langserver',
+	'bingo': 'github.com/saibing/bingo',
+	'golsp': 'golang.org/x/tools/cmd/golsp',
```
ramya-rao-a: @stamblerre Would this be the right import path to use in `go get` for the language server from your team?

src/goInstallTools.ts
```diff
@@ -183,6 +195,8 @@ export function installAllTools(updateExistingToolsOnly: boolean = false) {
 		'golangci-lint': 'Linter)',
 		'revive': '\t\t(Linter)',
 		'go-langserver': '(Language Server)',
+		'bingo': '(Language Server)',
+		'golsp': '(Language Server)',
```
ramya-rao-a: To help users differentiate lets add some suffixes here like

- Language Server by Sourcegraph
- Language Server by Google
- Language Server by saibing

src/goInstallTools.ts
ramya-rao-a: Thanks for that info @anjmao!
@saibing Can you confirm that non-go module support is ready for use?

ramya-rao-a:
@tohjustin The README.md would also need an update. There is a section there on the language server usage

src/goInstallTools.ts
```diff
@@ -134,10 +138,18 @@ function getTools(goVersion: SemVersion): string[] {
 		tools.push('revive');
 	}
 
-	if (goConfig['useLanguageServer']) {
+	if (goConfig['useLanguageServer'] === 'go-langserver') {
 		tools.push('go-langserver');
 	}
 
+	if (goConfig['useLanguageServer'] === 'bingo') {
```
ramya-rao-a: bingo is only supported from Go 1.11 onwards, so we should include that check here

src/goInstallTools.ts
tohjustin: @anjmao Thanks for the heads up! Kudos to @saibing for adding support for non-go module projects in such a short period of time 😅

@ramya-rao-a I guess I can remove the `isModSupported` checks?

src/goInstallTools.ts
tohjustin: Ah missed that out, will add the check it in the following patch.

src/goInstallTools.ts
tohjustin: Done

src/goInstallTools.ts
tohjustin: Done

tohjustin:
> @tohjustin The README.md would also need an update. There is a section there on the language server usage

Reworded quite a bit of the language server section (also ran it through Grammarly), please do let me know if you think it still needs more work 😅

src/goInstallTools.ts
ramya-rao-a: @tohjustin Yes, I just tested it out with GOPATH, and it works. So we can remove the `isModSupported` checks.

tohjustin: Done

tohjustin:
Hi @ramya-rao-a , I saw the issue opened by @stamblerre in https://github.com/saibing/bingo/issues/13 & seems like the plan is to focus on [golsp](https://golang.org/x/tools/cmd/golsp) development.

Any updates on this PR? 😅

ramya-rao-a:
@tohjustin Given that there is a good likelihood of all 3 language servers merging into 1 (Sourcegraph plans to use the one from Google, @saibing and Google are planning to work together, I want to avoid changing the schema of the `useLanguageServer` setting.

Instead, if one wants to use another language server, they can use the `go.alternateTools` setting as below:
```
"go.alternateTools": {
     "go-langserver": "bingo"
}
```

This would need some changes from the extension which I have already made in https://github.com/Microsoft/vscode-go/commit/ecb691c9a626ac9801bf5a59d8874428e3a8e61d

> Can we make go.languageServerExperimentalFeatures autocomplete and code formatting to be true by default? 

@anjmao Yes, we can do that

ramya-rao-a:
@anjmao On second thoughts, I don't want to enable the formatting feature by the language server by default just yet. This is because the language server doesn't respect the formatFlags or the formalTool settings. 

I don't want to enable the auto-completion feature by language server by default, because the language server doesn't support completions for unimported packages yet.

Once these 2 features from the language server has complete parity with what the extension does, I will turn them on by default.

tohjustin:
@ramya-rao-a Really appreciate the update! Yup the the `go.alternateTools` setting is working as intended. Got `bingo` to work with the extension using the following settings:
```

**Comment by tohjustin** [comment]
@ramya-rao-a Really appreciate the update! 

Yup the the `go.alternateTools` setting is working as intended. Got `bingo` to work with the extension using the following settings (also needed `go.useLanguageServer` set as `true`):

```
  "go.alternateTools": {
    "go-langserver": "bingo",
  },
  "go.languageServerExperimentalFeatures": {
    "format": true,
    "autoComplete": true
  },
  "go.useLanguageServer": true
```

I guess I'll leave it up to you to close this PR or put it on hold until the new Go language server is ready. 👍

ramya-rao-a:
> Yup the the go.alternateTools setting is working as intended.

Thanks for confirming the beta version @tohjustin!

I have updated the wiki on [Go modules support in VS Code](https://github.com/Microsoft/vscode-go/wiki/Go-modules-support-in-Visual-Studio-Code) appropriately.


> I guess I'll leave it up to you to close this PR or put it on hold until the new Go language server is ready.

I believe closing this PR is the right thing to do as I want to avoid changes in settings schema as much as possible.

Thanks for all your work on this @tohjustin, it is very much appreciated.

PR #2195, opened 2018-12-18
Added a check to prevent installing Delve in most unsupported platforms.

aggressivepixels:
This will prevent Delve from being installed in 32-bit machines (e.g. arm, i686) where is not supported.
The mips, mipsel, s390 and s390x entries are there because Node defines them as possible values and it's not clear if they are 64-bit or not, so the extension will try to install it anyway.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -81,9 +81,15 @@ function getTools(goVersion: SemVersion): string[] {
 		'go-outline',
 		'go-symbols',
 		'guru',
-		'gorename',
-		'dlv'
+		'gorename'
 	];
+
+	// Check if the system supports dlv (e.g. is 64-bit)
+	// There doesn't seem to be a good way to check if the mips and s390
+	// families are 64-bit, so just try to install it and hope for the best
+	if (process.arch.match(/^(arm64|mips|mipsel|ppc64|s390|s390x|x64)$/)) {
+		tools.push('dlv');
+	}
 
 	// gocode-gomod needed in go 1.11 & higher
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
@jhendrixMSFT The current possible values for `process.arch` are: 'arm', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 's390', 's390x', 'x32', and 'x64'.

See https://nodejs.org/api/process.html#process_process_arch

I am not familiar with 'mips', 'mipsel', 's390', 's390x'.
Any idea if it is useful to include these as supported platforms for delve?

jhendrixMSFT:
s390x and ppc64 are natively supported by Go (i.e. there are binary packages for it) so we should probably include those two.

ramya-rao-a:
> s390x and ppc64 are natively supported by Go (i.e. there are binary packages for it) so we should probably include those two.

But that doesnt necessarily mean delve is supported too right?

jhendrixMSFT:
I couldn't find any documentation about it but given that there are Linux distributions (e.g. Ubuntu) that support these architectures it seems likely that delve would work there too.  Given that these architectures are pretty uncommon it's probably fine to omit them (we can always add them later if it's an issue).

aggressivepixels:
Sorry about the inactivity, I don't have good internet access right now.
Anyway, I couldn't find anything about which platforms are supported by Delve, the only thing I found was that it doesn't work on 32-bit machines. Those other platforms are pretty uncommon and also there are no official Code builds for them AFAIK, so this should work well for now, and just like @jhendrixMSFT said, we can always add them later if it's an issue.

Issue #1010, opened 2017-05-27
Debugger does not display values in array of struct on VARIABLES pane and debug: Evaluate command.


YoskeOkamoto2:
I would like to inspect variables in array of struct.
I know 'Add to Watch' works fine for this request. but this is not a convenient way.

Here is the sample I wrote.
```go
package main

import (
	"fmt"
)

// ItemStruct is item
type ItemStruct struct {
	Num1 int
	Num2 int
}

// OwnerStruct holds item list and pointer list
type OwnerStruct struct {
	ItemList        []ItemStruct
	ItemPointerList []*ItemStruct
}

func main() {
	// initialize part
	list := make([]ItemStruct, 0, 2)
	plist := make([]*ItemStruct, 0, 2)
	for i := 0; i < 2; i++ {
		item := ItemStruct{Num1: i, Num2: i + 1}
		list = append(list, item)
		plist = append(plist, &item)
	}
	owner := OwnerStruct{
		ItemList:        list,
		ItemPointerList: plist}

	// watch variable here on debugger
	fmt.Println(owner)
}
```

and the screenshot.
![image](https://cloud.githubusercontent.com/assets/7908644/26517870/181180ca-42dd-11e7-9201-2706595c8efc.png)

ramya-rao-a:
Thanks for reporting @yoskeoka 
And thanks for the sample code as well, I can repro this.

@roblourens Can this be done?

roblourens:
Looking at the code, it's not obvious whether we can do it without evaluating all the variables like the watches window does. I'm not sure what other info we can get out of Delve.

Yussak:
Hi, I am bumping this up because being able to see the content of the variables we are debugging is actually important :)

Thanks for taking the time to look into this!

christian-boks:
Getting this fixed would be really nice. It is one of the small things that would make Go debugging support feel as complete as the Javascript debugging support. Which is just incredible.

ramya-rao-a:
If anyone wants to take a stab at this, the place to get started is the `variablesRequest` method in the `goDebug.ts` file. For setup, please see the simple instructions in the [README for debugging](https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter)

ramya-rao-a:
Nested content is now shown in the latest update (0.9.0) to the Go extension.
Huge thanks to @jhendrixMSFT who made this happen!

PR #2198, opened 2018-12-19
debug: Display nested content in the variables pane

jhendrixMSFT:
If a variable contains children but they have not been loaded issue an
expression command to delve to fetch them, then update the variable's
children array with the result.  This required refactoring the
evaluateRequest method into two parts so that delve expressions can be
evaluated synchronously.
Removed unreferenced param in convertDebugVariableToProtocolVariable.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -158,6 +158,7 @@ interface DebugVariable {
 	realType: string;
 	kind: GoReflectKind;
 	flags: GoVariableFlags;
+	onlyAddr: boolean;
 	DeclLine: number;
 	value: string;
 	len: number;
@@ -890,6 +891,7 @@ class GoDebugSession extends LoggingDebugSession {
 					realType: '',
 					kind: 0,
 					flags: 0,
+					onlyAddr: false,
 					DeclLine: 0,
 					value: '',
 					len: 0,
@@ -940,6 +942,7 @@ class GoDebugSession extends LoggingDebugSession {
 							realType: '',
 							kind: 0,
 							flags: 0,
+							onlyAddr: false,
 							DeclLine: 0,
 							value: '',
 							len: 0,
@@ -1043,18 +1046,20 @@ class GoDebugSession extends LoggingDebugSession {
 		let vari = this._variableHandles.get(args.variablesReference);
 		let variablesPromise: Promise<DebugProtocol.Variable[]>;
 		const loadChildren = async (exp: string, v: DebugVariable) => {
-			// for string types the value of v.len is the string's length so skip evaluation
-			if (v.type !== 'string' && v.len > 0 && v.children.length === 0) {
-				// from https://github.com/derekparker/delve/blob/master/Documentation/api/ClientHowto.md#loading-more-of-a-variable
+			// from https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#looking-into-variables
+			if (((v.kind === GoReflectKind.Array || v.kind === GoReflectKind.Slice || v.kind === GoReflectKind.Struct) && v.len > v.children.length) ||
+				(v.kind === GoReflectKind.Map && v.len > v.children.length / 2) ||
+				(v.kind === GoReflectKind.Interface && v.children.length > 0 && v.children[0].onlyAddr === true)) {
 				await this.evaluateRequestImpl({ 'expression': exp }).then(result => {
 					const variable = this.delve.isApiV1 ? <DebugVariable>result : (<EvalOut>result).Variable;
 					v.children = variable.children;
 				}, err => logError('Failed to evaluate expression - ' + err.toString()));
 			}
 		};
+		// expressions passed to loadChildren defined per https://github.com/derekparker/delve/blob/master/Documentation/api/ClientHowto.md#loading-more-of-a-variable
 		if (vari.kind === GoReflectKind.Array || vari.kind === GoReflectKind.Slice) {
 			variablesPromise = Promise.all(vari.children.map(async (v, i) => {
-					await loadChildren(`*(*${v.type})(${v.addr})`, v);
+					await loadChildren(`*(*${this.removeRepoFromTypeName(v.type)})(${v.addr})`, v);
 					let { result, variablesReference } = this.convertDebugVariableToProtocolVariable(v);
 					return {
 						name: '[' + i + ']',
@@ -1088,7 +1093,7 @@ class GoDebugSession extends LoggingDebugSession {
 				if (v.fullyQualifiedName === undefined) {
 					v.fullyQualifiedName = vari.fullyQualifiedName + '.' + v.name;
 				}
-				await loadChildren(`*(*${v.type})(${v.addr})`, v);
+				await loadChildren(`*(*${this.removeRepoFromTypeName(v.type)})(${v.addr})`, v);
 				let { result, variablesReference } = this.convertDebugVariableToProtocolVariable(v);
 				return {
 					name: v.name,
@@ -1103,6 +1108,16 @@ class GoDebugSession extends LoggingDebugSession {
 			this.sendResponse(response);
 			log('VariablesResponse', JSON.stringify(variables, null, ' '));
 		});
+	}
+
+	// removes all parts of the repo path from a type name.
+	// e.g. github.com/some/repo/package.TypeName becomes package.TypeName
+	private removeRepoFromTypeName(typeName: string): string {
+		const i = typeName.lastIndexOf('/');
+		if (i === -1) {
+			return typeName;
+		}
+		return typeName.substr(i + 1);
 	}
 
 	private handleReenterDebug(reason: string): void {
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Fixes https://github.com/Microsoft/vscode-go/issues/1010

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
 	protected variablesRequest(response: DebugProtocol.VariablesResponse, args: DebugProtocol.VariablesArguments): void {
 		verbose('VariablesRequest');
 		let vari = this._variableHandles.get(args.variablesReference);
-		let variables;
+		let variablesPromise: Promise<DebugProtocol.Variable[]>;
+		let loadChildren = async (exp: string, v: DebugVariable) => {
```
vladbarosan: nit: this can be a const.

vladbarosan: actually weird that tslint doesnt warn on this

vladbarosan: oh ok, I looked at the tslint.config, it seems some very old config, I guess we can make another PR to update that

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
...
+			if (v.len > 0 && v.children.length === 0) {
+				// from https://github.com/derekparker/delve/blob/master/Documentation/api/ClientHowto.md#loading-more-of-a-variable
+				await this.evaluateRequestImpl({ 'expression': exp }).then(result => {
```
vladbarosan: since we are inside an async function I think we dont need then, i.e. we can do
```ts
await this.X()
const variable = this.delve.isApiV1 ? <DebugVariable>result : (<EvalOut>result).Variable;
v.children = variable.children;
```

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
...
-					variablesReference
-				};
-			});
+			variablesPromise = Promise.all(vari.children.map(async (v, i) => {
+					await loadChildren(`*(*${v.type})(${v.addr})`, v);
+					let { result, variablesReference } = this.convertDebugVariableToProtocolVariable(v);
+					return {
+						name: '[' + i + ']',
+						value: result,
+						evaluateName: vari.fullyQualifiedName + '[' + i + ']',
```
vladbarosan: string interpolation ? ( `${code}texttext`

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
...
+					};
+				})
+			);
 		} else if (vari.kind === GoReflectKind.Map) {
-			variables = [];
-			for (let i = 0; i < vari.children.length; i += 2) {
-				if (i + 1 >= vari.children.length) {
-					break;
+			variablesPromise = Promise.resolve((async () => {
+				let variables = new Array<DebugProtocol.Variable>();
```
vladbarosan: const

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
...
-			variables = vari.children.map((v, i) => {
-				let { result, variablesReference } = this.convertDebugVariableToProtocolVariable(v, i);
-				return {
-					name: '[' + i + ']',
-					value: result,
-					evaluateName: vari.fullyQualifiedName + '[' + i + ']',
-					variablesReference
-				};
-			});
+			variablesPromise = Promise.all(vari.children.map(async (v, i) => {
```
vladbarosan: nit: does it makes sense to extract the code for each of these cases in separate smaller functions ?

src/debugAdapter/goDebug.ts
jhendrixMSFT: Will mark as const.

src/debugAdapter/goDebug.ts
jhendrixMSFT: In your example, what happens if `this.X()` fails, what will the value of `result` be?

src/debugAdapter/goDebug.ts
jhendrixMSFT: This was all existing code, was trying to avoid making the diff too noisy.

src/debugAdapter/goDebug.ts
vladbarosan: depends what you mean by fail. if it throws an error it will throw an error but that will be the same for then is you passed only the onFulfilled function

src/debugAdapter/goDebug.ts
```diff
@@ -1160,6 +1178,17 @@ class GoDebugSession extends DebugSession {
 
 	protected evaluateRequest(response: DebugProtocol.EvaluateResponse, args: DebugProtocol.EvaluateArguments): void {
 		verbose('EvaluateRequest');
+		this.evaluateRequestImpl(args).then(out => {
+			const variable = this.delve.isApiV1 ? <DebugVariable>out : (<EvalOut>out).Variable;
+			response.body = this.convertDebugVariableToProtocolVariable(variable);
+			this.sendResponse(response);
+			verbose('EvaluateResponse');
+		}, err => {
+			this.sendErrorResponse(response, 2009, 'Unable to eval expression: "{e}"', { e: err.toString() });
```
vladbarosan: magic nr ? although i see its used in other places also, there is no constants for them ?

src/debugAdapter/goDebug.ts
```diff
@@ -1171,17 +1200,14 @@ class GoDebugSession extends DebugSession {
...
-				logError('Failed to eval expression: ', JSON.stringify(evalSymbolArgs, null, ' '), '\n\rEval error:', err.toString());
-				return this.sendErrorResponse(response, 2009, 'Unable to eval expression: "{e}"', { e: err.toString() });
-			}
-			const variable = this.delve.isApiV1 ? <DebugVariable>out : (<EvalOut>out).Variable;
-			response.body = this.convertDebugVariableToProtocolVariable(variable, 0);
-			this.sendResponse(response);
-			verbose('EvaluateResponse');
+		};
+		const returnValue = this.delve.callPromise<EvalOut | DebugVariable>(this.delve.isApiV1 ? 'EvalSymbol' : 'Eval', [evalSymbolArgs]).then(val => {
+			return val;
```
vladbarosan: actually I think this one can be  just val => val

src/debugAdapter/goDebug.ts
jhendrixMSFT: We could but since the functions are only relevant to `variablesRequest()` I think it make sense to keep them contained within it (now perhaps refactoring variable requests into its own class would clean this up...).

src/debugAdapter/goDebug.ts
```diff
@@ -995,49 +995,67 @@ class GoDebugSession extends DebugSession {
...
-				let mapValue = this.convertDebugVariableToProtocolVariable(vari.children[i + 1], i + 1);
-				variables.push({
-					name: mapKey.result,
-					value: mapValue.result,
-					evaluateName: vari.fullyQualifiedName + '[' + mapKey.result + ']',
-					variablesReference: mapValue.variablesReference
-				});
-			}
+				return variables;
+			})());
```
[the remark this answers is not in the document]
jhendrixMSFT: That was my original implementation, I switched it though so that it's uniform with the other blocks.  I don't feel super strong about this though so if you think the other way is more clear I can change it.

src/debugAdapter/goDebug.ts
vladbarosan: well if by fail you mean throw an error it will be the same thing(the default onRejection just throws up the error). Also FYI its deprecated as of now to have [unhandled promise rejections and in the future it will cause the Node process to crash](https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections)

so in this case you can either wrap the await in a try catch block and do nothing in the catch or provide a onrejected method that does nothing

jhendrixMSFT: Oh good point will add an `onrejected` method, is `reason => reason` sufficient to do nothing?

src/debugAdapter/goDebug.ts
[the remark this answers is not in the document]
vladbarosan: I mean i think its easier to read if we put on name on it, otherwise ill spend a while reading it to understand what exactly it does

jhendrixMSFT:
@ramya-rao-a FYI

ramya-rao-a:
@jhendrixMSFT Great job! I have 2 main concerns here

- We are now sending parallel requests to delve to evaluate contents of the array. Do we know if delve is able to handle concurrent requests like this?
- These requests to delve will be sent even if there is nothing to expand. For example, take an array of strings. When the user expands that variable, the evaluate request will be sent of each string in the array. Can we somehow not send these requests based on the type of the array or map value?

ramya-rao-a:
> Good point on the spurious evaluations on strings, I will update loadChildren() to never perform the eval if the type is string.

Its not just when the variable itself is a string. This also applies to array of strings, numbers or any of the primitive types. These would not need further evaluations.
Same applies to maps. When the value in the key/value pair of the map is a simple string or number or other primitive type, we dont need further evaluation.

jhendrixMSFT:
I found a great description of when to load children [here](https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#looking-into-variables).  Will update based on this spec.

PR #2207, opened 2018-12-24
Attempt to fix the debugger jumping to gopark

### LINKED ISSUE #2133: Debugging goes into proc.go at every step
I'm using go modules, and every debug step into or step over goes into a proc.go file `gopark` function and then goes back. You can see it in this gif. https://gyazo.com/adb4a54cc467ce372a4b5595b72d150f

Steps to Reproduce:

Have some test and try to debug them. 

launch.json

`{
      "name": "Test",
      "type": "go",
      "mode": "test",
      "request": "launch",
      "remotePath": "",
      "port": 2346,
      "host": "127.0.0.1",
      "program": "${workspaceRoot}/test",
      "env": {
        "GOPATH": "/Users/user/golang"
      },
      "args": ["-test.short"],
      "showLog": false
    }`

segevfiner:
According to the spec you are first supposed to acknowledge the command and only than send the stopped event. Changing this seems to have fixed the jumping to `gopark`/`proc.go` for me. Please test that this works.

Debugging is still really slow, but that's a different issue, VS Code is simply requesting, and receiving, way too much information from delve on each step.

See: [Debug Adapter Protocol Specification - Next Request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Next).

Possible fix for #2133

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -1093,9 +1093,9 @@ class GoDebugSession extends DebugSession {
 			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
 			verbose('continue state', state);
 			this.debugState = state;
+			this.sendResponse(response);
 			this.handleReenterDebug('breakpoint');
 		});
-		this.sendResponse(response);
 		verbose('ContinueResponse');
 	}
 
@@ -1108,9 +1108,9 @@ class GoDebugSession extends DebugSession {
 			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
 			verbose('next state', state);
 			this.debugState = state;
+			this.sendResponse(response);
 			this.handleReenterDebug('step');
 		});
-		this.sendResponse(response);
 		verbose('NextResponse');
 	}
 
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -1093,9 +1093,9 @@ class GoDebugSession extends DebugSession {
 			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
 			verbose('continue state', state);
 			this.debugState = state;
+			this.sendResponse(response);
 			this.handleReenterDebug('breakpoint');
 		});
-		this.sendResponse(response);
 		verbose('ContinueResponse');
```
ramya-rao-a: The `verbose` call for logging should move as well

ramya-rao-a:
@segevfiner I am using 
https://play.golang.org/p/nYkRp-yG1aK (breakpoint in the Println line) which has the jump to proc.go issue. The fix in this PR doesn't fix the jump for me.

Regardless, I believe this is a change we should make anyway. The response should be sent by the adapter only after the call to delve completes.

> According to the spec you are first supposed to acknowledge the command and only than send the stopped event

The link you provided for the spec doesn't talk about any stopped event. Can you give me a link that does refer to it?

segevfiner:
> @segevfiner I am using
https://play.golang.org/p/nYkRp-yG1aK (breakpoint in the Println line) which has the jump to proc.go issue. The fix in this PR doesn't fix the jump for me.

Darn... Guess we'll have to keep digging...

> Regardless, I believe this is a change we should make anyway. The response should be sent by the adapter only after the call to delve completes.

It is there, from the link:

> The debug adapter first sends the response and then a ‘stopped’ event (with reason ‘step’) after the step has completed.

PR #2216, opened 2018-12-27
optimize godef-gomod cache, better support one workspace for multi go…

sunliver:
see this [issue](https://github.com/Microsoft/vscode-go/issues/2222)

**ChangeLog**
1. Now workspaceModCache stands for a workspace's `go.mod`'s path.
2. Rm pkgModCache, use folderModCache instead.
3. Add function `getModPath` to detect file's mod path. If `getModPath` returns none empty string, then `isModSupported` returns true.

**Known Problems**
1. If you are working with GO111MODULE projects and legacy GOPATH projects at the same time, you'd better set GO111MODULE to `auto`. Because we use `go env GOMOD` to find `go.mod`, if may fail when you're under GOPATH project.
2. Godef for gomod is slower than expect now.

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -53,7 +53,7 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 				position,
 				word,
 				includeDocs,
-				isMod: modPath != "",
+				isMod: modPath !== '',
 				modPath,
 			};
 			if (toolForDocs === 'godoc' || (ver && (ver.major < 1 || (ver.major === 1 && ver.minor < 6)))) {
@@ -94,9 +94,9 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
 		token.onCancellationRequested(() => killProcess(p));
 	}
 
-	let cwd: string
+	let cwd: string;
 	if (input.isMod) {
-		cwd = input.modPath
+		cwd = input.modPath;
 	} else {
 		cwd = getWorkspaceFolderPath(input.document.uri);
 	}
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -11,7 +11,7 @@ import path = require('path');
 import { byteOffsetAt, getBinPath, runGodoc, getWorkspaceFolderPath } from './util';
 import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 import { getGoVersion, SemVersion, goKeywords, isPositionInString, getToolsEnvVars, getFileArchive, killProcess } from './util';
-import { isModSupported, promptToUpdateToolForModules } from './goModules';
+import { promptToUpdateToolForModules, getModPath } from './goModules';
 
 const missingToolMsg = 'Missing tool: ';
 
@@ -31,6 +31,7 @@ interface GoDefinitionInput {
 	word: string;
 	includeDocs: boolean;
 	isMod: boolean;
+	modPath: string;
 }
 
 export function definitionLocation(document: vscode.TextDocument, position: vscode.Position, goConfig: vscode.WorkspaceConfiguration, includeDocs: boolean, token: vscode.CancellationToken): Promise<GoDefinitionInformation> {
@@ -46,13 +47,14 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 	}
 	let toolForDocs = goConfig['docsTool'] || 'godoc';
 	return getGoVersion().then((ver: SemVersion) => {
-		return isModSupported(document.uri).then(isMod => {
+		return getModPath(document.uri).then(modPath => {
 			const input: GoDefinitionInput = {
 				document,
 				position,
 				word,
 				includeDocs,
-				isMod
+				isMod: modPath != "",
+				modPath,
 			};
 			if (toolForDocs === 'godoc' || (ver && (ver.major < 1 || (ver.major === 1 && ver.minor < 6)))) {
 				return definitionLocation_godef(input, token);
@@ -91,7 +93,13 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
 	if (token) {
 		token.onCancellationRequested(() => killProcess(p));
 	}
-	const cwd = getWorkspaceFolderPath(input.document.uri);
+
+	let cwd: string
+	if (input.isMod) {
+		cwd = input.modPath
+	} else {
+		cwd = getWorkspaceFolderPath(input.document.uri);
+	}
 
 	return new Promise<GoDefinitionInformation>((resolve, reject) => {
 		// Spawn `godef` process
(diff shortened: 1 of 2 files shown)


src/goDeclaration.ts
```diff
@@ -80,7 +82,7 @@ export function adjustWordPosition(document: vscode.TextDocument, position: vsco
 
 const godefImportDefinitionRegex = /^import \(.* ".*"\)$/;
 function definitionLocation_godef(input: GoDefinitionInput, token: vscode.CancellationToken): Promise<GoDefinitionInformation> {
-	let godefTool = input.isMod ? 'godef-gomod' : 'godef';
+	let godefTool = 'godef';
```
ramya-rao-a: Let's not change this in this PR. I have a separate issue tracking the need to use `godef` instead of `godef-mod`, now that the changes from the latter have been merged to the former

src/goDeclaration.ts
```diff
@@ -91,11 +93,17 @@ function definitionLocation_godef(input: GoDefinitionInput, token: vscode.Cancel
...
+	if (input.isMod) {
+		cwd = input.modPath;
+	} else {
+		cwd = getWorkspaceFolderPath(input.document.uri);
+	}
 
 	return new Promise<GoDefinitionInformation>((resolve, reject) => {
 		// Spawn `godef` process
-		p = cp.execFile(godefPath, ['-t', '-i', '-f', input.document.fileName, '-o', offset.toString()], { env, cwd }, (err, stdout, stderr) => {
+		p = cp.execFile(godefPath, ['-t', '-f', input.document.fileName, '-o', offset.toString()], { env, cwd }, (err, stdout, stderr) => {
```
ramya-rao-a: What's the reason for removing the `-i` flag?
We had that in place to support the Go to definition feature on unsaved changes in the file

src/goModules.ts
```diff
@@ -1,94 +1,84 @@
...
+
+	const hit = folderModCache.get(folderPath);
+	if (hit !== undefined) {
+		return Promise.resolve(hit);
+	}
+
+	workspaceModCache.forEach((v, k) => {
+		if (folderPath.startsWith(k)) {
+			folderModCache.set(folderPath, k);
+			return Promise.resolve(k);
```
ramya-rao-a: This `return` applies just to the callback passed to the `forEach` call. If you want to exit the `getModPath` function when you `folderPath.startsWith(k)` is true, then you need to use a simple `for` loop

sunliver:
@ramya-rao-a  Thanks for you code review! Sorry for some incorrectly changes.

If I recall right, 
- the first run is, for example, my project uses module A, and I run `go to definition` in opened files of A
- the second run is, I just open a file from local mod cache, and I run `go to definition`?

ramya-rao-a:
The first run is when you run `go to definition` on a symbol from module A used in one of the files in your project.
This will open the file from module A from the module cache (I think this would be from GOPATH/pkg/mod)
The second run is when you run `go to definition` on any symbol in this file that got opened from module A

sunliver:
Actually it runs as you think.
I did not change the behavior of old GOPATH package, just add support for mod package.

If you try to `go to definition` under a package, 
- first use `go list -m -f {{.GoMod}}` to figure out whether it is a mod package or not.
- if not, run `godef` under workspace dir
- otherwise run `godef` under mod path


Why change `go env GOMOD` to `go list -m -f {{.GoMod}}`?
https://github.com/golang/go/issues/26500#issuecomment-407510940

ramya-rao-a:
> Why change go env GOMOD to go list -m -f {{.GoMod}}?

Are you asking why we are using `go env GOMOD` instead of `go list -m -f {{.GoMod}}`?
We are doing this as `go list` has been known to be much slower.

sunliver:
But `go env GOMOD` may fail sometimes. 
As I've tested in this issue, https://github.com/golang/go/issues/29433, though it is not a mod package , `go env GOMOD` still get the `go.mod` file. But it is not actually likely to occur.

ramya-rao-a:
@sunliver I have pushed a commit that gets the latest from the master branch. There was a bug where errors from `godef` resulted in an infinite loop which is now fixed with https://github.com/Microsoft/vscode-go/commit/8c6f03572421a49bd67f189ee6fd4152a4b03d47

Now going back to the case I was describing in https://github.com/Microsoft/vscode-go/pull/2216#issuecomment-452791035, lets see the below example

```
package main

import "rsc.io/quote"

func main() {
	quote.Hello()
}
```

Here if you run `Go to definition` on the `Hello()` function, the file `$GOPATH/pkg/mod/rsc.io/quote@v1.5.2/quote.go` opens. This file uses another package called `sampler`. Run `Go to definition` on `sampler.Hello()`.

Using the code in this PR, this `Go to definition` operation fails.
You can see the error in the output channel -> Log (Extension Host)

`Command failed: C:\GoTools\bin\godef.exe -t -i -f c:\GoCode\pkg\mod\rsc.io\quote@v1.5.2\quote.go -o 335
godef: no object`

This is because we are using the folder $GOPATH/pkg/mod as current working directory.

ramya-rao-a:
@sunliver What do you think of #2262?

It is basically the same approach you have taken, plus some simplification, but with not changing the logic for files under the module cache.

ramya-rao-a:
Closing this PR in favor of #2262 which does something similar and has other fixes

sunliver:
> @sunliver I have pushed a commit that gets the latest from the master branch. There was a bug where errors from `godef` resulted in an infinite loop which is now fixed with [8c6f035](https://github.com/Microsoft/vscode-go/commit/8c6f03572421a49bd67f189ee6fd4152a4b03d47)
> 
> Now going back to the case I was describing in [#2216 (comment)](https://github.com/Microsoft/vscode-go/pull/2216#issuecomment-452791035), lets see the below example
> 
> ```
> package main
> 
> import "rsc.io/quote"
> 
> func main() {
> 	quote.Hello()
> }
> ```
> Here if you run `Go to definition` on the `Hello()` function, the file `$GOPATH/pkg/mod/rsc.io/quote@v1.5.2/quote.go` opens. This file uses another package called `sampler`. Run `Go to definition` on `sampler.Hello()`.
> 
> Using the code in this PR, this `Go to definition` operation fails.
> You can see the error in the output channel -> Log (Extension Host)
> 
> `Command failed: C:\GoTools\bin\godef.exe -t -i -f c:\GoCode\pkg\mod\rsc.io\quote@v1.5.2\quote.go -o 335 godef: no object`
> 
> This is because we are using the folder $GOPATH/pkg/mod as current working directory.

@ramya-rao-a 
I've found this case a few days ago. Because `rsc.io/quote` use `sampler` , and `quote` does not add `go.sum` into the repo. And currently `x/package` requires `go.sum` go work, actually `go list` requires `go.sum` to work. So the go tool try to generate the `go.sum` file to work. But `mod` path is read only, and it fails.

Issue #2107, opened 2018-11-08
Wrong documentation when there are multiple functions with the same name


segevfiner:
While Go doesn't support function overloading, you can still have multiple functions with the same name but with a *different* receiver. For example: `testing` has the `Run` function.

It seems that `gocode` and `godef` don't supply us with the receiver. But `go doc` does support receiving it, like so `go doc testing T.Run`.

Steps to Reproduce:

1. Save as `foo_test.go`:
```go
package footest

import "testing"

func TestBla(t *testing.T) {
	t.Ru
}
```
2. Try to complete `t.Ru`.
3. You will get the documentation for `testing.B.Run` instead of `testing.T.Run`.

Run: `gocode -f=json -builtin -in=foo_test.go autocomplete 74 | python -m json.tool` to see what `gocode` gives us, `go doc testing Run` to see what we get from `go doc`, and `go doc testing T.Run` for what we would have liked to get.

vscode-go v0.7.0

ramya-rao-a:
@segevfiner Thanks for reporting!

This is a hard problem to solve...

We cant use `go doc testing T.Run` because there is no way to know that we should use `T.Run` and not `Run`. We dont get that information from `gocode`

`gocode` does give the function signature which I thought I could use to get the right info from `go doc testing Run` which gives docs for all 3 Run() implementations.

But this wouldnt work as `gocode` gives us the signature as `func(name string, f func(t *testing.T)) bool` where as `go doc testing Run` returns `func (t *T) Run(name string, f func(t *T)) bool`

i.e since T is qualified with the package name in the gocode output but not the one from go doc, its hard to get a match.

> You will get the documentation for testing.B.Run instead of testing.T.Run.

Actually the documentation for all the 3 Run() are shown (without the signature)

cc @segevfiner

segevfiner:
As I said, we will need to extend `gocode` and `godef` so that the also supply us with the receiver. Keeping backwards compatibility in mind, so we might have to add a flag so that they only print the receiver when asked.

ramya-rao-a:
Based on the current state, the problem for when using `gocode` was fixed with #2215

This leaves the docs that appear on hover.

@segevfiner Does this problem occur when using the language server as well?
If not, then I am inclined to close this issue in favor of using the language server

cc @stamblerre

stamblerre:
The language server does not have this problem, so I think we can close this issue.

PR #2223, opened 2018-12-31
Use the newly added receiver output from godef to get the correct documentation

segevfiner:
Use the newly added receiver output from godef to get the correct documentation

See rogpeppe/godef#105

Part of #2107

⚠️ Depends on https://github.com/Microsoft/vscode-go/pull/2215 and the godef PR being merged

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -9,7 +9,7 @@ import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
 import { byteOffsetAt, getBinPath, runGodoc, getWorkspaceFolderPath } from './util';
-import { promptForMissingTool } from './goInstallTools';
+import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
 import { getGoVersion, SemVersion, goKeywords, isPositionInString, getToolsEnvVars, getFileArchive, killProcess } from './util';
 import { isModSupported, promptToUpdateToolForModules } from './goModules';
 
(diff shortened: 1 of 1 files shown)


segevfiner:
Receivers are really common, I think it's best to have the user update IMHO.

segevfiner:
About https://github.com/Microsoft/vscode-go/blob/latest/src/goDeclaration.ts#L170... Does that even work?, It's a return inside the `child_process.execFile` callback, which is going to become the return value of the callback rather than the return value of the surrounding function.

ramya-rao-a:
I do recall testing this but now that I look at the code, I am doubtful. It should most likely be

```
if (stderr && stderr.startsWith('flag provided but not defined: -tags')) {
	p = null;
	return definitionLocation_gogetdoc(input, token, false).then(resolve, reject);
}
```

ramya-rao-a:
I take back my objection for keeping the prompt for updating the tool.

At this point, `godef` not only has the new `-r` flag but also has the module support. I was thinking of having some kind of update notification for the latter, but the current prompt will do the trick for both features.

segevfiner:
@ramya-rao-a Oh noes.... The PR for godef wasn't merged yet! Like that it's just going to request people to update godef ad nauseum...

ramya-rao-a:
@segevfiner I trusted you!!!!

Lol, kidding. I should have checked if the upstream PR was merged or not.
Let me reach out to the project maintainers there.
For now, I'll just push a commit to not use the `-r` flag

segevfiner:
I did add a warning about this in the PR description, guess it didn't stand out enough enough. :man_shrugging:

PR #2224, opened 2018-12-31
Revert "Attempt to fix the debugger jumping to gopark (#2207)"

segevfiner:
Whoops, vscode-go was actually already sending the response first. This PR was wrong and should be reverted. It probably caused the gopark issue to happen less because VS Code got stuck on the continue/next/etc taking a long time waiting for the actual break to happen.

This reverts commit 456fdab49f3e211ddb310c1abb4443c5dadd6c0f.

@ramya-rao-a

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -1090,10 +1090,10 @@ class GoDebugSession extends LoggingDebugSession {
 			const state = this.delve.isApiV1 ? <DebuggerState>out : (<CommandOut>out).State;
 			verbose('continue state', state);
 			this.debugState = state;
-			this.sendResponse(response);
-			verbose('ContinueResponse');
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
It definitely was sending the response first, and then only in the call back from delve was `handleReenterDebug` called which would send the stopped event.

I assumed you meant that we had to wait until delve is done to send the response.

segevfiner:
> It definitely was sending the response first, and then only in the call back from delve was `handleReenterDebug` called which would send the stopped event.
> 
> I assumed you meant that we had to wait until delve is done to send the response.

It's the other way around. The original code was correct. You need to acknowledge the continue/next/etc request after passing it to the debugger and it accepts it, this updates the frontend to indicate that the code is now running. Once the debugger stops again you send the breakpoint event.

PR #2245, opened 2019-01-13
Format go doc comments as intended rather than treat them as Markdown

### LINKED ISSUE #1486: Asterisk seems to be treated specially in documentation preview
**Description**

Documentation preview requires an asterisk to be escaped, while `godoc` does not require the escaping. This is a pretty minor issue but I'd expect vscode-go to behave as close to `godoc` as possible.

**Actual behavior**

This comment:

```go
// ValidFileName returns nil if the string is a valid file name, or an
// error otherwise. If pattern is true, file glob patterns are permitted
// ("*" and "?", e.g. "abc*")
```

is rendered like so:

actual

Notice that the * is missing between "".

**Expected behavior**

`godoc -http` renders it like so:

godoc

I'd expect this to be correct, though I haven't read godoc spec.

**Workaround**

You can escape the * in the comment (`\*`), like so:

```go
// ValidFileName returns nil if the string is a valid file name, or an
// error otherwise. If pattern is true, file glob patterns are permitted
// ("\*" and "?", e.g. "abc*")
```

Then it appears to be rendered as expected:

workaround

**Version info**

vscode-go 0.6.73

VSCode version info (OS: Windows 10):

```
Version 1.19.2
Commit 490ef761b76b3f3b3832eff7a588aac891e5fe80
Date 2018-01-10T15:55:03.538Z
Shell 1.7.9
Renderer 58.0.3029.110
Node 7.9.0
Architecture x64
```

segevfiner:
Go documentation comments are not Markdown and use their own minimal syntax. This is an attempt to handle them as such by re-implementing [golang/go//src/go/doc/comment.go](https://github.com/golang/go/blob/master/src/go/doc/comment.go) in TypeScript but outputting Markdown (Escaping as needed) instead of Text or HTML.

Also a drive-by fix to not use the deprecated `MarkedString`.

Needs some testing to make sure this works correctly as it's really easy to make a mistake, especially since I had to work hard around the crippling limitations of JS regex. (I didn't have the time to thoroughly test this yet, might be buggy!)

### Need to test the following
- [x] Just a normal hover/completion.
- [x] Headers
- [x] Pre-formatted blocks &ndash; Works in unit test at least. https://github.com/golang/go/issues/29708.
- [x] Links
- [x] Smart quotes
- [x] Newlines between the different parts &ndash; Hopefully correct... The original code relayed on re-wrapping which we don't need to do as Markdown already undergoes re-wrapping.
- [x] Some Markdown syntax that should be escaped.

Fixes #1486 

Pre-formatted blocks don't show due to https://github.com/golang/go/issues/29708

### DIFF
--- a/src/godocToMarkdown.ts
+++ b/src/godocToMarkdown.ts
@@ -3,24 +3,26 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------*/
 
+// Implementation is based on https://github.com/golang/go/blob/master/src/go/doc/comment.go
+
 import * as XRegExp from 'xregexp';
 
-function markdownEscape(text: string): string {
+// Regex is based on https://github.com/kemitchell/markdown-escape.js/blob/master/index.js
+const markdownEscape = (text: string): string => {
 	return text.replace(/([\\`*{}[\]()#+-.!_>~|\n"$%&',\/:;<=?@^])/g, '\\$1');
-}
+};
 
 // Escape comment text for Markdown. If nice is set,
 // also turn `` into '“' and '' into '”'.
-function commentEscape(text: string, nice: boolean): string {
-	if (nice) {
-		return markdownEscape(convertQuotes(text));
-	}
-	return markdownEscape(text);
-}
+const commentEscape = (text: string, nice: boolean): string => {
+	return nice
+		? markdownEscape(convertQuotes(text))
+		: markdownEscape(text);
+};
 
-function convertQuotes(text: string): string {
+const convertQuotes = (text: string): string => {
 	return text.replace('``', '“').replace('\'\'', '”');
-}
+};
 
 // Regexp for Go identifiers
 const identRx = '[\\p{L}_][\\p{L}_0-9]*';
@@ -50,24 +52,25 @@ const matchRx = '(' + urlRx + ')|(' + identRx + ')';
 // and the word is converted into a link. If nice is set, the remaining text's
 // appearance is improved where it makes sense (e.g., `` is turned into &ldquo;
 // and '' into &rdquo;).
-function emphasize(line: string, words: {string: string}, nice: boolean): string {
+const emphasize = (line: string, words: { [key: string]: string }, nice: boolean): string => {
 	const parts = [];
 
 	const regexp = XRegExp(matchRx, 'gu');
 	while (true) {
 		let m = XRegExp.exec(line, regexp);
-		if (m == null) {
+		if (m === null) {
 			break;
 		}
 		// m >= 3 (two parenthesized sub-regexps in matchRx, 1st one is urlRx)
 
 		// Write text before match
-		parts.push(commentEscape(line.slice(0, m.index)	, nice));
+		parts.push(commentEscape(line.slice(0, m.index), nice));
 
 		// Adjust match for URLs
 		let match = m[0];
 		if (match.includes('://')) {
-			let m0 = m.index, m1 = m[0].length;
+			const m0 = m.index;
+			let m1 = m[0].length;
 			for (const s of [['(', ')'], ['{', '}'], ['[', ']']]) {
 				const open = s[0], close = s[1];
 				// Require opening parentheses before closing parentheses (#22285)
@@ -83,7 +86,7 @@ function emphasize(line: string, words: {string: string}, nice: boolean): string
 					let prevMatch2 = null;
 					while (true) {
 						const match2 = XRegExp.exec(shortenedLine, parenRegex);
-						if (match2 == null) {
+						if (match2 === null) {
 							m1 = prevMatch2.index;
 							break;
 						}
@@ -137,29 +140,29 @@ function emphasize(line: string, words: {string: string}, nice: boolean): string
 	parts.push(commentEscape(line, nice));
 
 	return parts.join('');
-}
+};
 
-function indentLen(s: string): number {
+const indentLen = (s: string): number => {
 	let i = 0;
 	while (i < s.length && (s[i] === ' ' || s[i] === '\t')) {
 		i++;
 	}
 	return i;
-}
+};
 
-function isBlank(s: string): boolean {
+const isBlank = (s: string): boolean => {
 	return s.length === 0 || (s.length === 1 && s[0] === '\n');
-}
+};
 
-function commonPrefix(a: string, b: string): string {
+const commonPrefix = (a: string, b: string): string => {
 	let i = 0;
 	while (i < a.length && i < b.length && a[i] === b[i]) {
 		i++;
 	}
 	return a.slice(0, i);
-}
+};
 
-function unindent(block: string[]) {
+const unindent = (block: string[]) => {
 	if (block.length === 0) {
 		return;
 	}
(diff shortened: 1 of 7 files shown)


segevfiner:
Hopefully the list of characters I escape in Markdown is correct, I took it from https://github.com/wooorm/markdown-escapes. Please review this bit when reviewing this PR.

ramya-rao-a:
@segevfiner Thanks for the PR @segevfiner, this will take a little longer time than usual to look at :)

Can you create a separate PR for the changes you did to goExtraInfo.ts to remove the use of deprecated `MarkedString` so that we can get that in independently?

src/godocToMarkdown.ts
```diff
@@ -0,0 +1,372 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------*/
+
+import * as XRegExp from 'xregexp';
+
+function markdownEscape(text: string): string {
+	return text.replace(/([\\`*{}[\]()#+-.!_>~|\n"$%&',\/:;<=?@^])/g, '\\$1');
```
vladbarosan: @segevfiner where is this regex from ? I couldn't see it in the go impl

segevfiner: It isn't in the Go impl. The Go impl does plain text and HTML. This one does Markdown. In order to get Markdown that renders the same as the HTML we need to escape anything that would be interpreted as markup by Markdown that isn't part of the minimal processing that godoc does.

This is basically a list of characters that can have special meaning in Markdown and need to be backslash escaped.

I based this on this: https://www.npmjs.com/package/markdown-escapes

vladbarosan: @segevfiner thanks for the quick reply. ill put the reference next to it

test/go.test.ts
```diff
@@ -220,7 +220,7 @@ encountered.
 			[new vscode.Position(22, 5), 'main func()', '\n'],
```
vladbarosan: this seems to fail with 
`      AssertionError [ERR_ASSERTION]: '\n```go\nmain func()\n```\n' == '\n```go\nmain func()\n```\n\n'
      + expected - actual
       
       ```go
       main func()
       ```
      +
     `

vladbarosan: this seems to fail with an extra line
`      AssertionError [ERR_ASSERTION]: '\n```go\nmain func()\n```\n' == '\n```go\nmain func()\n```\n\n'
      + expected - actual
       
       ```go
       main func()
       ```
      +
     `

segevfiner:
Strange... Did it fail before you changed things? I remember the tests worked correctly before.

So it is either new tests that need to be updated post rebase, or it is caused by one of the changes. It had a few places where there was some strange code that was crucial for this to work correctly which you might have changed... Like turning `==` into `===` in places that needed to also check for `undefined`, or the weird `if (printed)` stuff. Yeah... getting the newlines to match what godoc does and Markdown needs was quite annoying. :grimacing:

You can check the relevant commit that updates the tests if it is just a test that needs updating.

vladbarosan:
@segevfiner thanks, ill have a look, those that you mention I think I inverted the conditions so should be fine, Ill debug and see, would like to make any crucial behavior explicit for better maintenance.

vladbarosan:
@segevfiner I see in a previous commit you added an extra "\n" to the test. that seems to cause an issue because it expects a \n\n at the end while the output has only one \n ( which seems correct ). I removed it let me know if you think there is any issue

segevfiner:
I think the double `\n` came from adding a newline somewhere where it was required to separate some sections in Markdown so they render correctly (e.g. pre-formatted blocks). But I haven't looked at the changes them self (I don't have the time ATM).

src/godocToMarkdown.ts
```diff
@@ -50,24 +52,25 @@ const matchRx = '(' + urlRx + ')|(' + identRx + ')';
...
 // and '' into &rdquo;).
-function emphasize(line: string, words: {string: string}, nice: boolean): string {
+const emphasize = (line: string, words: { [key: string]: string }, nice: boolean): string => {
 	const parts = [];
 
 	const regexp = XRegExp(matchRx, 'gu');
 	while (true) {
 		let m = XRegExp.exec(line, regexp);
-		if (m == null) {
+		if (m === null) {
```
kpajdzik: You need to check for `undefined` as well here

vladbarosan: not really, it always returns null on no match

src/godocToMarkdown.ts
```diff
@@ -246,28 +249,28 @@ interface Block {
 }
 
 function* blocks(text: string): IterableIterator<Block> {
-	let block: Block;
-	let para = [];
+	let para: any[] | string[] = [];
```
kpajdzik: What's the purpose of `| string[]`?

vladbarosan: automatic type detected by Typescript

vladbarosan:
@jhendrixMSFT , @ramya-rao-a changes look fine and seem to work. Feel free to merge if it looks good.

segevfiner:
> Is there a reason we can't simply call hoverTexts.appendText(definitionInfo.doc) in provideHover()?

I'm not sure I follow... But if you mean treating the doc comments as plain text, than it's true that Go doc comments aren't Markdown, but they aren't pure plain text either. They have their own few formatting rules that are rendered when viewed as HTML, so it is nice to render it into the proper Markdown that will render the same in VS Code.

jhendrixMSFT:
Got it, thanks for the clarification.
So it seems that formatting for Go comment blocks is contentious, e.g. poor support for lists.  Case in point, with this change bullet lists no longer format nicely which IMO is unfortunate.  If we are to take this I think we should make it configurable and off by default so the community can play around with it and provide feedback.

segevfiner:
The godoc site and tool renders documentation as done here (Barring bugs). The Go doc comments are not Markdown and treating them as such can result in random unintended formatting. I'm sure you can find plently of doc comments that render weirdly due to this.

Bullet lists aren't supported in the official format, and so does all other Markdown formatting as it's not Markdown and has little to no support for formatting (Unfortunately...). So if someone uses them, the only place they are going to show is VS Code. Once someone looks at the same comment in the godoc site or tool, it will render plain, so it might mislead developers to think that those comments are Markdown while they are not.

A setting can be added but I do believe we should default to formatting comments as intended, and done here, and not as Markdown. But that's just my 2 cents.

jhendrixMSFT:
There is some support for lists, at least when using a '-' prefix.  For an example see here.
https://golang.org/pkg/net/http/cookiejar/#PublicSuffixList
Is the official format documented someplace?

jhendrixMSFT:
I do agree that we should strive to not treat it as markdown, we just have to tread carefully here as some people might view the change as a regression even if it's technically correct.

segevfiner:
That's actually not a list, but a pre-formatted block that due to golang/go#29708 doesn't render.

There is also this: https://godoc.org/go/doc#ToHTML, but yeah the documentation is rather sparse.

JAicewizard:
Should the string `foo [ http://bar(])` be converted to `foo \[ [http\:\/\/bar\(](http://bar\()\]\)`?
Right now it escapes to `foo \[ [http\:\/\/bar\(](http://bar()\]\)` but I found some markdown viewers that don't render this properly(eg:[dillinger](https://dillinger.io/)).

ramya-rao-a:
Hey @segevfiner 

My apologies for the radio silence here.

We will have to close this PR as we are not accepting any PRs at the moment unless it is related to the move or critical bug fixes. 

Since we are moving towards using `gopls` which provides the hover and code completion features, I am not sure if it is the right step to add more code to the older implementation for these features. But, if you believe there is good value in here, please consider creating this PR in the new home i.e https://github.com/golang/vscode-go.

cc @stamblerre, @hyangah

segevfiner:
@ramya-rao-a I assume it's expected to eventually retire the old tools and migrate purely to gopls. In the gopls world, this kind of transformation to Markdown should be done by gopls so this PR is not needed but is a reference for when implementing this in gopls.

ramya-rao-a:
That is the right assumption, thanks for your efforts here!

JAicewizard:
@segevfiner The gopls implementation is based on the go comment parser with the markdown specific things based on this PR. This PR helped me a lot for the initial implementation.

PR #2263, opened 2019-01-20
Show coverage for files when tests fails

### LINKED ISSUE #2193: Show test coverage when test(s) fail.
I'm manually porting an existing large/complex codebase to Go and am using failed test cases to work out where there is a mismatch in implementation.  In doing so, it helps to look at the code path of a failed test.

As the codebase is large, it helps to use the code coverage feature to limit the manual visual diff when failures occur.

Unfortunately, however, vscode-go does **not show code coverage when the test fails**, despite Go producing the coverprofile file.  A workaround is to force tests to pass by removing test assertions, but this is time-consuming.  This took me an hour to work out: I was especially puzzled as to why no code coverage was shown, despite the GOPATH looking OK and the coverprofile file being generated OK.

Can we include failed tests into the code coverage?

katakonst:
Fixes #2193 by showing coverage in files even if the test fails

### DIFF
--- a/src/goCover.ts
+++ b/src/goCover.ts
@@ -305,7 +305,6 @@ export function toggleCoverageCurrentPackage() {
 		return goTest(testConfig).then(success => {
 			if (!success) {
 				showTestOutput();
-				return [];
 			}
 			return applyCodeCoverageToAllEditors(tmpCoverPath, testConfig.dir);
 		});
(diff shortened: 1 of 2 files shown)


src/goCover.ts
```diff
@@ -305,7 +305,7 @@ export function toggleCoverageCurrentPackage() {
 		return goTest(testConfig).then(success => {
 			if (!success) {
 				showTestOutput();
-				return [];
+				return applyCodeCoverageToAllEditors(tmpCoverPath, testConfig.dir);
```
ramya-rao-a: We can skip this line altogether as we are returning the same outside this `if` block.

PR #2280, opened 2019-01-28
Add gomultifmt as one of go.formatTool

ZhiweiWang:


ramya-rao-a:
Thanks for the PR @ZhiweiWang 

From what I read in the README at https://github.com/Eun/gomultifmt, `gomultifmt` will run multiple formatters one after the other.

We have had reports of existing formatters taking too long a time to finish. Due to this, the `editor.formatOnSaveTimeout` kicks in and aborts the formatting process.

Therefore, I am not very comfortable to add a new option that will run multiple formatters and thus increasing the total time taken.

But, if you want to use this tool, there is another way to do this without doing any code changes to this extension. Use the setting `go.alternateTools` to map whatever is the value for the `go.formatTool` (default is goreturns) to `gomultifmt`. For example:
```
"go.alternateTools": {
    "goreturns": "gomultifmt"
}
```

By doing this, the extension will run `gomultifmt` instead of `goreturns`

Issue #2260, opened 2019-01-18
gotest -run flag needs to override user setting supplied -run flag


DanMaples:
The tool to run tests inserts a `-run TestNameOne|TestNameTwo` when ran by clicking the `run file tests` or the `run test` hover text in a *_test.go file. (see vscode-go/src/testUtils.ts:375)

This conflicts with user settings supplied `-run regex` flags, resulting in the error of:
`go test: run flag may be set only once`

In our project, our tests have suffixes of "_unit", "_acceptance", "_regression", etc... so our user settings file has:
```
"go.testFlags": [
        "-run",
        "^(.*_unit)$"
    ],
```
in it so that when tests are run on a file save, only the unit tests run.

Clicking `run test` results in:
```
Running tool: /usr/local/go/bin/go test github.com/CompanyName/projectName/packageName -run ^(TestNameOne_unit)$ -run ^(.*_unit)$

go test: run flag may be set only once
run "go help test" or "go help testflag" for more information
Error: Tests failed.
```

When clicking on the `run package tests` hover text, no `-run regex` is inserted by the tool so the command looks like:
```
Running tool: /usr/local/go/bin/go test -timeout 120s github.com/CompanyName/projectName/packageName -v -run ^(.*_unit)$
```
and it runs as expected.

It seems to us that there are a few potential solutions here, but we don't want to dictate any solution to you.  It might just be easiest to say "Ignore the user settings `-run regex` commands when the tool is creating its own`

Thank you for your help, please let me know if there is anything else I can provide.
-Dan

ramya-rao-a:
Thanks for reporting this @DanMaples!

I believe we add the `-run` followed by the test functions when running individual tests or when running all the tests in the file. As you have already noticed, we dont do this when running all the tests in the package.

> It might just be easiest to say "Ignore the user settings -run regex commands when the tool is creating its own`

This would be the right approach.

Also, another place where the `go.testFlags` settings are used where we should probably ignore the user provided run regexes is in the goBuild.ts file when we run `go test` to find build errors in test files. 

PRs are most welcome to make this change.
Code Pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- The `getTestFlags()` function in the testUtils.ts file is the one that reads from the settings. All we need to do is to go through the different features that call this and prune out the run regex as needed

[...]

PR #2285, opened 2019-01-29
Remove user supplied -run flags when running tests

scriptonist:
Towards: #2260 

I've tried to address the bug in accordance with the issue comments. Given that I'm very new to the codebase and TS in general, there should be ample scope for improvement. I would really appreciate feedback and suggestions.


ramya-rao-a:
Hey @scriptonist 

We haven't heard back from you in a while, so I have taken the liberty to tweak your PR a bit and take it to conclusion.

I have reverted the commits that were made due to my earlier suggestion to make `getTestFlags` take an arg to decide whether `-run` should be removed or not.

I realized that this forces us to think from each entry point and instead there is a simpler way of doing this by removing the user provided `-run` flag only when the extension explicitly uses it.

Thanks for the work you put in here earlier 

Happy Coding!

PR #2289, opened 2019-01-29
Cache results of 'go list' when fetching package info to get package variables

jhendrixMSFT:
The 'go list' command is expensive so the result should be cached.
There was a cache in place but it wasn't being populated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,7 +500,7 @@
               },
               "showGlobalVariables": {
                 "type": "boolean",
-                "default": false,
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
You will also need to update https://github.com/Microsoft/vscode-go/blob/98c965110942a4d9fbdcde9e9a8af8d9398c4699/src/debugAdapter/goDebug.ts#L560

PR #2301, opened 2019-02-04
Expand nested content only for non array, maps

ramya-rao-a:
Delve doesn't load strings or arrays or slices whose length is greater than 64. This value can be changed by using the setting `dlvLoadConfig` whose default values are as follows

```
	"go.delveConfig": {
		"dlvLoadConfig": {
			"followPointers": true,
			"maxVariableRecurse": 1,
			"maxStringLen": 64,
			"maxArrayValues": 64,
			"maxStructFields": -1
		}
	}
```

In https://github.com/Microsoft/vscode-go/pull/2198, we added the ability to expand nested content in the variables pane to fix https://github.com/Microsoft/vscode-go/issues/1010.

But while providing the feature to expand nested content, we also ended up expanding arrays and strings that have length > 64 if the said arrays/strings are children of one of the variable.

This PR restricts such expansion to just structs and interfaces

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
     * Use unique stack frame ids to be compliant with Debug Adapter Protocol. [PR 2130](https://github.com/Microsoft/vscode-go/pull/2130)
 
 * [Joel Hendrix (@jhendrixMSFT)](https://github.com/jhendrixMSFT)
-    * Display nested content in variables pane. Fixes [Bug 1010](https://github.com/Microsoft/vscode-go/issues/1010) with [PR 2198](https://github.com/Microsoft/vscode-go/pull/2198)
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Not that I'm pushing back but curious what prompted this change?

ramya-rao-a:
I had this point in my head the first time we went through your PR, but I hadnt sorted my thoughts yet and didnt get a chance to actually try the various scenarios.

Yesterday, I did get some time, and so the prompt for the change :)

jhendrixMSFT:
Do you have a test case where you see it loading more than the 64 items?  For me it only loads 64 even if there are more in the array.  Maybe I need to update my test.

ramya-rao-a:
Hmm... now that I check it, I do see that it loads only 64.

Do you have a test case, where we would need to expand the arrays/slices/maps ?
With mine, they expanded just fine even after the changes in this PR

PR #2308, opened 2019-02-07
Override the path to dlv executable

jhendrixMSFT:
This makes it possible to override the dlv executable from the
launch config file.

### DIFF
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -2,7 +2,7 @@
 
 import vscode = require('vscode');
 import path = require('path');
-import { getCurrentGoPath, getToolsEnvVars, sendTelemetryEvent, getBinPath } from './util';
+import { getCurrentGoPath, getToolsEnvVars, sendTelemetryEvent, getBinPath, resolvePath } from './util';
 import { promptForMissingTool } from './goInstallTools';
 
 export class GoDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Microsoft/vscode-go/issues/2329

src/goDebugConfiguration.ts
```diff
@@ -80,7 +80,11 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['dlvLoadConfig'] = dlvConfig['dlvLoadConfig'];
 		}
 
-		debugConfiguration['dlvToolPath'] = getBinPath('dlv');
+		if (!debugConfiguration['dlvToolPath']) {
+			debugConfiguration['dlvToolPath'] = getBinPath('dlv');
```
ramya-rao-a: Can you remind me why using `go.alternateTools` wont work here?
```
"go.alternateTools": {
   "dlv": "your tool path to delve"
}
```

jhendrixMSFT: Copying from our internal conversation.
ok I looked into this, the ability to override dlv via go.alternateTools isn't hooked up...i.e. we don't call `getBinPathWithPreferredGopath()` when resolving the location for dlv....at any rate I don't think this is a great solution as it means that I have modify settings.json and launch.json, plus I can't easily toggle among different configurations as I believe the options in settings.json apply to all configurations (correct me if I'm wrong)

ramya-rao-a: > we don't call `getBinPathWithPreferredGopath()` when resolving the location for dlv.

Do you mean we wont end up calling it if using alternateTools or that we shouldnt call it?
 `getBinPath` internally does call `getBinPathWithPreferredGopath`

> at any rate I don't think this is a great solution as it means that I have modify settings.json and launch.json,

we would be modifying just the settings, why launch.json

> I can't easily toggle among different configurations as I believe the options in settings.json apply to all configurations (correct me if I'm wrong)

You are right about that.

src/goDebugConfiguration.ts
```diff
@@ -80,7 +80,11 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['dlvLoadConfig'] = dlvConfig['dlvLoadConfig'];
 		}
 
-		debugConfiguration['dlvToolPath'] = getBinPath('dlv');
+		if (!debugConfiguration['dlvToolPath']) {
+			debugConfiguration['dlvToolPath'] = getBinPath('dlv');
+		} else {
+			debugConfiguration['dlvToolPath'] = resolvePath(debugConfiguration['dlvToolPath']);
```
ramya-rao-a: Few cases to take care of
- what if user given path is not a valid absolute or relative path?
- where are we expanding user given relative path to absolute path
- check for if resolved path actually exists

Issue #2240, opened 2019-01-10
gocode autocompletion not working when using // in Sprintf


norilt:
If you are using fmt.Sprintf("//%s", #), no autocompletion on postion # is available. However with vim and mdempsky/gocode it is working.

Example Code:
```go
package main

import (
	"fmt"
)

type test struct {
	Test2 string
}

func main() {
	var t test
	fmt.Sprintf("http://%s", ####)
}
```

When you are running gocode in debug mode, you can see, that no request against gocode is made. It looks like vscode is consuming the request.

```
vscode 1.30.2
ms-vscode.Go@0.8.0
go version go1.11.4 linux/amd64
```

ramya-rao-a:
Thanks for reporting this @norilt 

This is happening because there is a logic in place to avoid calling gocode when anything is typed after `//` as it is thought to be a comment.

We need to be a little more smarter there and not consider `//` inside a string as comment

ramya-rao-a:
PRs are welcome to fix this bug

Code Pointers:
- [Build and Debug extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code#building-and-debugging-the-extension)
- This is where we exit early when cursor is after a `//`: https://github.com/Microsoft/vscode-go/blob/0.8.0/src/goSuggest.ts#L126
- There is a helper method `isPositionInString` that we can use to check that the position where the `//` is at, is in a string or not.

rajdhandus:
@ramya-rao-a  - can i please work on this issue?

oneslash:
@rajdhandus feel free to fix it

mossplix:
Is someone working on this?

oneslash:
since there was no PR from @rajdhandus, I have done mine #2316

ramya-rao-a:
The latest [beta version](https://github.com/Microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) (0.9.3-beta.2) of this extension has the fix for this issue

ramya-rao-a:
This bug fix is out in the latest release(0.10.0) of the Go extension

PR #2316, opened 2019-02-09
fix: #2240 - gocode autocompletion not working when using // in Sprintf

oneslash:


oneslash:
@ramya-rao-a, sure, I am little bit overloaded with the new job and the city :) Will do later next week

ramya-rao-a:
That's fine @oneslash, I can take care of the tests

Issue #1026, opened 2017-06-09
Consider dropping support for Go 1.5


ramya-rao-a:
With golint and guru (and maybe more) tools dropping support for Go 1.5, its time to consider if the Go extension would do the same as it depends on these tools for core features.

This issue is to track the findings around the usage of the Go extension with Go 1.5 and if we can support a smaller feature set for Go 1.5 or completely drop support for it

AlekSi:
Go community tends to use 3 versions: tip/master, current and previous one. Currently, that's soon-to-be-1.9, 1.8 and 1.7.

ramya-rao-a:
That's good to know @AlekSi , thanks!

@rakyll @campoy thoughts?

campoy:
I'm totally ok with dropping support for Go 1.5, we're about to release Go 1.9!

ramya-rao-a:
@fatih @zmb3 Do you support Go 1.5 for vim-go and go-plus respectively?

fatih:
@ramya-rao-a  We don't have a specific requirement. Also it shouldn't be a problem if Guru is not compatible with 1.5. What matters is the binary, the user still can use Go 1.5 but can download the binary from another location (if any) or can build `guru` from a different Go version. Unless it makes your life harder, I would suggest to keep it and add a warning if people try to install `guru` via `vscode`, though it's up to you how you want to handle it :)

zmb3:
gogetdoc doesn’t build with 1.5 and we don’t go out of our way to support it. I used to get an occasional issue or two related to 1.5 support but haven’t seen one in a year or two. If it makes your life easier I say drop it. What’s more fun than deleting code?

ramya-rao-a:
@fatih its not just `guru`. `golint` has dropped support for Go 1.5 as well.

@zmb3 Yes, deleting code does bring a ton of joy :)

I'll go with the warning solution. Thanks all

ramya-rao-a:
With #2319 we have not stopped guarding against versions older than 1.9

PR #2319, opened 2019-02-10
fix: #1026 remove go 1.5 support

oneslash:


ramya-rao-a:
Thanks @oneslash 
I have pushed some more cleanup related commits

PR #2332, opened 2019-02-15
Fix debug file not getting deleted

### LINKED ISSUE #2265: Debug file not getting cleared after debugging on Windows
1. Start debug
2. Generate debug file

image

Is this normal?Can I hide it？

vladbarosan:
- execute detach even when halt has failed. Fixes #2265

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   "extensionDependencies": [],
   "dependencies": {
     "diff": "^3.5.0",
+    "fs-extra": "^7.0.1",
     "json-rpc2": "^1.0.2",
     "vscode-debugadapter": "^1.32.1",
     "vscode-debugprotocol": "^1.32.0",
@@ -47,7 +48,6 @@
     "@types/fs-extra": "^5.0.4",
     "@types/mocha": "^5.2.5",
     "@types/node": "^6.14.0",
-    "fs-extra": "^7.0.0",
     "tslint": "^5.11.0",
     "typescript": "^3.1.3",
     "vscode": "^1.1.26"
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
   "extensionDependencies": [],
   "dependencies": {
     "diff": "^3.5.0",
-    "fs-extra": "^7.0.1",
     "json-rpc2": "^1.0.2",
     "vscode-debugadapter": "^1.32.1",
     "vscode-debugprotocol": "^1.32.0",
@@ -47,7 +46,8 @@
   "devDependencies": {
     "@types/fs-extra": "^5.0.4",
     "@types/mocha": "^5.2.5",
-    "@types/node": "^6.14.0",
+    "@types/node": "^8.10.43",
+    "fs-extra": "^7.0.1",
     "tslint": "^5.11.0",
     "typescript": "^3.1.3",
     "vscode": "^1.1.26"
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -532,13 +532,13 @@ class Delve {
 			}
 			clearTimeout(timeoutToken);
 
-			const targetHasExited: boolean = haltErrMsg.endsWith('has exited with status 0');
+			const targetHasExited: boolean = haltErrMsg && haltErrMsg.endsWith('has exited with status 0');
 			const shouldDetach: boolean = !haltErrMsg || targetHasExited;
 			let shouldForceClean: boolean = !shouldDetach;
 			if (shouldDetach) {
 				log('DetachRequest');
 				try {
-					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }]);
+					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: isLocalDebugging }]);
 				} catch (err) {
 					log('DetachResponse');
 					logError(`Failed to detach - ${(err.toString() || '')}`);
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -515,6 +515,7 @@ class Delve {
 			const timeoutToken: NodeJS.Timer = !this.isRemoteDebugging() && setTimeout(() => {
 				log('Killing debug process manually as we could not halt delve in time');
 				killTree(this.debugProcess.pid);
+				this.ensureDebugeeExecutableIsRemoved();
 				resolve();
 			}, 1000);
 
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -526,7 +526,7 @@ class Delve {
 					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }]);
 				} catch (err) {
 					log('DetachResponse');
-					logError(`The timeout will kill the debug process manually as we failed to detach - ${(err.toString() || '')}`);
+					logError(`Failed to detach - ${(err.toString() || '')}`);
 				}
 			}
 			return resolve();
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -499,9 +499,16 @@ class Delve {
 	}
 
 	/**
-	 * To close the debugging session we do the following; If its a local process,
-	 * we try to halt and detach within a timeout otherwise we kill the debugger. If its a remote
-	 * process we try to halt and restart the remote target process running under the debugger.
+	 * Closing a debugging session follows different approaches for local vs remote delve process.
+	 *
+	 * For local process, since the extension starts the delve process, the extension should close it as well.
+	 * To gracefully clean up the assets created by delve, we send the Detach request with kill option set to true.
+	 *
+	 * For remote process, since the extension doesnt start the delve process, we only detach from it without killing it.
+	 *
+	 * The only way to detach from delve when it is running a program is to send a Halt request first.
+	 * Since the Halt request might sometimes take too long to complete, we have a timer in place to forcefully kill
+	 * the debug process and clean up the assets in case of local debugging
 	 */
 	public close(): Thenable<void> {
 		if (this.noDebug) {
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -5,7 +5,8 @@
 
 import * as path from 'path';
 import * as os from 'os';
-import * as fs from 'fs-extra';
+import * as fs from 'fs';
+import * as util from 'util';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { DebugSession, InitializedEvent, TerminatedEvent, ThreadEvent, StoppedEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles, LoggingDebugSession, Logger, logger } from 'vscode-debugadapter';
 import { existsSync, lstatSync } from 'fs';
@@ -13,6 +14,10 @@ import { basename, dirname, extname } from 'path';
 import { spawn, ChildProcess, execSync, spawnSync, execFile } from 'child_process';
 import { Client, RPCConnection } from 'json-rpc2';
 import { parseEnvFile, getBinPathWithPreferredGopath, getInferredGopath, getCurrentGoWorkspaceFromGOPATH, envPath, fixDriveCasingInWindows } from '../goPath';
+
+const access = util.promisify(fs.access);
+const unlink = util.promisify(fs.unlink);
+
 
 // This enum should stay in sync with https://golang.org/pkg/reflect/#Kind
 
@@ -548,8 +553,11 @@ class Delve {
 
 	private async ensureDebugeeExecutableIsRemoved(): Promise<void> {
 		try {
-			if (this.localDebugeePath && await fs.pathExists(this.localDebugeePath)) {
-				await fs.remove(this.localDebugeePath);
+			const fileExists = await access(this.localDebugeePath)
+				.then(() => true)
+				.catch(() => false);
+			if (this.localDebugeePath && fileExists) {
+				await unlink(this.localDebugeePath);
 			}
 		} catch (e) {
 			logError(`Failed to potentially remove leftover debug file ${this.localDebugeePath} - ${e.toString() || ''}`);
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -479,12 +479,11 @@ class Delve {
 		});
 	}
 
-	callPromise<T>(command: string, args: any[]): Thenable<T> {
+	public callPromise<T>(command: string, args: any[]): Thenable<T> {
 		return new Promise<T>((resolve, reject) => {
 			this.connection.then(conn => {
-				conn.call<T>('RPCServer.' + command, args, (err, res) => {
-					if (err) return reject(err);
-					resolve(res);
+				conn.call<T>(`RPCServer.${command}`, args, (err, res) => {
+					return err ? reject(err) : resolve(res);
 				});
 			}, err => {
 				reject(err);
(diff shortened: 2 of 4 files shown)


vladbarosan:
Sorry for getting so late to this one guys, please have a look when you have time

src/debugAdapter/goDebug.ts
```diff
@@ -479,65 +479,63 @@ class Delve {
...
-						})
-						.then(() => resolve());
+				catch (err) {
+					log('RestartResponse');
+					logError(`Failed to restart - ${(err || '').toString()}`);
 				}
-			}, err => {
-				const errMsg = err ? err.toString() : '';
-				log('Failed to halt - ' + errMsg.toString());
-				if (errMsg.endsWith('has exited with status 0')) {
```
vladbarosan: in the previous logic it seems we continued with `restart/detach` if it failed for any other reason than that the debugged processed finished. It seems that we would want exactly the opposite, i.e. continue with detach if the proc already exited.  

Question remains if we want to still continue (as we do today ) if halt fails from another reason than that of the debugged process being finished.

src/debugAdapter/goDebug.ts
```diff
@@ -479,65 +479,63 @@ class Delve {
...
 			}
 
-			this.callPromise('Command', [{ name: 'halt' }]).then(() => {
-				if (timeoutToken) {
-					clearTimeout(timeoutToken);
+			log('HaltResponse');
+			if (!this.debugProcess) {
+				log('RestartRequest');
+				try {
+					await this.callPromise('Restart', this.isApiV1 ? [] : [{ position: '', resetArgs: false, newArgs: [] }]);
```
jhendrixMSFT: @ramya-rao-a a bit unrelated but do you know why we attempt to restart the debuggee during close?  I believe this is related to remote debugging as that's the only time `this.debugProcess` is null.

src/debugAdapter/goDebug.ts
```diff
@@ -479,65 +479,63 @@ class Delve {
...
+			} else {
+				log('DetachRequest');
+				try {
+					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }]);
 				}
-			});
+				catch (err) {
+					log('DetachResponse');
+					logError(`Killing debug process manually as we failed to detach - ${(err.toString() || '')}`);
+					killTree(this.debugProcess.pid);
```
jhendrixMSFT: This seems redundant with the timeout token.  I think you could remove this code and just call `clearTimeout(timeoutToken)` after the above call to detach.

src/debugAdapter/goDebug.ts
jhendrixMSFT: When halt fails can we tell if it was due to the target process exited?

vladbarosan: well kind of, based on the message, like the `if` that was before these changes. So thats pretty much what im asking, do we want to detach only when the target exited, or always when halt fails ( like I mentioned, today we do the exact opposite, we always detach EXCEPT if the target process exited ).

 TBF, not sure why we wouldn't try to detach always if halt fails, I assume there is no point in continuing to be attached, but maybe @ramya-rao-a  or you have more context if there is a reason not to.

src/debugAdapter/goDebug.ts
vladbarosan: so I included the detach request within the timeout so removing the extra killTree here

src/debugAdapter/goDebug.ts
ramya-rao-a: Apologies for the late response.

To summarize, what we are trying to debate here is that when halt fails, if we should restart/detach only if the failure was due to target program exit or for all failures, correct?

If the target program is still running, and halt fails, there is actually no point in restart/detach as these requests would fail as well. If the target program is still running, delve needs to be halted in order to accept any further requests

src/debugAdapter/goDebug.ts
ramya-rao-a: Yes, in case of remote debugging, we don't start the delve process. It was started by the user beforehand. Therefore, we shouldn't be killing the delve process.

I wonder if instead of restart, we should just detach without the kill switch.

@vladbarosan Can you give that a try?

src/debugAdapter/goDebug.ts
```diff
@@ -479,69 +479,68 @@ class Delve {
...
-					return this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }])
-						.then(null, err => {
-							log('DetachResponse');
-							logError(`Killing debug process manually as we failed to detach - ${(err || '').toString()}`);
-							killTree(this.debugProcess.pid);
-						})
-						.then(() => resolve());
+				catch (err) {
+					log('DetachResponse');
+					logError(`The timeout will kill the debug process manually as we failed to detach - ${(err.toString() || '')}`);
```
ramya-rao-a: Why not kill the debug process right away instead of waiting for the timeout to do so.

src/debugAdapter/goDebug.ts
vladbarosan: @ramya-rao-a  correct thats what we are saying. and to make sure we are on the same line, you are suggesting that : 
 - if halt succes -> then detach
 - if halt fails because target has exited -> then detach 
  - if halt fails because of whatever other reason -> doesn't really matter if we detach since it will fail but preferably not so we dont do an extra unnecessary operation

src/debugAdapter/goDebug.ts
vladbarosan: i think that might make sense, I think Delve in API mode can accept multiple clients, in that case I dont think we would want to restart it.

vladbarosan: also, given those, does it even make sense to halt a remote program ?

src/debugAdapter/goDebug.ts
ramya-rao-a: Yes

src/debugAdapter/goDebug.ts
ramya-rao-a: we definitely need to detach. And since we can't detach without halting, we need to halt.

Are you saying that we dont need to detach either?

vladbarosan: well i think we need to detach, maybe im not familiar enough with the remote debugging and I need to try it . I thought delve is started outside our responsability already with a target program ? then do we really need to halt it from our side in that case ?

ramya-rao-a: > I thought delve is started outside our responsability already with a target program

Yes, that's right. Delve is started by the user in the remote machine.

> then do we really need to halt it from our side in that case

I believe we need to detach, else the delve process will continue to think that there is a client attached to it.

vladbarosan: yes agree to `detach`, was saying we dont need to `halt` if the above mentioned are correct. So do we agree on the path forward here ?
for remote, don't even halt, just detach ? if so I can have a look see if everything goes well in that in this direction.

ramya-rao-a: > for remote, don't even halt, just detach

The problem is that we can't detach without halting a delve process (regardless of its remote or not)

vladbarosan: oh really ? that's a bummer, i would expect delve to allow you to detach without halting. ill have a look and see what we can do for remote.

vladbarosan:
@ramya-rao-a  so there are a couple of scenarios for which I think there are tracking issues:
1. User starts delve with an executable that runs and finishes or some code that runs only once
 a.  in this case normally we can just halt and detach which would close the target and delve remotely ( so users would need to restart delve with the program )
 b. if delve is started with `--accept-multiclient`, we could actually restart without halting or detaching so then on new attaches the code is rerun.
 
2. Program starts and executes and we just "attach" and "detach". 

Right now we do for 1 the a option regardless of "accept-multiclient" and we don't support 2. but 1 would be easy to add if we want and 2 is doable also.

@ramya-rao-a @jhendrixMSFT  I know for 2 there is an open issue, not sure if there is an ask for 1.b ?

Thoughts ? Otherwise let me know if this is good to go.

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +479,69 @@ class Delve {
...
+			}
+			return resolve();
 		});
 	}
+
+	private isRemoteDebugging() {
+		return !this.debugProcess;
+	}
+	private targetHasExited(errMsg: string) {
+		return errMsg.endsWith('has exited with status 0');
```
jhendrixMSFT: This assumes the process has a zero exit code, I would either omit the zero or change it to a regex (if you do be sure to handle negative error codes as well).

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +479,69 @@ class Delve {
...
+			}
+			clearTimeout(timeoutToken);
+
+			if (this.shouldDetach(errMsg)) {
+				log('DetachRequest');
+				try {
+					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }]);
+				} catch (err) {
+					log('DetachResponse');
+					logError(`The timeout will kill the debug process manually as we failed to detach - ${(err.toString() || '')}`);
```
jhendrixMSFT: Is this true?  Looks to me that the timeout token has already been cleared.

src/debugAdapter/goDebug.ts
vladbarosan: ehh to be fair i dont like at all checking the exited code. let me look a little if there is a better way to know this. I presume if it requires an extra call to the debug server is not the end of the world ?

src/debugAdapter/goDebug.ts
vladbarosan: yea, we clean it after halt, seems in some thread above we want that for halt failing

src/debugAdapter/goDebug.ts
ramya-rao-a: For the scope of this PR, lets keep this check as is. Its the same what we do today

src/debugAdapter/goDebug.ts
ramya-rao-a: @jhendrixMSFT is right. The timeout is cleared right after halt

vladbarosan:
@ramya-rao-a @jhendrixMSFT  as a final thing to close this I added an extra delete if for some reason the file is still around. For htis I moved the `fs-extra` as a dependency ( instead of a dev dependency). Any issue with that? and in general, ill take the occasion to ask what is our policies on adding dependencies ( I saw they are fairly minimal right now ).

ramya-rao-a:
@vladbarosan Why not use the built in `fs` module instead?

vladbarosan:
@ramya-rao-a , I was thinking of force removing directories but in this case is always a file so I can. Still the question regarding dependencies stands.

test/go.test.ts
```diff
@@ -544,22 +544,18 @@ It returns the number of bytes written and any write error encountered.
...
-
-				const packageNames = groupedSymbolNames[vscode.SymbolKind.Package];
-				const variableNames = groupedSymbolNames[vscode.SymbolKind.Variable];
-				const functionNames = groupedSymbolNames[vscode.SymbolKind.Function];
-				const structs = groupedSymbolNames[vscode.SymbolKind.Struct];
-				assert.equal(packageNames[0], 'main');
-				assert.equal(variableNames, undefined);
-				assert.equal(functionNames[0], 'print');
-				assert.equal(functionNames[1], 'main');
+			new GoDocumentSymbolProvider().provideDocumentSymbols(document, null).then(outlines => {
```
jhendrixMSFT: Does this fix the test that sometimes fails in CI?

ramya-rao-a:
> what is our policies on adding dependencies

We dont have any official policy on this yet. The basic idea is that we won't add dependencies that we can avoid, as it will increase the size of the extension payload.

Did you have any other specific concerns?

package.json
```diff
@@ -46,8 +46,8 @@
   "devDependencies": {
     "@types/fs-extra": "^5.0.4",
     "@types/mocha": "^5.2.5",
-    "@types/node": "^6.14.0",
-    "fs-extra": "^7.0.0",
+    "@types/node": "^8.10.43",
+    "fs-extra": "^7.0.1",
```
ramya-rao-a: We can remove this now that we are using the builtin `fs`

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +487,92 @@ class Delve {
...
+				try {
+					await this.callPromise('Detach', [this.isApiV1 ? true : { Kill: true }]);
+				} catch (err) {
+					log('DetachResponse');
+					logError(`Failed to detach - ${(err.toString() || '')}`);
 				}
-			});
+			}
+			if (!this.isRemoteDebugging()) {
+				await this.ensureDebugeeExecutableIsRemoved();
```
ramya-rao-a: Shouldnt we do this in case of timeout as well?

package.json
vladbarosan: we are using this for the tests , so I think we still need it ?

src/debugAdapter/goDebug.ts
vladbarosan: yes, good catch

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +487,93 @@ class Delve {
...
```
jhendrixMSFT: If detach fails I wonder if we can actually delete the debugee.  Presumably the debugger has an open handle to the debugee so it should fail.  Seems like we shouldn't clear the timer until we know that halt and detach have succeeded.

package.json
ramya-rao-a: Ah!, got it.

src/debugAdapter/goDebug.ts
ramya-rao-a: The timer should be cleared right after Halt, because it is to guard against the case where Halt request doesnt return in time.

But you bring up a good point. 

In case for local debugging, if detach fails, we should run the `killTree` to kill the process before deleting the file

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +487,93 @@ class Delve {
...
-			let timeoutToken: NodeJS.Timer;
-			if (this.debugProcess) {
-				timeoutToken = setTimeout(() => {
-					log('Killing debug process manually as we could not halt and detach delve in time');
-					killTree(this.debugProcess.pid);
-					resolve();
-				}, 1000);
-			}
+		return new Promise(async resolve => {
+			const timeoutToken: NodeJS.Timer = !this.isRemoteDebugging() && setTimeout(() => {
```
ramya-rao-a: Instead of `isRemoteDebugging` being a function, we can have it as a `const` that is declared probably right before the new Promise is created here.

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +487,93 @@ class Delve {
...
-					if (timeoutToken) {
-						clearTimeout(timeoutToken);
-					}
-					return resolve();
+				errMsg = err ? err.toString() : '';
+				log(`Failed to halt - ${errMsg}`);
+			}
+			clearTimeout(timeoutToken);
+
+			if (this.shouldDetach(errMsg)) {
```
ramya-rao-a: `shouldDetach()` and `targetHasExited()` are not being used anywhere else. Can we have them as `const` before this `if` instead of functions?

src/debugAdapter/goDebug.ts
jhendrixMSFT: For local debugging I think we can simplify the entire flow.  Note that this assumes that the end result is that dlv and the debugee have been terminated and that the debug file has been deleted.
1. start a timer that will kill the process tree and delete the debugee; this MUST be no-fail
2. call halt and once that returns call detach; if either of these fail do nothing as the timer will clean up
3. do NOT clear the timer, if #2 succeeded it's a no-op
Is there any reason why this approach won't work?

vladbarosan: @ramya-rao-a  thoughts ?

ramya-rao-a: About step 2 & 3:

Halt is known to not return immediately based on the program being debugged. That was the reason I added the timer. When halt succeeds (or target exited) and we run `detach` i.e happy case, I want to allow detach the time it needs to successfully detach. We dont want the timer to timeout and do forceful kill/remove when detach is running.

If halt fails (but not because target exited) or if detach fails, then we want the forceful kill/remove.

If detach succeeds but doesnt remove the debuggee file, I don't want us to forcefully remove the debuggee file. If we see this, we should log a bug in delve.

vladbarosan: updated the logic, i think it should follow the conditions specified:
1. cleartimeout after halt
2. if localDebug && (halt fails because of anything else than target has exited || detach fails) => force kill  + force remove

jhendrixMSFT: > We dont want the timer to timeout and do forceful kill/remove when detach is running.
Why?  The end result is the same either way and while I agree that we should do best effort to kill things gracefully I'm not convinced that the extra complexity to cover every corner case is worth it.

ramya-rao-a: @jhendrixMSFT Because I'd much rather have delve to the clean up (in the detach request), than kill it when the detach request is running.
If for what ever reason, the detach request doesnt complete in a reasonable time, then that would be a bug with delve and we will have a chance to report it.

src/debugAdapter/goDebug.ts
```diff
@@ -479,71 +487,78 @@ class Delve {
...
 				reject(err);
 			});
 		});
 	}
 
-	close(): Thenable<void> {
+	/**
+	 * To close the debugging session we do the following; If its a local process,
+	 * we try to halt and detach within a timeout otherwise we kill the debugger. If its a remote
+	 * process we try to halt and restart the remote target process running under the debugger.
```
ramya-rao-a: @vladbarosan For remote debugging, we don't seem to restart the remote delve process anymore. I dont' recall the discussion that lead to this decision.

I remember saying that we can consider using detach without the kill option instead of restart, but the current code calls detach with the kill option.

vladbarosan: @ramya-rao-a  yes I remember, let me test the behavior , and ill update

ramya-rao-a: @vladbarosan https://github.com/Microsoft/vscode-go/issues/1609#issuecomment-410111323 is another reference to the need for detach (without kill) instead of restart in case of remote debugging

vladbarosan:
@jhendrixMSFT  @ramya-rao-a  sorry i got back late to this. I think I addressed the remaining comments.

ramya-rao-a:
@vladbarosan I pushed 2 commits with tiny changes, please take a look and let me know if you have any concerns

ramya-rao-a:
Also, did you get a chance to try the remote debugging scenario?

vladbarosan:
@ramya-rao-a  but again reattaching in this scenario wont work unless delve is started with multiclient. @jhendrixMSFT  is working on the attach feature. I think it might even make sense to remove the remote debugging part from launch and just leave it for attach ( since we aren't really launching)

ramya-rao-a:
@vladbarosan For future PRs, please use the Squash option while merging.

vladbarosan:
ah yea, @ramya-rao-a  can we just enable only squash merge on the repo ?

ramya-rao-a:
Good point. Done.

PR #2338, opened 2019-02-16
fix: #2309 use goimports instead of goreturns when using modules

### LINKED ISSUE #2309: goreturns not unstable in module mode
For about a year the auto-formatting using goreturns or goimports has continuously alternated between working and not working (with more time spent in the not working state). gofmt, of course, always works fine.

From one of several (closed) issues about this same problem:
* Does `goreturns` work when you manually format the file in VS Code i.e right-click -> Format Document?
- no (but I have experienced this problem with this plugin in the past)
* Are there any errors in the console? Help -> Toggle Developer Tools -> Console
- no
* Can you try updating the tool? Run `Go: Install/Update Tools`, select `goreturns` to update
- yes, I did

Again, I expect this to suddenly start working again randomly at some point. But then it will stop working again randomly at some point. 

I do use go modules and for some reason have the vague impression that it's related.

jamesgeorge007:
Closes #2309
* Updated `goFormat.ts` to use `goimports` if user's choice is `goreturns` in module mode

### DIFF
--- a/src/goFormat.ts
+++ b/src/goFormat.ts
@@ -21,10 +21,9 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
 
 		let filename = document.fileName;
 		let goConfig = vscode.workspace.getConfiguration('go', document.uri);
-		let formatTool = goConfig['formatTool'] || 'goreturns';
 		let formatFlags = goConfig['formatFlags'].slice() || [];
 
-		this.getFormatTool(goConfig, document.uri)
+		return this.getFormatTool(goConfig, document.uri)
 		.then((formatTool) => {
 			// We ignore the -w flag that updates file on disk because that would break undo feature
 			if (formatFlags.indexOf('-w') > -1) {
--- a/src/goFormat.ts
+++ b/src/goFormat.ts
@@ -8,8 +8,8 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
-import { isModSupported } from './goModules';
 import { promptForMissingTool, promptForUpdatingTool } from './goInstallTools';
+import { isModSupported } from './goModules';
 import { sendTelemetryEvent, getBinPath, getToolsEnvVars, killTree } from './util';
 
 export class GoDocumentFormattingEditProvider implements vscode.DocumentFormattingEditProvider {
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
@ramya-rao-a What do you think?

jamesgeorge007:
@ramya-rao-a Is this fine :thinking:

src/goFormat.ts
```diff
@@ -23,6 +24,13 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
 		let formatTool = goConfig['formatTool'] || 'goreturns';
 		let formatFlags = goConfig['formatFlags'].slice() || [];
 
+		isModSupported(document.uri).then(isMod => {
+			formatTool = 'goimports';
```
lloiser: `isModSupported` returns a promise. Promises continue with the `then` function asynchronously. Which means that the `formatTool` variable will be updated _after_ the `runFormatter` (see a few lines below) function has been called. So it probably does not what you wanted it to do :)

You have two options now: 
* move all the code after the `catch` (so starting with the `// We ignore ...` up to the `return this.runFormatter(...`) into the `then` function here
* or use [async/await](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await) to wait for the promise to be done and continues with the rest of the code

jamesgeorge007:
@lloiser I'm not sure if this is the right implementation. Kindly let me know about the required changes to be made further :+1:

src/goFormat.ts
```diff
@@ -23,30 +24,37 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
...
-		// Fix for https://github.com/Microsoft/vscode-go/issues/613 and https://github.com/Microsoft/vscode-go/issues/630
-		if (formatTool === 'goimports' || formatTool === 'goreturns') {
-			formatFlags.push('-srcdir', filename);
-		}
-
-		// Since goformat supports the style flag, set tabsize if user has not passed any flags
-		if (formatTool === 'goformat' && formatFlags.length === 0 && options.insertSpaces) {
-			formatFlags.push('-style=indent=' + options.tabSize);
-		}
+		isModSupported(document.uri).then(isMod => {
```
lloiser: you have to return here so that the `runFormatter` inside is properly returned

src/goFormat.ts
```diff
@@ -23,30 +24,37 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
...
+			formatTool = 'goimports';
```
lloiser: both the issue and the description of your pull request say:

> use `goimports` if user's choice is `goreturns` in module mode

But this will effectivly overwrite formatTool also if the value is `gofmt`.

It's also unnecessary to call `isModSupported` if the `formatTool` is already `goimports`.

jamesgeorge007: @lloiser I'm not sure how do you intend to invoke `isModSupported` as per the above mentioned case.
Do you mean something like checking for whether `formatTool` is `goimports` and invoking the method appropriately :thinking: 
```js
if (formatTool === 'goimports') {
 //
}
```

I'm not familiar with the codebase as mentioned before. Kindly suggest the required changes :smile:

ramya-rao-a: If the chosen tool is either gofmt or goimports, then no change is needed.
Else, you need to invoke `isModSupported` to figure out if modules are being used or not.
Something like the below

```typescript
private getFormatTool(goConfig: vscode.WorkspaceConfiguration, fileUri: vscode.Uri): Thenable<string> {
	let formatTool = goConfig["formatTool"] || "goreturns";
	if (formatTool === "gofmt" || formatTool === "goimports") {
		return Promise.resolve(formatTool);
	}
	return isModSupported(fileUri).then(isMod => {
		return isMod ? "goimports": formatTool;
	})
} 
```

On second thoughts, I wonder if we should choose to use `gofmt` instead. `gofmt` is installed by default for all, whereas one has to specifically install `goimports`. Also, older versions of `goimports` don't support modules, so `goimports` has to be installed, meaning we need to prompt the user to do so.

Regardless, we are choosing to use a different tool than what the user meant to use. Therefore, we should notify the user of this change, and update workspace settings so that we don't have to make this check every time the file needs to be formatted.

For reference, this updating of workspace settings when using modules is already being done for the `go.infergopath` setting  at https://github.com/Microsoft/vscode-go/blob/0.9.2/src/goModules.ts#L55. We can update the same

jamesgeorge007: @ramya-rao-a Is this the way of implementation that you intended :thinking:

src/goFormat.ts
```diff
@@ -23,30 +24,43 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
 		let formatTool = goConfig['formatTool'] || 'goreturns';
```
lloiser: remove this because `getFormatTool` does this.

src/goFormat.ts
```diff
@@ -23,30 +24,43 @@ export class GoDocumentFormattingEditProvider implements vscode.DocumentFormatti
...
-		// Fix for https://github.com/Microsoft/vscode-go/issues/613 and https://github.com/Microsoft/vscode-go/issues/630
-		if (formatTool === 'goimports' || formatTool === 'goreturns') {
-			formatFlags.push('-srcdir', filename);
-		}
-
-		// Since goformat supports the style flag, set tabsize if user has not passed any flags
-		if (formatTool === 'goformat' && formatFlags.length === 0 && options.insertSpaces) {
-			formatFlags.push('-style=indent=' + options.tabSize);
-		}
+		this.getFormatTool(goConfig, document.uri)
```
lloiser: `return this.getFormatTool...`

src/goFormat.ts
jamesgeorge007: :+1:

src/goFormat.ts
jamesgeorge007: :+1:

jamesgeorge007:
@lloiser Hope it's fine now :thinking:

ramya-rao-a:
@jamesgeorge007, @lloiser  I figured a simpler way to use `goimports` when using modules that doesnt involve checking `isModSupported` for each formatting request. See https://github.com/Microsoft/vscode-go/commit/81013f5781e9d8327fbba6e297c46db9ff754d73

Therefore, am closing this PR. Thanks for all the effort here.

Issue #1886, opened 2018-08-28
Feature Request: Syntax highlight for go.mod and go.sum


yookoala:
Please provide the syntax highlight for the go module files.

ramya-rao-a:
I wonder if https://github.com/atom/language-go is the right place for this feature request. For go itself, we use the grammar from there.

@joefitzgerald, @zmb3 thoughts?

brainsnail:
I'll take a look at this one on the Atom repository. Should be fun 👍

ramya-rao-a:
The latest [beta version](https://github.com/Microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) (0.9.3-beta.2) of this extension, the `go.mod` and `go.sum` files will now have syntax highlighting.

Thanks to @calebdoxsey and @brainsnail!

ramya-rao-a:
This feature is out in the latest release(0.10.0) of the Go extension

PR #2344, opened 2019-02-22
add go.mod and go.sum support

calebdoxsey:
For: https://github.com/Microsoft/vscode-go/issues/1886

I was 90% done with this before I realized there was already a PR out there from a few months ago. If this isn't useful, feel free to close.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
           "go.sum"
         ],
         "aliases": [
-          "Go Sum File"
+          "Go Checksum File"
         ]
       }
     ],
--- a/package.json
+++ b/package.json
@@ -72,6 +72,36 @@
         "aliases": [
           "Go"
         ]
+      },
+      {
+        "id": "go.mod",
+        "extensions": [
+          "go.mod"
+        ],
+        "aliases": [
+          "Go Module File"
+        ]
+      },
+      {
+        "id": "go.sum",
+        "extensions": [
+          "go.sum"
+        ],
+        "aliases": [
+          "Go Sum File"
+        ]
+      }
+    ],
+    "grammars": [
+      {
+        "language": "go.mod",
+        "scopeName": "go.mod",
+        "path": "./syntaxes/go.mod.tmGrammar.json"
+      },
+      {
+        "language": "go.sum",
+        "scopeName": "go.sum",
+        "path": "./syntaxes/go.sum.tmGrammar.json"
       }
     ],
     "snippets": [
--- a/syntaxes/go.sum.tmGrammar.json
+++ b/syntaxes/go.sum.tmGrammar.json
@@ -1,0 +1,44 @@
+{
+	"scopeName": "go.sum",
+	"patterns": [
+		{
+			"include": "#checksum"
+		},
+		{
+			"include": "#semver"
+		},
+		{
+			"include": "#unquoted_string"
+		}
+	],
+	"repository": {
+		"checksum": {
+			"comment": "Checksum",
+			"match": "h1:([^\\s]+)=",
+			"captures": {
+				"1": {
+					"patterns": [
+						{
+							"match": "[a-zA-Z\\d+\\/]{43}",
+							"name": "string.unquoted.go.sum"
+						},
+						{
+							"match": ".*",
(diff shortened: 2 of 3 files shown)


ramya-rao-a:
Thanks for the PR @calebdoxsey. Looks like this is your first contribution to this project, Thanks & Welcome!

> I realized there was already a PR out there from a few months ago

Which PR are you referring to here?

calebdoxsey:
Here's the PR in the atom repo: https://github.com/atom/language-go/pull/156

ramya-rao-a:
@brainsnail Do you know if there is an ETA for your PR (https://github.com/atom/language-go/pull/156)?

If not, then would you be interested in reviewing this PR and give us your feedback?

brainsnail:
@ramya-rao-a As far as I know there hasn't been an update on that PR, but I'm fine with taking a look at this PR instead 😄

I'll take a look at it this soon @calebdoxsey.

syntaxes/go.sum.tmGrammar.json
```diff
@@ -0,0 +1,44 @@
...
+							"match": ".*",
+							"name": "invalid.illegal.unknown-hash.go.sum"
+						}
+					]
+				}
+			}
+		},
+		"semver": {
+			"comment": "Semver version strings (v1.2.3)",
+			"match": "v(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)(?:-[\\da-z-]+(?:\\.[\\da-z-]+)*)?(?:\\+[\\da-z-]+(?:\\.[\\da-z-]+)*)?",
```
OneOfOne: the "[official](https://github.com/golang/go/blob/df557fe3ea94ff8888abd6a6ab4cc5d5351d77a6/src/cmd/go/internal/modfetch/pseudo.go#L89)" regexp is: `^v[0-9]+\.(0\.0-|\d+\.\d+-([^+]*\.)?0\.)\d{14}-[A-Za-z0-9]+(\+incompatible)?$`

calebdoxsey: I think that's for the special pseudo version: https://tip.golang.org/cmd/go/#hdr-Pseudo_versions

brainsnail: I think I agree about using that regex instead - I wrote one by hand on the Atom repo, but we may want to just use the same regex as the one the language uses.

brainsnail:
I tried it out with a few different repos that had modules enabled and did some `go mod init`-ing on a few others to see how it handled the highlighting. As far as that goes - I think it works great. It highlighted the proper keywords on the `go.mod` files, gave appropriate highlighting on versions and kept comment formatting. The `go.sum` file highlights the most important piece - the version.

Super glad to see it moving forward 👍

package.json
```diff
@@ -72,6 +72,36 @@
...
+          "Go Module File"
+        ]
+      },
+      {
+        "id": "go.sum",
+        "extensions": [
+          "go.sum"
+        ],
+        "aliases": [
+          "Go Sum File"
```
brainsnail: Can we call this `Go Checksum File`? It's a minor nitpick.

syntaxes/go.sum.tmGrammar.json
calebdoxsey: The issue is it only matches pseudo versions: 

```v0.0.0-20161113214103-89cd22812c4f```

It won't match 

```v1.2.3```

At least that's how I read the regex. Is there another one that matches both?

package.json
calebdoxsey: Yep. Good idea.

syntaxes/go.sum.tmGrammar.json
brainsnail: I think you are right - the code they use to validate the semver is [here](https://github.com/golang/go/blob/aab0b704d88b7dad10c767ad941619d59887203c/src/cmd/go/internal/semver/semver.go#L150). 

In testing out your branch locally I didn't experience any problems.

ramya-rao-a:
Thanks a lot @brainsnail and @calebdoxsey!

Issue #2323, opened 2019-02-11
Allow changing showGlobalVariables in settings


aschade92:
If I'm using the extension version `0.8.1-beta.7` debugging is very fast and only `local` appears in the variable window:
![image](https://user-images.githubusercontent.com/39062967/52595388-c411b400-2e1b-11e9-9737-c312559a93c6.png)
config: 
```
{
    "go.delveConfig": {
        "apiVersion": 1,
        "dlvLoadConfig": {
            "followPointers": true,
            "maxVariableRecurse": 1,
            "maxStringLen": 600,
            "maxArrayValues": 64,
            "maxStructFields": -1,
            "showGlobalVariables": false
        },
    }
}
```

Now if I install `0.8.1-beta.12` or `0.9.1`, debugging is slightly to significantly slower depending on the amount of variables that are getting collected by the debugger. Setting `showGlobalVariables` to false (as mentioned here: https://github.com/Microsoft/vscode-go/issues/2133#issuecomment-461569056) doesn't really seem to do anything for me (there might be a slight improvement but there is noticeable lag here that is not present in the beta version `0.8.1-beta.7` I mentioned above). Regardless of whether or not `showGlobalVariables` is set to `false`, `global` still appears in the frame: 

![image](https://user-images.githubusercontent.com/39062967/52595811-c45e7f00-2e1c-11e9-8a20-591425796a89.png)

I'm guessing that there is still some other processing being done here in the latest versions that wasn't being done in versions where the debugger works much more fluidly. Is there a way to just completely disable global variables? Is this even expected when `showGlobalVariables` is set to `true`? Global variables aren't really recommended in general by the go community anyways so an option to disable them completely in favor of higher performance would be appreciated.

Steps to Reproduce:

1. Run debug session with above config
2. Debugging is slow, global variables still appear in the variables pane.

ramya-rao-a:
@aschade92 The `showGlobalVariables` needs to be set in the debug configuration in your launch.json file and not in the `go.delveConfig` setting. Can you try doing that?

> Global variables aren't really recommended in general by the go community anyways

Can you point us to some docs or discussion around this?

aschade92:
Cool yeah that worked for me. Is there any way to apply this settings from the workspace settings so that it can be used for running tests? (ie pressing the test or debug button above a test?)

> Can you point us to some docs or discussion around this?

Here are a couple that discuss this issue but I don't think that's super important for the purposes of the thread anymore:

[...]

PR #2351, opened 2019-02-25
Added showGlobalVariables as a workspace level parameter

aschade92:
Before, this option was only available in `launch.json` but this PR makes `showGlobalVariables` as a workspace level setting so that things like the debug and run test actions can use this setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1225,11 +1225,6 @@
                   "type": "number",
                   "description": "MaxStructFields is the maximum number of fields read from a struct, -1 will read all fields",
                   "default": -1
-                },
-                "showGlobalVariables": {
-                  "type": "boolean",
-                  "description": "Boolean value to indicate whether global package variables should be shown in the variables pane or not.",
-                  "default": true
                 }
               },
               "description": "LoadConfig describes to delve, how to load values from target's memory",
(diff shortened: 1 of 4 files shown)


aschade92:
https://github.com/Microsoft/vscode-go/issues/2323

package.json
```diff
@@ -1233,7 +1238,8 @@
                 "maxVariableRecurse": 1,
                 "maxStringLen": 64,
                 "maxArrayValues": 64,
-                "maxStructFields": -1
+                "maxStructFields": -1,
+                "showGlobalVariables": false
```
ramya-rao-a: I believe we need to pull `showGlobalVariables` one level up.
The properties in LoadConfig is a 1:1 to what is at https://github.com/go-delve/delve/blob/4c9a72e486f1f0d0c90ecede8415a871dced8117/service/api/conversions.go#L276

aschade92:
@ramya-rao-a updated

ramya-rao-a:
@jhendrixMSFT FYI, I am changing the default value for `showGlobalVariables` to `false` as it is still causing perf issues for some users and the benefits of having it enabled doesnt outweigh the issues.

ramya-rao-a:
@aschade92 Looks like this is your first PR contribution to this project, Thanks & Welcome!

I have pushed 2 more commits. 1 to update the description and default value for the `dlvLoadConfig` setting, the other is explained at https://github.com/Microsoft/vscode-go/pull/2351#issuecomment-468959010

aschade92:
@ramya-rao-a sounds good, thanks for reviewing!

PR #2354, opened 2019-02-25
Changed swapped $2 and $0 position for anonymous go function snippet.

uroybd:
For anonymous go function (the snippet with the prefix "go") I think it will be more convenient if we end snippets in the body instead of the variable field. The function body is the largest of these 3 placeholders and will be in need of completion features more than any other. Therefore, I think, it is reasonable if we give it the $0 place.


jhendrixMSFT:
Good idea, thanks for the contribution.

PR #2377, opened 2019-03-14
test go 1.12 in CI

vladbarosan:
- test 1.12 in CI
 - remove 1.9 from CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: go
 
 go:
-  - 1.9.x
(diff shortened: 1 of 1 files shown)


vladbarosan:
@ramya-rao-a  since Golang team itself only supports last 2 minor versions, whats our policy on this ? 

I left 1.10, but I removed 1.9 to not overload the CI

ramya-rao-a:
We usually do 3 versions

PR #2392, opened 2019-03-20
Fix "Cancel Running Tests" option

Hittorp:
The original issue is: 
on _Linux_ machine when a user presses 'Cancel Running Tests' to stop a test that was running through 'run test' button in the code editor, the go-test process is orphaned and stays alive because runningTestProcesses.kill(...) kills the parent process only.

To fix this I suggest to kill spawned process by group (with 'minus PID' notation).

One more notice: may be killing the spawned processes with SIGKILL is too hard because SIGKILL cannot be handled. I think we can let them to have a chance to handle the interruption. So I suggest kill them with SIGTERM.

### DIFF
--- a/src/testUtils.ts
+++ b/src/testUtils.ts
@@ -6,6 +6,7 @@ import cp = require('child_process');
 import path = require('path');
 import util = require('util');
 import vscode = require('vscode');
+import process = require('process');
 
 import { getCurrentPackage } from './goModules';
 import { GoDocumentSymbolProvider } from './goOutline';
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Apologies for the very late reply here @Hittorp 

We have recently moved to using the [tree-kill](https://www.npmjs.com/package/tree-kill) module to kill processes and can now move away from using `SIGKILL` to cancel the tests.

The [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) of this extension has this change. Please give it a try and let me know if this solves the problem in linux for you.

ramya-rao-a:
If the [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) of this extension doesnt solve the problem, please log an issue and we will follow up.

Thanks for bringing this to our attention!

Issue #1746, opened 2018-06-20
Support for multiple envFile in launch.json


alexus1024:
There are multiple env files for each environment in my project - one for "public things" such as server addreses  and another for "private things" such as passwords. Currently I can pass only one of them with "envFile" attribute.

It would be cool to pass both of them in launch.json somehow.

alexus1024:
They all should be used. Each next file adds its contents to env and rewrite existing variables. So, order is significant.
Same as docker-compose files, for example. https://docs.docker.com/compose/compose-file/#env_file

ricochet:
I encountered this. My workflow is usually centered around kicking off a script to start a delve server and then launching my debug config in code. I worked around this by adding to that little script to create a mega-env.

```bash
ENV_GEN_FILE=".env_generated"
printf "#\n# Do not edit this file. It is generated!\n#\n" > $ENV_GEN_FILE

# Look for run specific details - first try in $HOME, then override with local
for ENV in $HOME/.env .env
do
    if [ -f $ENV ]; then
        cat $ENV >> $ENV_GEN_FILE
    fi
done
```

ramya-rao-a:
My apologies @alexus1024, for taking so long to get back to this issue.

So, following the example in https://docs.docker.com/compose/compose-file/#env_file, we can update our support of the `envFile` attribute to also take in either a string or an array of strings.

PRs are most welcome to add this feature.

Code Pointers:
- See [Build and Debug Extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code) to set up local dev environment
- The `package.json` file needs to be updated to support both string and array of string as valid inputs 
- We currently use [parseEnvFile](https://github.com/Microsoft/vscode-go/blob/0.8.0/src/goPath.ts#L116) to parse a given env file. We can add a new function `parseEnvFiles` that can take in an array of paths, and internally use `parseEnvFile` in the right order
- Then, update the types for the `envFile` property in `goDebug.ts` file and call `parseEnvFiles` instead of `parseEnvFile`

SteelPhase:
@ramya-rao-a could you take a look at #2395 and see if this is what you expected for a solution to this?

offero:
Now that I think about it, this should be contributed in a place where it can be used from multiple launch configurations such as node also.

jcbit:
I agree, I am working with node and looking for a way to use multiple envfiles

[...]

PR #2395, opened 2019-03-22
Add support for multiple environment files

SteelPhase:
I've attempted to resolve the issue presented in #1746. I'm sure some things will need to match coding standards and etc, but this initial implementation appears to work with a single file, and an array of many files.


SteelPhase:
Looks like gogetdoc isn't happy on osx 1.11.x test, but not sure it's related to my PR

package.json
```diff
@@ -463,7 +463,13 @@
...
               "envFile": {
-                "type": "string",
+                "type": [
+                  "string",
+                  "array"
+                ],
+                "items": {
+                  "type": "string"
+                },
                 "description": "Absolute path to a file containing environment variable definitions.",
```
quoctruong: The description should be updated to reflect the fact that both string and array are supported now.

src/debugAdapter/goDebug.ts
```diff
@@ -315,14 +315,23 @@ class Delve {
...
-			let fileEnv = {};
+			const envs = [process.env];
+
 			try {
-				fileEnv = parseEnvFile(launchArgs.envFile);
+				if (typeof launchArgs.envFile === 'string') {
+					envs.push(...parseEnvFiles([launchArgs.envFile]));
+				}
+
+				if (launchArgs.envFile instanceof Array) {
```
quoctruong: You can also do `Array.isArray(launchArgs.envFile)`.

src/debugAdapter/goDebug.ts
```diff
@@ -315,14 +315,23 @@ class Delve {
 			}
 
 			// read env from disk and merge into env variables
-			let fileEnv = {};
+			const envs = [process.env];
+
 			try {
-				fileEnv = parseEnvFile(launchArgs.envFile);
+				if (typeof launchArgs.envFile === 'string') {
+					envs.push(...parseEnvFiles([launchArgs.envFile]));
```
quoctruong: I think we can also do `envs.push(parseEnvFile(launchArgs.envFile))`

src/goPath.ts
```diff
@@ -142,6 +142,19 @@ export function parseEnvFile(path: string): { [key: string]: string } {
 	}
 }
 
+export function parseEnvFiles(paths: string[]): { [key: string]: string }[] {
+	const envs: { [key: string]: any }[] = [];
```
quoctruong: I think you can shorten this function with `return paths.map(path => parseEnvFile(path));`

package.json
SteelPhase: I've added a better description

SteelPhase:
@quoctruong I've made the suggested changes.

ramya-rao-a:
Apologies for not getting to this sooner @SteelPhase 
It somehow got dropped from my radar.

There have been a lot of changes to the `goDebug.ts` file.
Do you mind getting the latest from the master branch and resolving the conflicts here?

Issue #2423, opened 2019-04-06
go.mod - syntax highlighting for comments has some quirks


brainsnail:
There are a few scenarios where the comment highlighting doesn't work as expected without a space preceding the comment pattern -`//` :

**Doesn't recognize the comment**
* After the `go` keyword and version
* After the `module` keyword and module 
<img width="244" alt="image" src="https://user-images.githubusercontent.com/444193/55674293-c4e92400-5867-11e9-984c-10317b609fad.png">

**Recognizes the comment string, but breaks subsequent highlighting**
* After the `replace` keyword before a paren
<img width="1002" alt="image" src="https://user-images.githubusercontent.com/444193/55674298-d8948a80-5867-11e9-810b-2d496298505f.png">

_Originally posted by @brainsnail in https://github.com/Microsoft/vscode-go/issues/2410#issuecomment-480531910_

ramya-rao-a:
cc @calebdoxsey

calebdoxsey:
@ramya-rao-a @brainsnail I think https://github.com/Microsoft/vscode-go/pull/2424 should fix these issues.

ramya-rao-a:
This bug fix is now available in the latest update (0.11.0) to the Go extension. Thanks @calebdoxsey and @brainsnail!

PR #2424, opened 2019-04-07
go.mod: allow no space before comment, do not allow double slash in unqouted string

calebdoxsey:
I believe this will fix the issues identified in #2423 .

A space is no longer required after the directive, so both

```require// this is now a comment```

and

```require(```

should be matched.

Comments will also be detected within unquoted strings:

```
    module k8s.io/kubernetes// this is now a comment
```

Which I believe replicates this line from the parser: https://github.com/golang/go/blob/master/src/cmd/go/internal/modfile/read.go#L546.

Finally anything unmatched is marked as invalid.

### DIFF
--- a/syntaxes/go.mod.tmGrammar.json
+++ b/syntaxes/go.mod.tmGrammar.json
@@ -6,6 +6,9 @@
 		},
 		{
 			"include": "#directive"
+		},
+		{
+			"include": "#invalid"
 		}
 	],
 	"repository": {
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Thanks @calebdoxsey!

@brainsnail thoughts?

brainsnail:
Just loaded it up and tested out @calebdoxsey changes. Looks good to me regarding the comment formatting after the keywords. 

I'm not super sure about the formatting for unmatched/invalid lines though. 

Here's an example where I just entered a line break prior to a matched `(` - 
image

The invalid or unmatched highlighting is a nice addition though once that tweak is made. 

Thanks again, @calebdoxsey!

marco-m:
hello @ramya-rao-a would you have time to have a look?

marco-m:
thanks!

calebdoxsey:
Thanks @ramya-rao-a

PR #2518, opened 2019-05-15
support disabling document links, enable diagnostics by default

stamblerre:
The issue of stuck diagnostics seems to be solved, so we can now enable diagnostics by default.
Also, adds configurability for documentLink, which seems to have been requested by a few people, both on the issue tracker and on Slack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -980,8 +980,13 @@
             },
             "diagnostics": {
               "type": "boolean",
-              "default": false,
+              "default": true,
               "description": "If true, the language server will provide build, vet errors and the extension will ignore the `buildOnSave`, `vetOnSave` settings."
+            },
+            "documentLink": {
+              "type": "boolean",
+              "default": true,
+              "description": "If true, the language server's document link feature will be enabled."
             },
             "incrementalSync": {
               "type": "boolean",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -980,9 +980,14 @@
...
             "diagnostics": {
               "type": "boolean",
-              "default": false,
+              "default": true,
               "description": "If true, the language server will provide build, vet errors and the extension will ignore the `buildOnSave`, `vetOnSave` settings."
             },
+            "documentLink": {
+              "type": "boolean",
+              "default": true,
+              "description": "If true, the language server's document link feature will be enabled."
```
ramya-rao-a: Can you update the description to include a little more details on what this feature would mean to the end user?

stamblerre: Done.

package.json
```diff
@@ -980,9 +980,14 @@
...
             "diagnostics": {
               "type": "boolean",
-              "default": false,
+              "default": true,
               "description": "If true, the language server will provide build, vet errors and the extension will ignore the `buildOnSave`, `vetOnSave` settings."
             },
+            "documentLink": {
+              "type": "boolean",
+              "default": true,
+              "description": "If true, the language server will provide clickable Godoc links for import statements."
```
ramya-rao-a: This is so cool!

marlonfan: Looking forward to release

PR #2524, opened 2019-05-17
Use tsconfig.json to determine which typescript files to lint

stuartgrigg:
I noticed that the `npm run lint` command was not covering all the typescript files in the repo and was therefore missing some linting errors. I have changed this command and the `npm run fix-lint` command so that they use the tsconfig.json file to determine which files will be linted. I have also used the fix-lint command to fix the errors which my change had uncovered.


stuartgrigg:
I think the ci failures are the same as on master.

PR #2530, opened 2019-05-22
Add command to restart language server

### LINKED ISSUE #2500: [features request] Add a command to restart language server just like restart a typescript language server
The same as #2324 

Today I use vscode to develop a golang project , and find I am so in need this features if without it just can kill language server in task manager and wait vscode restart the language server 

the best developer also can't catch various situations, like move a go file to another package the language server can't know how to do and I lose tips , why not give the language server a choice to restart

SteelPhase:
This seems pretty straight forward, and does actually work. I would ask that this included for the time being in some form. Please let me know what you thing of this. I originally thought about adding some logging around it, but I didn't spot any example code on how to do just that.

This should resolve #2500

### DIFF
--- a/package.json
+++ b/package.json
@@ -300,6 +300,11 @@
         "command": "go.godoctor.var",
         "title": "Go: Extract to variable",
         "description": "Extract to variable using godoctor."
+      },
+      {
+        "command": "go.languageserver.restart",
+        "title": "Go: Restart Language Server",
+        "description": "Restart the running instance of the language server"
       }
     ],
     "debuggers": [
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
Thanks for the PR @SteelPhase! Apologies for not getting to this sooner.

There is one small issue of the restarted server not being added to the subscriptions that keep the list of disposables in the extension.

I have made the relevant changes in the PR, didnt get time to test it though.
Can you give it a try and let me know how it goes?

SteelPhase:
I will test it tomorrow, and post here with an update.

SteelPhase:
@ramya-rao-a it appears to be working as I'd expect.

Issue #2543, opened 2019-05-28
go.alternateTools not able to use relative paths or variable substitutions


pseudo-su:
I'm not sure if this is a feature request or a bug. I want to use `go.alternateTools` to specify a binary to execute relative to my workspace folder.

This works:
```json
"go.alternateTools": {
    "golangci-lint": "/absolute/path/to/my/project/bin/golangci-lint",
  },
```
These don't:
```json
"go.alternateTools": {
    "golangci-lint": "${workspaceFolder}/bin/golangci-lint",
    "golangci-lint": "./bin/golangci-lint",
    "golangci-lint": "bin/golangci-lint",
  },
```

Steps to Reproduce:

1. `git clone --branch bug/vscode-go-alternate-tools git@github.com:pseudo-su/golang-service-template.git`
2. Ensure `golangci-lint` is not available in PATH (so `vscode-go` doesn't fall-back to using it)
3. use command in readme to install `golangci-lint` into local `bin` directory
    ```shell
    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.16.0
    ```
4. expected behavior is for the linter to detect an issue on `http.ListenAndServe(":80", r)` because `err` is not handled. Running it directly with `./bin/golangci-lint run` does this correctly
5. If you then change the alternateTools config to an absolute path it will detect the issue in the IDE

ramya-rao-a:
Thanks for reporting @pseudo-su 
This is a valid request and PRs are welcome.

Code Pointers:
- [Set up local dev environment](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code)
- `getBinPathWithPreferredGopath()` function in `goPath.ts` file is the one that needs to be updated.
- In VS Code, the extension and the debug adapter registered by the extension run in separate processes.  The process for the debug adapter does not have access to the `vscode` module which is generally used to figure out the path to the current workspace. Since the code from `goPath.ts` is used in both processes, we avoid making references to any code that uses the `vscode` module in `goPath.ts` file. Therefore, the path resolution should happen before calling `getBinPathWithPreferredGopath()`

ramya-rao-a:
This feature is now available in the latest update (0.11.0) to the Go extension.

ramya-rao-a:
This feature is now available in the latest update (0.11.0) to the Go extension. Thanks @pseudo-su!

PR #2544, opened 2019-05-28
feat: allow workspaceFolder substition in go.alternateTools

pseudo-su:
This is a feature I'd like, raised an issue describing it [here](https://github.com/microsoft/vscode-go/issues/2543)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -371,7 +371,13 @@ function resolveToolsGopath(): string {
 }
 
 export function getBinPath(tool: string): string {
-	return getBinPathWithPreferredGopath(tool, (tool === 'go') ? [] : [getToolsGopath(), getCurrentGoPath()], vscode.workspace.getConfiguration('go', null).get('alternateTools'));
+	const alternateTools: { [key: string]: string; } = vscode.workspace.getConfiguration('go', null).get('alternateTools');
+
+	return getBinPathWithPreferredGopath(
+		tool,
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -371,7 +371,13 @@ function resolveToolsGopath(): string {
 }
 
 export function getBinPath(tool: string): string {
-	return getBinPathWithPreferredGopath(tool, (tool === 'go') ? [] : [getToolsGopath(), getCurrentGoPath()], vscode.workspace.getConfiguration('go', null).get('alternateTools'));
+	const alternateTools: { [key: string]: string; } = vscode.workspace.getConfiguration('go', null).get('alternateTools');
+
+	return getBinPathWithPreferredGopath(
+		tool,
+		(tool === 'go') ? [] : [getToolsGopath(), getCurrentGoPath()],
+		resolveObjectPaths(alternateTools),
```
ramya-rao-a: `getBinPathWithPreferredGopath` only needs to deal with the current tool.
So, instead of resolving the path for all the alternate tools, why not resolve the path for the given tool and just pass that to `getBinPathWithPreferredGopath` ?

i.e update `getBinPathWithPreferredGopath` to take `alternateToolPath` (a string) instead of `alternateTools` (a key-value map)

pseudo-su: That does make more sense, I'll change it now. I was focussing more on trying to make the build pass 😄.

PR #2583, opened 2019-06-19
Improve "generate interface stubs" interface

Ackar:
This improves the interface for the "generate interface stubs" command as per https://github.com/microsoft/vscode-go/issues/1823.
This adds a way to select a package and interface as well as automatic receiver name generation.

Example usage:
out

I couldn't find a way to list interfaces from a package with the current tools so I added my own, but I'd be happy to change that if I missed something.

### DIFF
--- a/src/goImpl.ts
+++ b/src/goImpl.ts
@@ -7,54 +7,251 @@
 
 import vscode = require('vscode');
 import cp = require('child_process');
-import { getBinPath, getToolsEnvVars } from './util';
+import { getBinPath, getToolsEnvVars, byteOffsetAt } from './util';
 import { promptForMissingTool } from './goInstallTools';
 import { dirname } from 'path';
+import { getAllPackages } from './goPackages';
 
-// Supports only passing interface, see TODO in implCursor to finish
-const inputRegex = /^(\w+\ \*?\w+\ )?([\w./]+)$/;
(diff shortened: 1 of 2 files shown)


kunzeal:
@Ackar Any progress of this? Looking forward to it.

Ackar:
I'm just waiting for review 🙂 

@ramya-rao-a Any thoughts on this? I'm happy to make any changes if needed.

pankrator:
Any chances of getting this nice functionality soon?

ramya-rao-a:
Hello all, thanks for your patience

@stamblerre, @hyangah Are you aware of any existing Go tools that can list interfaces from a given package?

ramya-rao-a:
Hello all,

We would prefer a solution where we don't have to add a dependency on yet another Go tool that will have to shelled out from the extension. Moreover, as mentioned in https://github.com/microsoft/vscode-go/issues/1823#issuecomment-629032526, further development on the stubbing interfaces is being explored via gopls. See https://github.com/golang/go/issues/37537.

Therefore, closing this PR.

Thanks for all your efforts here.
I'd like to encourage people who upvoted this PR and looking forward to this change to subscribe to https://github.com/golang/go/issues/37537 and take part in the discussion there.

PR #2584, opened 2019-06-19
Fixed GDPR annotation

lramos15:
There was a comma missing on this GDPR annotation which caused VS Code's telemetry tooling to fail when extracting the events.


ramya-rao-a:
Thanks!

PR #2611, opened 2019-06-28
Fix: convert fillstruct's output to character offset

stdupp:
The 'start', 'end' filed in fillstruct output is byte offset, but the parameter of 'editor.document.positionAt' is character offset, so if a source code include multi-byte string, vscode can't fill the struct into the right position.


stdupp:
PTAL @ramya-rao-a

nicoxb:
there was a conflict, and not merged? The issue is already exist.

stdupp:
please take a look @ramya-rao-a

nicoxb:
problem not fixed yet!

ramya-rao-a:
@stdupp, @nicoxb Can either of you provide a small code snippet where we can try to replicate the problem?

nicoxb:
@ramya-rao-a Hi, there are some code snippet:
```go
type User struct {
	ID   int64
	Name string
	Age  int
}
// expected
func foo1() {
	u := &User{
		ID:   0,
		Name: "",
		Age:  0,
	}
	fmt.Println(u)
}

// unexpected
func foo2() {
	// multi-byte char
	// 多字节字符
	u := &User{}
	fmUser{
		ID:   0,
		Name: "",
		Age:  0,
	}tln(u)
}
```

ramya-rao-a:
The latest update 0.14.2 has the fix for this, Thanks @stdupp!

PR #2625, opened 2019-07-02
goModules: change gopls prompt message

stamblerre:
This change clarifies that gopls is experimental when prompting the user.


stamblerre:
Makes sense. Added a bit about modules, since the better performance should be for all language features, not just completion.

Issue #2559, opened 2019-06-04
No warning when debugging a file that has been modified after starting the debugging session


aarzilli:
1. Debug any program
2. Modify the source code without stopping the debugger
3. Continue using the debugger

There is no warning that the executable being debugged is stale,m now the line numbers reported by the debugger will be nonsensical and behavior of the debugged program will diverge from what would be expected from the source code displayed on screen.

ramya-rao-a:
Fair point
We have listeners on each file save `vscode.workspace.onDidSaveTextDocument`
We have a way to figure out if there is an active debug session `vscode.debug.activeDebugSession`

Therefore, we can definitely use the two and show a warning that a file change has been saved while the debug session is in progress.

Thanks for reporting!

ramya-rao-a:
PRs are welcome to add this feature

Code Pointers:
- [Set up local dev environment](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code) for the Go extension
- The `activate()` function in the `goMain.ts` is where all the listeners are added and so here is where the warning (`vscode.window.showWarningMessage`) should be added

haryps:
Hi @ramya-rao-a,

I would like to fix this issue and i have made the PR as attached.

However, i wonder whether this must be fix from each extension or visual studio code core code.

ramya-rao-a:
> However, i wonder whether this must be fixed from each extension or visual studio code core code.

VS Code will not be aware of which code files should be allowed to be edited and which shouldnt when a debug session is started. Therefore, it cannot make that decision and each language extension is in the best place to make this decision.

> Second, should the warning message added when the file has been changed or only when the file has been changed then saved? Please advise.

On save should be enough.

> In visual studio, there is condition where we are not allowed to edit code during debugging. Can we identify that condition in this extension?

Can you elaborate?

ramya-rao-a:
There is now a warning that shows up if you save changes to a Go file when there is an active debug session

PR #2653, opened 2019-07-15
Show warning message when modifying debugged file

haryps:
Fix issue #2559.

Show warning message when the file is edited then saved.


ramya-rao-a:
Thanks for the PR @haryps 
Sorry it took a while to get to it :)

PR #2660, opened 2019-07-18
Added configuration for alternate gopls path.

anurbol:
This is to allow configuring an alternate path for gopls as well. Everything else in the extension is ready for this configuration, but the configuration feature itself is missing. By "ready" I mean these lines:
https://github.com/microsoft/vscode-go/blob/4b516f9a872f3a3086d6d56f8e346d4b5a7e53ef/src/goInstallTools.ts#L559
https://github.com/microsoft/vscode-go/blob/4b516f9a872f3a3086d6d56f8e346d4b5a7e53ef/src/util.ts#L375

### DIFF
--- a/package.json
+++ b/package.json
@@ -1500,6 +1500,11 @@
               "default": "gopkgs",
               "description": "Alternate tool to use instead of the gopkgs binary or alternate path to use for the gopkgs binary."
             },
+            "gopls": {
+              "type": "string",
+              "default": "gopls",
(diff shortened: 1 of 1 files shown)


anurbol:
@ramya-rao-a Hi! I am making a tooling software and I want to reuse gopls' symbols to avoid parsing source files again. To do that I want to use a proxy between gopls and vscode-go.

ramya-rao-a:
Understood, good luck!

Issue #1921, opened 2018-09-10
Support codelens to run and debug tests supported by the check package


ramya-rao-a:
Tests that use the [check](http://labix.org/gocheck) package can be run as part of the whole package or file but not individually in VS Code.

This feature request is to support the below for individual tests using the check package in VS Code
- `Go: Test function at cursor` command
- Run Tests codelens
- Debug Tests codelens

Charliekenney23:
I'll take this!

ramya-rao-a:
@Charliekenney23 Thanks! Do you need any pointers?

Charliekenney23:
@ramya-rao-a Yeah, pointers would be great if you have the time!

mapuri:
@ramya-rao-a @Charliekenney23 Is this issue still being worked on? 

I would benefit from this feature. One of the repos that I work with uses the check package and vscode-go won't let me run/debug those tests individually.

Charliekenney23:
Please feel free to take this! I'm no longer effected by this issue and forgot about it as a result.

mapuri:
@Charliekenney23 thanks for letting me know. I haven't coded in node yet but I can give it a shot :)

@ramya-rao-a following your code pointers I think we would need to add the go-check package related flags in [targetArgs](https://github.com/Microsoft/vscode-go/blob/0.7.0/src/testUtils.ts#L356) function that is called from the [goTest](https://github.com/Microsoft/vscode-go/blob/0.7.0/src/testUtils.ts#L199) function. We don't yet have the information about the imported packages in that function but we can get that in similar lines from the **editor** object as we do it in [getTestFunctions](https://github.com/Microsoft/vscode-go/blob/0.7.0/src/testUtils.ts#L102). 

```node
function targetArgs(testconfig: TestConfig): Array<string> {

	const editor = vscode.window.activeTextEditor;
	if (!editor) {
		vscode.window.showInformationMessage('No editor is active.');
		return;
	}
	if (!editor.document.fileName.endsWith('_test.go')) {
		vscode.window.showInformationMessage('No tests found. Current file is not a test file.');
		return;
	}
...
	if (testMethods.length > 0) {
		if ( containsThirdPartyTestPackages(editor.document, null, [goCheckPkg]) ) {
			params = params.concat(['-check.f', util.format('^(%s)$', testMethods.join('|'))]);	
		} else if (containsThirdPartyTestPackages(editor.document, null, [testifyPkg])) {
[...]

PR #2680, opened 2019-08-02
add run test and debug codelens support for tests that use gocheck test package

mapuri:
fixes: https://github.com/microsoft/vscode-go/issues/1921

I have tried the changes locally by launching the extension in debug mode and they seem to work for all the 3 scenarios mentioned in #1921 but it is possible I might have overlooked other cases. I briefly checked the existing unit tests but I couldn't figure how to add tests for this change using those. I will appreciate any pointers to tests or utils that I could use to add unit tests for my changes.

Also this is my first PR for the go plugin (as well in node/ts) so it may not be up to the mark wrt to the language style and the conventions of this project. I would be happy to adjust and improve the PR as needed.


mapuri:
The test failures seem unrelated to my changes but please let me know if I am missing something.

mapuri:
closing the PR as it is out of date with master and I won't be able to work on it in near future.

Issue #2592, opened 2019-06-22
Debug with attach will terminate remote process when detach


BaronChen:
I believe this PR https://github.com/microsoft/vscode-go/pull/2125 suppose to support attaching and detaching from remote debug server without killing the process. However, when I host the GO process with delve in a docker, the remote debug process still killed the remote process. 
Here is my vs code launch.json:
```
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Attach: Debug users service",
      "type": "go",
      "request": "attach",
      "mode": "remote",
      "remotePath": "/go/src/app",
      "cwd":"${workspaceFolder}/users",
      "port": 2345,
      "host": "127.0.0.1",
      // "debugServer": 4711,
      "apiVersion": 2
    }
  ]
}
```
delve is running with:
```
/go/bin/dlv debug /go/src/cmd/server --accept-multiclient -l 0.0.0.0:2345 --headless=true --log --log-output debugger,rpc --api-version=2 
```

And the delve log:
```
Recreating bc-system-go_users_1 ... done
Attaching to bc-system-go_users_1
users_1  | API server listening at: [::]:2345
users_1  | time="2019-06-22T05:36:27Z" level=info msg="launching process with args: [/go/src/app/debug]" layer=debugger
users_1  | time="2019-06-22T05:36:31Z" level=debug msg="<- RPCServer.GetVersion(api.GetVersionIn{})" layer=rpc
users_1  | time="2019-06-22T05:36:31Z" level=debug msg="-> *api.GetVersionOut{\"DelveVersion\":\"Version: 1.2.0\\nBuild: $Id: 068e2451004e95d0b042e5257e34f0f08ce01466 $\",\"APIVersion\":2} error: \"\"" layer=rpc
[...]

PR #2685, opened 2019-08-06
Polish remote debug experience

quoctruong:
Fix a bug where setting breakpoint fails if the remote program is started through `dlv` with `--continue` switch (https://github.com/microsoft/vscode-go/issues/2690).
Fix a bug where disconnecting (after attaching to) the remote program terminates it (https://github.com/microsoft/vscode-go/issues/2592).
Fix a bug where setting breakpoint will fail if a breakpoint already exists (if dlv is started through multi client and another client sets the breakpoint).


src/debugAdapter/goDebug.ts
```diff
@@ -781,6 +776,11 @@ class GoDebugSession extends LoggingDebugSession {
 
 	protected disconnectRequest(response: DebugProtocol.DisconnectResponse, args: DebugProtocol.DisconnectArguments): void {
 		log('DisconnectRequest');
+		// For remote process, we have to issue a continue request
+		// before disconnecting.
+		if (this.delve.request === 'attach' && !this.delve.debugProcess) {
```
ramya-rao-a: request can be launch and the user could be using the old of way doing remove debugging.

I would suggest adding a property `isLocalDebugging` on `this.delve` so that we can use that directly instead of checks like this

src/debugAdapter/goDebug.ts
```diff
@@ -781,6 +776,11 @@ class GoDebugSession extends LoggingDebugSession {
 
 	protected disconnectRequest(response: DebugProtocol.DisconnectResponse, args: DebugProtocol.DisconnectArguments): void {
 		log('DisconnectRequest');
+		// For remote process, we have to issue a continue request
+		// before disconnecting.
```
ramya-rao-a: Is it enough that the continue request is issued?
Do we need to wait for the response before moving ahead?

src/debugAdapter/goDebug.ts
```diff
@@ -869,26 +869,34 @@ class GoDebugSession extends LoggingDebugSession {
 				breakpointIn.loadLocals = this.delve.loadConfig;
 				breakpointIn.cond = breakpoint.condition;
 				return this.delve.callPromise('CreateBreakpoint', [this.delve.isApiV1 ? breakpointIn : { Breakpoint: breakpointIn }]).then(null, err => {
+					if (err.toString().startsWith('Breakpoint exists at')) {
```
ramya-rao-a: Is there no other way to check for this error other than checking the error message this way? Maybe an error code or error name?
Error messages can change in the future, so I am a little weary of depending on them this way.

quoctruong:
@ramya-rao-a PTAL

jhendrixMSFT:
@quoctruong when running dlv with `--continue`, if I set a breakpoint after remote-attach I don't see it being set.  However if the breakpoint is set then I remote-attach it is set.  Could you please take a look?

quoctruong:
@jhendrixMSFT I believe this should fix it. I tried detaching and setting breakpoints a few times. Thanks for the catch!

src/debugAdapter/goDebug.ts
```diff
@@ -900,12 +946,21 @@ class GoDebugSession extends LoggingDebugSession {
...
+		try {
+			// If a program is launched with --continue, the program is running
+			// before we can run attach. So we would need to check the state.
+			// We use NonBlocking so the call would return immediately.
+			this.debugState = await this.delve.getDebugState();
+		} catch (error) {
+			logError(`Failed to get state ${String(error)}`);
+		}
+
+		if (!this.debugState.Running && !this.continueRequestRunning) {
```
ramya-rao-a: If you failed to get the state in the above step, then wont debugState be null/undefined?

Issue #1919, opened 2018-09-08
Tools install into open workspace if go.mod present


FiloSottile:
After upgrading my macOS system Go to 1.11, two tools fail to install.

```
Installing 2 tools at /Users/filippo/.vscode/gopath/bin
  megacheck
  gotests

Installing honnef.co/go/tools/... FAILED
Installing github.com/cweill/gotests/... FAILED

2 tools failed to install.

megacheck:
Error: Command failed: /usr/local/bin/go get -u -v honnef.co/go/tools/...
Fetching https://honnef.co/go/tools?go-get=1
Parsing meta tags from https://honnef.co/go/tools?go-get=1 (status code 200)
get "honnef.co/go/tools": found meta tag get.metaImport{Prefix:"honnef.co/go/tools", VCS:"git", RepoRoot:"https://github.com/dominikh/go-tools"} at https://honnef.co/go/tools?go-get=1
go: finding honnef.co/go/tools/... latest
Fetching https://honnef.co/go/tools?go-get=1
Parsing meta tags from https://honnef.co/go/tools?go-get=1 (status code 200)
get "honnef.co/go/tools": found meta tag get.metaImport{Prefix:"honnef.co/go/tools", VCS:"git", RepoRoot:"https://github.com/dominikh/go-tools"} at https://honnef.co/go/tools?go-get=1
go: finding honnef.co/go/tools latest
Fetching https://honnef.co/go?go-get=1
Parsing meta tags from https://honnef.co/go?go-get=1 (status code 200)
Fetching https://honnef.co?go-get=1
Parsing meta tags from https://honnef.co?go-get=1 (status code 200)
go get honnef.co/go/tools/...: no matching versions for query "latest"
Fetching https://honnef.co/go/tools?go-get=1
Parsing meta tags from https://honnef.co/go/tools?go-get=1 (status code 200)
get "honnef.co/go/tools": found meta tag get.metaImport{Prefix:"honnef.co/go/tools", VCS:"git", RepoRoot:"https://github.com/dominikh/go-tools"} at https://honnef.co/go/tools?go-get=1
go: finding honnef.co/go/tools/... latest
Fetching https://honnef.co/go/tools?go-get=1
Parsing meta tags from https://honnef.co/go/tools?go-get=1 (status code 200)
[...]

PR #2700, opened 2019-08-15
Install tools with modules enabled

stamblerre:
This change installs the Go tools with modules enabled, if the user is on Go >= 1.11. The reason for this is to ensure that the user gets the latest tagged version for a tool, rather than the master (which can be unstable).

In particular, this is necessary for `gopls`, since we have begun tagging stable versions. To avoid modifying the user's `go.mod` file (https://github.com/microsoft/vscode-go/issues/1919), we run the `go get` command in a temporary directory.


src/goInstallTools.ts
```diff
@@ -323,7 +330,15 @@ export function installTools(missing: string[], goVersion: SemVersion): Promise<
...
 
 	return missing.reduce((res: Promise<string[]>, tool: string) => {
-		return res.then(sofar => new Promise<string[]>((resolve, reject) => {
+		// Disable modules for staticcheck and gotests,
+		// which are installed with the "..." wildcard.
+		// TODO: ... will be supported in Go 1.13, so enable these tools to use modules then.
+		if (modulesOff || tool === "staticcheck" || tool === "gotests") {
+			envForTools['GO111MODULE'] = 'off';
+		} else {
+			envForTools['GO111MODULE'] = 'on';
```
ramya-rao-a: How does this affect tools that have not onboard their source code to use modules?

src/util.ts
```diff
@@ -81,6 +81,11 @@ let vendorSupport: boolean = null;
 let telemtryReporter: TelemetryReporter;
 let toolsGopath: string;
 
+// Assumes the Go version is always 1.x.
+export function isBelow(goVersion: SemVersion, minor: number): boolean {
```
ramya-rao-a: I'd prefer that this function take `major` as an argument and skip the assumption that the major version is always 1

src/goInstallTools.ts
```diff
@@ -369,21 +384,36 @@ export function installTools(missing: string[], goVersion: SemVersion): Promise<
...
-				cp.execFile(goRuntimePath, args, { env: envForTools }, (err, stdout, stderr) => {
-					if (stderr.indexOf('unexpected directory layout:') > -1) {
-						outputChannel.appendLine(`Installing ${tool} failed with error "unexpected directory layout". Retrying...`);
-						cp.execFile(goRuntimePath, args, { env: envForTools }, callback);
-					} else if (!err && tool.endsWith('-gomod')) {
-						const outputFile = path.join(toolsGopath, 'bin', process.platform === 'win32' ? `${tool}.exe` : tool);
-						cp.execFile(goRuntimePath, ['build', '-o', outputFile, getToolImportPath(tool, goVersion)], { env: envForTools }, callback);
-					} else {
-						callback(err, stdout, stderr);
+				// Install tools in a temporary directory, to avoid altering go.mod files.
```
ramya-rao-a: Do we need the temp dir to be around after the tools get installed?

src/goInstallTools.ts
stamblerre: It won't. You can still download tools that don't have `go.mod` files when you are in module mode.

src/goInstallTools.ts
stamblerre: Nope, added a call to remove it.

src/util.ts
stamblerre: Done.

src/goInstallTools.ts
```diff
@@ -369,21 +387,37 @@ export function installTools(missing: string[], goVersion: SemVersion): Promise<
...
+				fs.mkdtemp(path.join(os.tmpdir(), 'go-tools-'), (err, toolsTmpDir) => {
```
ramya-rao-a: We use a specific dir inside the tmp dir in this extension. Can you use that here instead?

```suggestion
				fs.mkdtemp(getTempFilePath('go-tools')), (err, toolsTmpDir) => {
```

ramya-rao-a: Also, do we need to create this temp dir for each tool? Why not create it once outside the `reduce()` call?

ramya-rao-a: Few more thoughts on the temp dir
- Is there a reason to have a different dir for each tool? I would imagine we create the dir once outside of the `reduce()` call and use it for all the tools
- We have a utility `getTempFilePath` that you can use which will give you a path under the temp dir that is created for this extension and is used for other purposes as well. I suggest you use this `getTempFilePath('go-tools')` to get a path and create a directory for this path using `fs.mkdirSync`
- When extension is de-activated (say when VS Code window closes or reloads), the folder would get cleaned up

src/goInstallTools.ts
```diff
@@ -369,21 +387,37 @@ export function installTools(missing: string[], goVersion: SemVersion): Promise<
...
+					if (err) {
+						resolve([...sofar, null]);
+						return;
 					}
+					const args = ['get', '-v'];
+					// Only get tools at master if we are not using modules.
+					if (modulesOff) {
+						args.push('-u');
```
ramya-rao-a: In module mode we don't need -u to get the latest?

PR #2701, opened 2019-08-15
separate tool information from installation

stamblerre:
This change refactors some of the code in `src/goInstallTools.ts`. Create a type for a `Tool` and add some helper functions to simplify a bit of the code. This should generally make it easier to add or remove a new tool.

PR #2703, opened 2019-08-17
Refactor language server handling

stamblerre:
This change refactors `goMain.ts` to move everything related to the language server into its own file, `goLanguageServer.ts`. It's not possible to totally decouple the language server from the rest of the initialization because, if a user disables a certain feature from the language server, we still need to register a provider for that feature. However, this change will simplify things going forward, as, now, everything related to the language server is in a single file.

I also removed the special handling for `bingo` flags. It seems to me that this complicates the extension unnecessarily. It should be fairly simple to request that a `bingo` user who is used to providing a flag simply switch to disabling diagnostics via the VS Code config.


stamblerre:
This change is stacked on top of https://github.com/microsoft/vscode-go/pull/2701.

src/goLanguageServer.ts
```diff
@@ -0,0 +1,377 @@
...
+		toolName,
+		{
+			command: path,
+			args: ['-mode=stdio', ...config.flags],
+			options: {
+				env: getToolsEnvVars(),
+			},
+		},
+		{
+			initializationOptions: {},
```
ramya-rao-a: The reason to skip `funcSnippetEnabled` and `incrementalSync` from initialization options is?

src/goLanguageServer.ts
```diff
@@ -0,0 +1,377 @@
...
+					if (!config.features.format) {
+						return [];
+					}
+					return next(document, options, token);
+				},
+				provideCompletionItem: async (document: vscode.TextDocument, position: vscode.Position, context: vscode.CompletionContext, token: vscode.CancellationToken, next: ProvideCompletionItemsSignature) => {
+					if (!config.features.completion) {
+						return [];
+					}
+					return next(document, position, context, token);
```
ramya-rao-a: When completion is coming from the language server, we still want to include some completion items from the extension like the completions for unimported packages. The code for that was present in `goMain.ts` file. Any reason for not keeping it?

src/goLanguageServer.ts
stamblerre: We don't actually respect those when passed into `gopls` so they have no effect right now. `incrementalSync` is already enabled by default, so the correct setting would actually be `noIncrementalSync`, but that's only useful for clients that don't support sending diffs. We are still working on a more thorough way to pass in configs, so I'll send a follow-up PR when we work out something that will be effective for all editors.

src/goLanguageServer.ts
stamblerre: We just added support for this feature in `gopls` (just standard library completions for now, but packages in the module cache/`GOPATH` will be coming soon). I expect that this will be completed before the next release of the VSCode extension.

stamblerre:
Thank you for making those fixes!

src/goLanguageServer.ts
ramya-rao-a: makes sense, thanks!

src/goLanguageServer.ts
ramya-rao-a: Below is what comes from the extension even when language server is enabled
- Completion for user tries to type in `package`. The completion item provided is for the package name based on the folder
- Completion if you are typing the first word in a line comment right above an exported member. This simply provides the name of the exported member as the completion item
- Completion for keywords. `gocode` never had this. Does `gopls` plan to add this? It would be very straight forward to do so.
- Completion for unimported packages. When used chooses one of these items, not only is the current word completed, but an import statement gets added too

Can you share the plans on each of the above in gopls?

I would prefer we keep the above for now, and then remove them one by one as `gopls` adds support for them

stamblerre: Completion for comments, keywords, and package names we could probably do fairly simply in `gopls`. I will file feature requests for these. Unimported packages is coming in the next version of `gopls`.

My reasoning for removing this code is primarily because `gopls` has a fair bit of logic for ranking completions, and adding additional completions from the extension will likely mess with the ordering in some way. Also, having this combination of completions hides the areas in which `gopls` is lacking - I wouldn't have known about these additional features if you hadn't just explained them to me right now. If we want to move all completion logic into `gopls`, I think we need begin the process of noticing where `gopls` is lacking. Maybe as a compromise we could add a temporary setting to force completions to be `gopls`-only?

ramya-rao-a: My only concern with this approach is the noise of incoming GitHub issues saying that the completion for unimported packages is missing when not using modules. But if you believe this feature will get added to gopls before we release an update to this extension, then we should be good as the answer to such issues would be to update gopls

PR #2711, opened 2019-08-23
Fix typings for goVersion and undefined error when goVersion is null

### LINKED ISSUE #2705: Install/Update tools fails silently when go binary is built from non-stable release
### Steps to Reproduce

1. Build go from source at go1.13beta1
```sh
$ pwd
/home/goodwine/go

$ go version
go version devel +60f14fddfe Wed Jun 26 16:35:14 2019 +0000 linux/amd64

$ git log -n 1
commit 60f14fddfee107dedd76c0be6b422a3d8ccc841a (HEAD, tag: go1.13beta1)
[...]
```

2. In VSCode `Ctrl + Shift + P` -> "Go: Install/Update Tools"
    * This works!
3. Go to VSCode settings and add:
```json
  "go.useLanguageServer": true,
```
4. In VSCode `Ctrl + Shoft + P` -> "Go: Install/Update Tools"
    * This **doesn't** work!

### Why it happens
I opened the devtools screen and saw this exception:
```
TypeError: Cannot read property 'major' of null
	at getTools (/home/goodwine/.vscode/extensions/ms-vscode.go-0.11.4/out/src/goInstallTools.js:117)
	at getMissingTools (/home/goodwine/.vscode/extensions/ms-vscode.go-0.11.4/out/src/goInstallTools.js:507)
	at util_1.getGoVersion.then.goVersion (/home/goodwine/.vscode/extensions/ms-vscode.go-0.11.4/out/src/goInstallTools.js:444)
```

Then I look at the code that checks whether to display `gopls` and noticed that there's a check [here](https://github.com/microsoft/vscode-go/blob/094225996e3ee75a1561eb190702a853dbb36f84/src/goInstallTools.ts#L125-L127) that looks at `goVersion` to see that the semver is > 1.10

The problem is that for builds that aren't from a stable release, the goversion obtained from utils.ts is `null`.

**Related issue**: https://github.com/golang/go/issues/11972

Goodwine:
Fixes #2705

### DIFF
--- a/src/goDeclaration.ts
+++ b/src/goDeclaration.ts
@@ -59,7 +59,7 @@ export function definitionLocation(document: vscode.TextDocument, position: vsco
 		goConfig = vscode.workspace.getConfiguration('go', document.uri);
 	}
 	const toolForDocs = goConfig['docsTool'] || 'godoc';
-	return getGoVersion().then((ver: SemVersion) => {
+	return getGoVersion().then(() => {
 		return getModFolderPath(document.uri).then(modFolderPath => {
 			const input: GoDefinitionInput = {
 				document,
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -46,7 +46,7 @@ const allToolsWithImportPaths: { [key: string]: string } = {
 	'godoctor': 'github.com/godoctor/godoctor',
 };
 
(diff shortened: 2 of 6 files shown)


ramya-rao-a:
My apologies @Goodwine 
I didn't see your PR in time before I pushed a fix for this bug :(

PR #2719, opened 2019-08-29
check the gopls version and update if outdated

stamblerre:
Sorry about the huge diff, this PR is actually stacked on top of all of my previous ones. The main interesting change is at the very end of the `goLanguageServer.ts` file, where we detect the `gopls` version and determine if it is the most recent one.


ramya-rao-a:
@stamblerre If the main change is only in the `goLanguageServer.ts` file, then I would suggest to close this PR and make the relevant changes in your other PR #2703

Your older PR #2701 has been merged and I have updated PR #2703 with the latest changes from the master branch along with a few other fixes

stamblerre:
This change adds a few dependencies, so I figured it would make sense to logically separate it from the previous one. I'm happy to wait for #2703 to be merged, and then I can resolve the merge conflicts so that you can review this one. Thanks!

ramya-rao-a:
@stamblerre Your other PRs have been merged to master. Once you resolve the merge conflicts in this PR, I'll start reviewing.

stamblerre:
Thanks for the reviews! I just resolved the merge conflicts here, so this change is ready for review.

src/goInstallTools.ts
```diff
@@ -112,7 +112,7 @@ export function installTools(missing: Tool[], goVersion: SemVersion): Promise<vo
 	// This ensures that users get the latest tagged version, rather than master,
 	// which may be unstable.
 	let modulesOff = false;
-	if (isBelow(goVersion, 1, 11)) {
+	if (semver.lt(goVersion, '1.11.0')) {
```
ramya-rao-a: How does `semver.lt` behave when the first argument is null?
Same question for `gt` or other comparison functions that are being used.

In our case `goVersion` will be null when using Go from the tip
You will need to add a null check on `goVersion` every time we use the `lt`, `gt` function

src/goLanguageServer.ts
```diff
@@ -379,3 +393,96 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
...
+
+	// The second line should be the sum line.
+	// It should look something like this:
+	//
+	//    golang.org/x/tools/gopls@v0.1.3 h1:CB5ECiPysqZrwxcyRjN+exyZpY0gODTZvNiqQi3lpeo=
+	//
+	// TODO: We should use a regex to match this, but for now, we split on the @ symbol.
+	// The reasoning for this is that gopls still has a golang.org/x/tools/cmd/gopls binary,
+	// so users may have a developer version that looks like "golang.org/x/tools@(devel)".
+	const moduleVersion = lines[1].trim().split(' ')[0];
```
ramya-rao-a: Why split at ' ' at all?
Why not do a direct `const split = lines[1].trim().split('@');`?

src/goLanguageServer.ts
```diff
@@ -379,3 +393,96 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
...
+
+	// Get the relevant portion, that is:
+	//
+	//    golang.org/x/tools/gopls@v0.1.3
+	//
+	const split = moduleVersion.trim().split('@');
```
ramya-rao-a: `moduleVersion` can be `undefined` if `lines[1]` had no `' '`

stamblerre: I don't believe it would ever be. I tested a few cases, and in the case where there is no element to split on, split just returns an array of 1 element containing the original string.

src/goInstallTools.ts
stamblerre: I reworked some of the Go version handling to tackle developer versions. There's definitely more that can be done to improve handling, but this should work for now.

src/goPackages.ts
```diff
@@ -6,8 +6,9 @@
 import vscode = require('vscode');
 import cp = require('child_process');
 import path = require('path');
+import semver = require('semver');
```
ramya-rao-a: This import is no longer used.

```suggestion
```

src/goTools.ts
```diff
@@ -5,8 +5,9 @@
 
 'use strict';
 
-import { SemVersion, isAbove, getGoConfig, isBelow } from './util';
+import vscode = require('vscode');
```
ramya-rao-a: This import is not needed.

```suggestion
```

src/goVet.ts
```diff
@@ -4,8 +4,9 @@
  *--------------------------------------------------------*/
 
 import path = require('path');
+import semver = require('semver');
```
ramya-rao-a: Import no longer needed.

```suggestion
```

src/testUtils.ts
```diff
@@ -6,6 +6,7 @@ import cp = require('child_process');
 import path = require('path');
 import util = require('util');
 import vscode = require('vscode');
+import semver = require('semver');
```
ramya-rao-a: Import no longer needed.

```suggestion
```

stamblerre:
Fixed all the unused imports, sorry about that. I wonder why the linter didn't report them.

src/util.ts
```diff
@@ -71,27 +72,58 @@ export const goBuiltinTypes: Set<string> = new Set<string>([
...
+		sendTelemetryEvent('getGoVersion', { version: this.format() });
+	}
 
-let goVersion: SemVersion = null;
-let vendorSupport: boolean = null;
-let telemtryReporter: TelemetryReporter;
-let toolsGopath: string;
+	format(): string {
+		if (this.sv) {
+			return `${this.sv.major}.${this.sv.minor}`;
```
ramya-rao-a: Apparently there is a property "raw" that would give the version in the string format. Have you tried that?

src/goLanguageServer.ts
```diff
@@ -379,3 +393,96 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
...
+
+	// Coerce the versions into SemVers so that they can be sorted correctly.
+	const versions = [];
+	for (const version of data.trim().split(os.EOL)) {
+		versions.push(semver.coerce(version));
+	}
+	if (versions.length === 0) {
+		return false;
+	}
+	versions.sort();
```
ramya-rao-a: This wouldn't work that well... We will have to pass a compareFn to sort()
While 1.12 would come after 1.11, 1.9 will also come after 1.11

ramya-rao-a: Something like the below
```javascript
versions.sort((a,b) => {
    if (semver.lt(a,b)) {
        return -1;
    }
    if (semver.gt(a,b)) {
        return 1;
    }
    return 0;
});
```

stamblerre: Done. Sorry, I must have misread some documentation and assumed that the correct comparison would be applied automatically in the sort.

src/util.ts
stamblerre: Used the `.format()` function instead. Thanks!

Issue #2690, opened 2019-08-08
Remote Attach fails to set a breakpoint if program is started with `--continue` option


quoctruong:
Steps to Reproduce:

1. Start a remote program through `dlv` with `--continue` option (recently added to `dlv` through https://github.com/go-delve/delve/pull/1585). For example:
```
dlv exec /app --continue --accept-multiclient --api-version=2 --headless --listen=:3000
```
2. Attach the VSCode Go Debugger and set a breakpoint.
3. Breakpoints won't get hit.

nagstaku:
I would guess that maybe there is a mis-match of files on your local system to the remote system and that is why it's never hitting (or at least that is what I experienced recently).

the applicable settings I had to set were:
"remotePath": "/app/root/dir/",
"cwd": "${workspaceFolder}/app/root/dir"

Although you should also post the rest of your launch.json

The vscode debug experience seems very rough currently, it doesn't seem to indicate if it's even able to reach a debug host for instance (ie, the port isn't even open on the target system, and never even connects).

You can add:
"tracing": "verbose",

and examine its output to look for oddities. Also, if possible, consider adding tcpflow to your target system, and then running this while attempting to attach debuggers (this is how i initially found my problems). For an alpine based docker container:

apk add tcpflow
tcpflow -i any -c

(-i is interface, and -c is Console print. Print the contents of packets, without storing in a file.)

quoctruong:
I actually have a PR (https://github.com/microsoft/vscode-go/pull/2685) to fix this bug. It seems to happen because VSCode does not halt `dlv` before setting a breakpoint in this case.

ramya-rao-a:
This should now be fixed in the latest version (0.11.5) of the extension

PR #2744, opened 2019-09-10
Fix a bug where clearing breakpoints failed

quoctruong:
This is related to https://github.com/microsoft/vscode-go/issues/2690. When we remote attach a program and set a breakpoint, if we detach and reattach, we cannot clear the breakpoints because we are not retrieving the breakpoint ID.


ramya-rao-a:
@jhendrixMSFT Can you please review this?

ramya-rao-a:
@jhendrixMSFT This looks good to me. Please do take a look from your side as well

PR #2745, opened 2019-09-10
Handle pseudoversions when updating gopls version

stamblerre:
This change handles the possibility of a user having a pseudoversion for their `gopls` version.


ramya-rao-a:
@stamblerre Can you elaborate the scenario we are trying to fix here?

stamblerre:
In my previous change, I had forgotten to account for the fact that users may have downloaded `gopls` manually by running `go get golang.org/x/tools/gopls@master`. This would result in the user getting a [pseudoversion](https://tip.golang.org/cmd/go/#hdr-Pseudo_versions), which we should also handle correctly. This change detects pseudoversions and compares the timestamp to that of the latest version by querying https://proxy.golang.org/golang.org/x/tools/gopls/@v/v0.1.3.info (or whatever the number is for the latest version).

Issue #2738, opened 2019-09-09
Test Plan for September 2019 release


ramya-rao-a:
If you got a notification from this issue, then you are part of the next update to the Go extension which I plan to release early next week.
Thanks and Welcome!

Here, I am listing some of the features and the bug fixes that will go in the next update and for which I need some help in testing. Since I dont work with Go in my day job, I always feel that someone who does Go more than me can help in testing tremendously.

And when I say "test", you can do a minimum pass of just installing the update and then do nothing special other than going about using VS Code like you do everyday. Or you can test the feature you requested/implemented or the bug you logged/fixed. Or you can test all of the below. Every bit helps.

Drop in a comment in this issue after 2 or 3 days of [installing the update](https://github.com/Microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) i.e the beta version with your findings.

- Warn if any edits to a Go file is saved when a debug session is in progress @haryps 
- Use the `latest` tag to install `gopls` instead of master branch @stamblerre, @banks
- Prompt to update `gopls` if you have an older version @stamblerre
- Tools are now installed in module mode @stamblerre
- `gopls` can now be used when you are using Go from the tip @stamblerre @Goodwine
- Allow alternate tools to be used in place of `gopls` @anurbol
- Show status bar for "Modules" when in module mode @mattetti 
- Debugging improvements @quoctruong, @BaronChen
     - Fix the bug where setting breakpoint fails if the remote program is started through dlv with --continue switch. Bug #2690
     - Fix the bug where disconnecting (after attaching to) the remote program terminates it. Bug #2592
    -  Fix the bug where setting breakpoint will fail if a breakpoint already exists (if dlv is started through multi client and another client sets the breakpoint).

stamblerre:
Regarding the automatic updates for gopls, I realized that I didn't add correct handling for people who use `gopls` at master by running `go get golang.org/x/tools/gopls@master`. I will send a PR for that later today.

quoctruong:
There is a bug on clearing breakpoints after detaching from the remote process. I'll send out a PR later today for that.

mattetti:
> Show status bar for "Modules" when in module mode

Confirmed to work as expected (status bar only shows up when go modules is on). But there is an important caveat: the status doesn't change unless extension/vs code is restarted. In other words, a user might change their go module settings in they env variables, that change won't be reflected in the editor. I can't really think of a good solution so I think we are good to go.

[...]

PR #2747, opened 2019-09-12
Clear pkg mod cache when env var for mod changed

ramya-rao-a:
Follow up from https://github.com/microsoft/vscode-go/issues/2738#issuecomment-530578727

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -9,7 +9,7 @@ import vscode = require('vscode');
 import * as goGenerateTests from './goGenerateTests';
 import { setGlobalState } from './stateUtils';
 import { updateGoPathGoRootFromConfig, installAllTools, offerToInstallTools, installTools, promptForMissingTool } from './goInstallTools';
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
@stamblerre This should be enough to update the modules status, right?

stamblerre:
These changes would need to be propagated to any server like `gopls` or `gocode`, but I think this is perfectly fine for now (I don't think `gopls` could handle a change to the module status, and if you are turning on and off modules, I think you can reasonably expect to reload your VSCode window).

Unrelated, but the name of the pkgModCache variable is a bit confusing/misleading, since Go modules has the concept of a module cache, and you would definitely not want to be clearing that :) Doesn't need to be changed or anything, but figured I'd just bring it up.

ramya-rao-a:
> if you are turning on and off modules, I think you can reasonably expect to reload your VSCode window

In that case, I would rather show a notification prompting for a reload than update the status which might mislead the user to assume that everything would work fine when they keep changing the value of GO111MODULE

stamblerre:
That seems reasonable. Another case I hadn't even considered is that, with `gocode`, you would have to go as far as killing `gocode-gomod` and starting up `gocode`, or vice versa.

Issue #763, opened 2017-02-02
debug - stopOnEntry not working


roblourens:
When I launch with `"stopOnEntry": true`, the debugger never pauses. Same on Mac and Windows. But I'm surprised that nobody has filed an issue on this. And there are even issues where people mention that they are using stopOnEntry successfully.

roblourens:
Debugging it, I see that we send an initial StoppedEvent, then vscode asks for the list of threads. We call ListGoroutines, Delve returns an empty list that we send back to vscode. Then that's it. It seems like Delve doesn't want to do anything unless we send a Continue command at startup.

dibalavs:
I have the same issue:
vscode: 1.82.2
delve debugger: 1.1.0
go extension: 0.6.93

ramya-rao-a:
Fixed with #2762 
The fix is now available in the latest version of the update (0.11.7)

PR #2762, opened 2019-09-19
fix stopOnEntry

jhendrixMSFT:
Send a dummy thread to vscode when stopped on entry as there are no
threads in the process yet.
Send the stop event with the same dummy thread ID.

### DIFF
--- a/package.json
+++ b/package.json
@@ -592,11 +592,6 @@
                 "description": "Indicates local or remote debugging.  Local maps to the dlv 'attach' command, remote maps to 'connect'.",
                 "default": "local"
               },
-              "stopOnEntry": {
-                "type": "boolean",
-                "description": "Automatically stop program after launch.",
-                "default": false
-              },
               "showLog": {
                 "type": "boolean",
                 "description": "Show log output from the delve debugger.",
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Microsoft/vscode-go/issues/763

ramya-rao-a:
cc @quoctruong

jhendrixMSFT:
Good question, I tried it and it doesn't work.  While I'm sure it can be fixed, I want to ask if we think this scenario is valuable?  If not then I suggest removing `stopOnEntry` from the `attach` option.
I personally feel like this isn't really useful and in my experience I don't recall other debuggers doing this.  However I might be overlooking some use case.

quoctruong:
Yes, I think by default, the program for remote attach scenario wouldn't start until a debugger is connected to it anyway. Therefore, this option may not be needed for that.

jhendrixMSFT:
@ramya-rao-a are you ok with removing `stopOnEntry` from `attach`?

ramya-rao-a:
@jhendrixMSFT Yes, I am fine with removing stopOnEntry for attach scenarios and ignore it for the launch request in debug mode as well

Issue #2768, opened 2019-09-20
language server(gopls) not working


1yefuwang1:
After setting `go.useLanguageServer` to `true`, this plugin basically stops working: no auto-completion, no go-to-definition, no quick-info-on-hover.
When I try to restart the language server, the following message shows up.
![image](https://user-images.githubusercontent.com/18023393/65328293-929a6b80-dbe8-11e9-8d7c-2463dd467544.png)

Detailed info of my vscode:
Version: 1.38.1
Commit: b37e54c98e1a74ba89e03073e5a3761284e3ffb0
Date: 2019-09-11T13:31:32.854Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Darwin x64 18.7.0

Version of the this plugin I am using: 0.11.5

Version of gopls: 
golang.org/x/tools/gopls v0.1.7
    golang.org/x/tools/gopls@v0.1.7 h1:YwKf8t9h69++qCtVmc2q6fVuetFXmmu9LKoPMYLZid4=

1yefuwang1:
seems like the same issue as #2767

enix223:
got the same issue, after upgrading vscode-go to `0.11.5`, `gopls` daemon does not start when vscode started...

liuhangyu:
I also encountered the same problem, unable to pull up the gopls version at 0.11.5, 0.11.4 is normal

ramya-rao-a:
Thanks for reporting everyone

Can one of you check the logs and share any errors that you see there?
- Run the command `Developer: Toggle Developer Tools`. 
- Share any errors from here related to the go extension.
- If there is too much noise here, then reload the window, and you should be able to see errors if nay

Another place to check is the logs from the Extension Host
- Select `Output` from the `View` menu
- In the output panel that opens, select `Log (extension host)` from the drop down on the top right corner.

@stamblerre has found a potential reason for the language server to not load and has a fix for that. Please give that a try as well via the [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) of this extension

enix223:
* Log from Developer tools for version `0.11.5`:

```
[Extension Host] Language Server Config: [object Object]
console.ts:137 [Extension Host] Language Server Path: /Users/enix/go/bin/gopls
console.ts:137 [Extension Host] Language server tool name: gopls
[...]

Issue #2767, opened 2019-09-20
Go tools (gopls) stop working via Remote SSH after update to 0.11.5


halega:
I'm using "Remote - SSH" plugin to work on a Go project. I have Windows host and VirtualBox with Ubuntu VM.

Steps to Reproduce:

1. Open VSCode on Windows, install Remote - SSH plugin and connect to Ubuntu.
2. Do remote install of Go 0.11.5 and install all Go tools (installation completed sucessfully).
3. Open any *.go file.

**Expected**
gopls started on Ubuntu machine and VSCode language features work as expected.

**Actual**
gopls not started and nothing works: no tooltips, no completion, etc.

**Comments**
All works with local projects on Windows.
vscode-go 0.11.4 worked in Remote-SSH.

Version: 1.38.1 (user setup)
Commit: b37e54c98e1a74ba89e03073e5a3761284e3ffb0
Date: 2019-09-11T13:35:15.005Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Windows_NT x64 10.0.18362

gopls version
golang.org/x/tools/gopls v0.1.7
    golang.org/x/tools/gopls@v0.1.7 h1:YwKf8t9h69++qCtVmc2q6fVuetFXmmu9LKoPMYLZid4=

golang-vo version 0.11.5

halega:
Reverted remote vscode-go to 0.11.4. All works again.

stamblerre:
Thank you for the issue report. @1yefuwang1 and @enix223, I'll continue the discussion of https://github.com/microsoft/vscode-go/issues/2768 here. Can you add a bit more information about your remote machines?

@ramya-rao-a: Are you aware of any tools for debugging extensions running remotely? My best idea right now is that I can add some more debug logging to my version of the extension, package a new .vsix file, and ask someone to install it remotely and inspect the logs.

stamblerre:
[...]

PR #2773, opened 2019-09-21
Gate the check for gopls updates using a setting

ramya-rao-a:
As seen in #2768 and #2767, the implicit calls to https://proxy.golang.org/golang.org/x/tools/gopls/@v/list can fail in multiple scenarios.

In VS Code, we also want to follow the guideline of not making any outgoing calls over the network unless it is done as part of a feature resulting from an explicit user action which the user is aware of.

Exceptions to this rule should be documented and a setting should be provided for the user to opt in/out.

Keeping this in mind, the recent feature of making calls to https://proxy.golang.org/golang.org is being updated as follows in this PR
- Provide a setting to opt in to the automatic update check. This is similar to the `extensions.autoCheckUpdates` setting in VS Code
- If the above is not enabled, then hard-code the latest known latest version of `gopls` in the extension

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -56,7 +56,7 @@ interface LanguageServerConfig {
 		implementation: boolean;
 		documentLink: boolean;
 	};
-	checkForUpdates: boolean
+	checkForUpdates: boolean;
 }
 
 // registerLanguageFeatures registers providers for all the language features.
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -56,6 +56,7 @@ interface LanguageServerConfig {
 		implementation: boolean;
 		documentLink: boolean;
 	};
+	checkForUpdates: boolean
 }
 
 // registerLanguageFeatures registers providers for all the language features.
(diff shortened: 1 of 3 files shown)


ramya-rao-a:
cc @stamblerre

src/goLanguageServer.ts
```diff
@@ -79,7 +80,7 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
 
 	// The user may not have the most up-to-date version of the language server.
 	const tool = getTool(toolName);
-	const update = await shouldUpdateLanguageServer(tool, path);
+	const update = await shouldUpdateLanguageServer(tool, path, config.checkForUpdates);
```
stamblerre: Maybe just:

```typescript
if config.checkForUpdates {
    const update = await shouldUpdateLanguageServer(tool, path);
    if (update) {
        promptForUpdatingTool(toolName); 
    }
}
```

ramya-rao-a: I'd still want to check against the version that is hard-coded in the extension if the setting is disabled.
How about renaming the setting to `go.checkGoProxyForToolUpdates`?
For now, we can use it only for `gopls`. 
In the future the same approach can be taken for other tools.

PR #2774, opened 2019-09-21
use the user's GOPROXY settings when requesting the gopls version

stamblerre:
I think this change, in conjunction with the setting to disable `gopls` automatic updates might be enough to solve the issue with making requests directly.


stamblerre:
After some discussion with teammates, this is the correct approach for the `gopls` automatic updates. `$GOPROXY` wasn't on by default when I made the original PR, which is why I had used https://proxy.golang.org directly. With Go 1.13, `GOPROXY` is on by default, so this should be fine. It will be used when running the `go` command anyway, so I think it's fine to have this as opt-out with a setting.

src/goLanguageServer.ts
```diff
@@ -589,3 +572,53 @@ async function goplsVersion(goplsPath: string): Promise<string> {
...
+			});
+		} catch (e) {
+			return null;
+		}
+		return data;
+	}
+	return null;
+}
+
+async function goProxy(): Promise<string[]>  {
```
ramya-rao-a: We plan to do this not more than once per each VS Code session right?
If so, then [on start up](https://github.com/microsoft/vscode-go/blob/0.11.7/src/goInstallTools.ts#L330), we do run `go env GOPATH GOROOT`.
We can append GOPROXY to that and have the value stored in `process.env['GOPROXY']` and skip the below.

ramya-rao-a: I believe re-using that `go env` call is a good move. If you agree, you can cherry pick [this commit of mine](https://github.com/ramya-rao-a/vscode-go/commit/1d7700a1ef30a9b1c6fd5f15cccfbb921103267c)

stamblerre: Done, thanks!

PR #2775, opened 2019-09-21
Hard-code latest version info for gopls

ramya-rao-a:
Hard-coding latest version info for gopls for now while we explore if using go proxy as the source of truth for the `gopls` version info is the right approach or not

cc @stamblerre


stamblerre:
This looks good to me.

Issue #2770, opened 2019-09-20
Error: TypeError: Invalid Version: undefined


ransombriggs:
### Steps to Reproduce:
1. Ubuntu
```
-> % cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.6 LTS"
```
2. Visual Studio
```
-> % code --version      
1.38.1
b37e54c98e1a74ba89e03073e5a3761284e3ffb0
x64
```
3. Go
```
-> % go version
go version go1.12.9 linux/amd64
```
4. main.go
```
-> % cat main.go 
package main

func main() {

}
```
5. Save main.go
![image](https://user-images.githubusercontent.com/62452/65357253-230e8700-dbbc-11e9-94b9-9566a686abc4.png)

### Debugging
I found that if I turned off `buildOnSave` and `vetOnSave` that the error goes away if that helps at all with debugging.
```
{
    "go.buildOnSave": "off",
    "go.vetOnSave": "off"
}

ramya-rao-a:
Are you able to see this consistently? If yes, then can you share the stack trace for this error? You can find this in one of 2 places
- The logs for the extension host
     - Select `Output` from the `View` menu, and then select `Log (Extension Host)` from the drop down on the top right corner of the output panel
- The dev tools for VS Code
      - Run the command `Developer: Toggle Developer Tools`


In case there is too much noise in these logs, you can clear them and do what is needed to replicate the issue to see the corresponding logs

mrpanc:
Same error. "Error: TypeError: Invalid Version: undefined" is the only thing i've got. There is no other log or more information.

mrpanc:
It happens after upgrading to the latest version.

ramya-rao-a:
@mrpanc Did you try the output panel or the console as suggested in https://github.com/microsoft/vscode-go/issues/2770#issuecomment-533922924 for more information?

ramya-rao-a:
@mrpanc Are you also able to get around the error by disabling the build on save and vet on save feature like @ransombriggs ?

@mrpanc, @ransombriggs Can you share the version of Go you are using? Also share any Go related settings you might have added/edited. You can get these by running the command `Preferences: Open Settings (JSON)`

[...]

PR #2782, opened 2019-09-23
Additional error handling and testing for getGoVersion

stamblerre:
I don't think this will actually fix https://github.com/microsoft/vscode-go/issues/2770, but this is at least a step in identifying the problem.


stamblerre:
Makes sense. Based on the user reports, looks like `go version` command is failing in some way. I'll close this PR.

PR #2789, opened 2019-09-25
Remove configurations for language server features

stamblerre:
A few changes here:

- Clear the diagnostics when the language server has been restarted to avoid lingering errors.
- Remove the middleware. `gopls` doesn't yet watch configuration changes, so we suggest that users reload the server if they change a setting. Furthermore, it is the incorrect approach for two providers to be registered and for the extension to ignore the results of one. We don't want the language server to be doing redundant work.


ramya-rao-a:
The purpose of the middleware was not to react to config changes.
We have a listener set up to prompt users to reload if there was a change in any of the language server related settings.

The purpose of the middleware is to ignore the results from the language server for a features that the user doesnt want to use the language server for.

The purpose of the second part of the code that is being deleted is to register providers for features that the language server doesnt support yet or for features that the user has chosen not to get from the language server

With this PR, the user can no longer pick and choose which features they want to enable from the language server resulting in an all or nothing scenario.

ramya-rao-a:
The all or nothing scenario is fine by me once `gopls` comes out of the alpha/beta stage.

stamblerre:
If you would be ok with it, I would propose that we remove this configurability now. I'm worried that people are just picking and choosing the features that they like rather than filing issues, so we won't actually be aware of missing features until we release v1.0, which will be too late. What do you think?

ramya-rao-a:
We are encouraging people to use the language server, especially if they are using modules. If they have been using the configuration to pick and choose because they have found some of the language server features to not be very performant, forcing them to now use the language server as a whole might result in them disabling it altogether.

v1.0 is not the end of journey, we can always add more feature and fix bugs. So I dont believe it will be too late.

stamblerre:
The problem with this approach is that we have no way of knowing which features people are having issues with. I think people are less likely to disable the server altogether than they are to file an issue (if they have been using it), and since the server is still opt-in, it is more reasonable for us to make large configuration changes now rather than later.

v1.0 is the point at which we will begin recommending the language server as the default experience for modules, and I would prefer that we become aware of as many bugs as possible before then.

ramya-rao-a:
The all or nothing decision might need some more thoughts, can you create a separate PR for the clearing diagnostics on gopls restart part so that we can have that merged and use this PR to continue the discussions on middleware?

stamblerre:
Done (https://github.com/microsoft/vscode-go/pull/2817). I will modify this PR to remove all of the configurations.

ramya-rao-a:
Hey @stamblerre

An update on the usage of the `go.languageserverexperimentalfeatures` setting. Below are the top 5 features that are getting disabled by users
- format
- documentLink
- diagnostics
- autocomplete
- findReferences

Given that at any point, the auto complete and find references features are better supported in gopls than individual tools, I propose that for the next update of the extension, we keep only the below settings and remove the rest
- format
- documentLink
- diagnostics

stamblerre:
Thank you for letting me know! I agree with this approach.

ramya-rao-a:
Great!

Please take a look at https://github.com/microsoft/vscode-go/commit/63f1b040793604dfc75fce3c211fc92411ea234d

PR #2791, opened 2019-09-25
Refactor some of the installation code that calls the go command

stamblerre:
I'm hoping to refactor some of the code that calls the `go` command, since it gets called in many different places and there's a fair amount of duplication. This is the first of the changes that will do this refactoring.

### DIFF
--- a/src/goCommand.ts
+++ b/src/goCommand.ts
@@ -1,6 +1,6 @@
 import util = require('util');
 import cp = require('child_process');
-import { getBinPath } from "./util";
+import { getBinPath } from './util';
 
 const execFile = util.promisify(cp.execFile);
 
@@ -33,7 +33,7 @@ async function goCommand(cmd: string, path: string, flags: string[], opts: any):
 	let err: Error;
 
 	try {
-		let result = await execFile(goRuntimePath, args, opts);
+		const result = await execFile(goRuntimePath, args, opts);
 		stdout = result.stdout.toString();
 		stderr = result.stderr.toString();
 	} catch (e) {
--- a/src/goCommand.ts
+++ b/src/goCommand.ts
@@ -19,12 +19,18 @@ export async function goBuild(path: string, flags: string[], opts: any): Promise
 async function goCommand(cmd: string, path: string, flags: string[], opts: any): Promise<GoCommandResult> {
 	const goRuntimePath = getBinPath('go');
 	if (!goRuntimePath) {
-		return null;
+		return new Promise<GoCommandResult>(resolve => {
+			resolve({
+				stdout: '',
+				stderr: '',
+				err: new Error('failed to determine path to `go` binary'),
+			})
+		});
 	}
 
 	const args = [cmd, ...flags, path];
 
-	return new Promise<GoCommandResult>((resolve) => {
+	return new Promise<GoCommandResult>(resolve => {
 		cp.execFile(goRuntimePath, args, opts, (err, stdout, stderr) => {
 			resolve({
 				stdout: stdout.toString(),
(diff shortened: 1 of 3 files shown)


src/goGetPackage.ts
```diff
@@ -23,24 +22,17 @@ export function goGetPackage() {
...
-			outputChannel.show();
-			outputChannel.clear();
-			outputChannel.appendLine(stderr);
-			buildCode();
-			return;
-		}
-
-		// go get -v doesn't write anything when the package already exists
+	// go get -v doesn't write anything when the package already exists
+	if (result.stderr === '') {
```
ramya-rao-a: What is the reason to flip the `if` condition here?
If package already exists, there is no need to build the code or show the output channel

src/goCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+
+async function goCommand(cmd: string, path: string, flags: string[], opts: any): Promise<GoCommandResult> {
+	const goRuntimePath = getBinPath('go');
+	if (!goRuntimePath) {
+		return null;
+	}
+
+	let args: string[] = [cmd];
+	args = args.concat(flags);
+	args.push(path);
```
ramya-rao-a: The above can be simplified to
`const args = [cmd, ...flags, path];`

src/goCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+	let err: Error;
+
+	try {
+		const result = await execFile(goRuntimePath, args, opts);
+		stdout = result.stdout.toString();
+		stderr = result.stderr.toString();
+	} catch (e) {
+		err = e;
+	}
+	return { stdout: stdout, stderr: stderr, err: err };
```
ramya-rao-a: If the callers of `goGet` and `goBuild` have to deal with `stderr`, `stdout` and `err` specifically, then I dont see the benefit of promisifying the call to cp.execFile

What happens to the case where `execFile` throws an error, but there is also information available in `stderr`?

src/goGetPackage.ts
stamblerre: Fixed to add a return. I flipped it because I always find it easier to understand code when the uncommon case is in the if conditions.

src/goCommand.ts
stamblerre: Done.

src/goCommand.ts
stamblerre: Makes sense. I've been trying to switch more of the code to use async functions because I find them easier to reason about and understand, but I see here that it's simpler to just use the callback.

src/goGetPackage.ts
```diff
@@ -23,24 +22,17 @@ export function goGetPackage() {
 		return;
 	}
 
-	const goRuntimePath = getBinPath('go');
-	if (!goRuntimePath) {
-		return vscode.window.showErrorMessage(`Failed to run "go get" to get package as the "go" binary cannot be found in either GOROOT(${process.env['GOROOT']}) or PATH(${envPath})`);
-	}
-
 	const env = Object.assign({}, process.env, { GOPATH: getCurrentGoPath() });
+	const result = await goGet(importPath, ['-v'], { env });
```
ramya-rao-a: `result` can be `null` when path to go binary is not found.

ramya-rao-a:
There is a package [execa](https://www.npmjs.com/package/execa) that I have been considering to use in place of the child process.

It gives "promisifies" the calls and also provides a neat way to handle outputs.
Any thoughts on using that?

src/goGetPackage.ts
stamblerre: Added error handling for this.

stamblerre:
I'm not sure that I'm familiar enough with the existing libraries to make a decision on if this additional dependency is worth it, but if you feel there would be significant benefit, I am happy to make the switch.

Issue #2115, opened 2018-11-11
go.testFlags are not applied correctly when using the "debug test" code lens


segevfiner:
Steps to Reproduce:

1. Save:

`foo_test.go`:
```go
package foo

import (
	"flag"
	"testing"
)

var ok = flag.Bool("ok", false, "")

func TestFoo(t *testing.T) {
	if !*ok {
		t.Fail()
	}
}
```

`.vscode/settings.json`:
```json
{
    "go.testFlags": ["-ok"]
}

```
2. Open `foo_test.go` and click the "run test" code lens, the test will pass.
3. Click the "debug test" code lens, the test will fail.

ramya-rao-a:
I believe this is because the `run test` codelens runs the test as `go test -run ^TestFoo$ -ok` whereas the debug codelens starts delve and passes `-ok` as value for the `--build-flags` argument.

The debug test codelens always had its limitations, especially when the test relied on arguments being passed to it. This is such a case and the workaround is to fallback to using a debug configuration in the launch.json file and pass `-ok` as part of the `args` property.

rubenzinho07:
😈

Em Ter, 13 de nov de 2018 03:02, Ramya Rao <notifications@github.com>
escreveu:

segevfiner:
From `go test --help`
```
The 'go test' command rewrites or removes recognized flags,
as appropriate, both before and after the optional package list,
before invoking the test binary.

For instance, the command

	go test -v -myflag testdata -cpuprofile=prof.out -x

will compile the test binary and then run it as

	pkg.test -test.v -myflag testdata -test.cpuprofile=prof.out
```

And in code [go/src/cmd/go/internal/test/testflag.go#testFlags](https://github.com/golang/go/blob/ca33f33b1420cf333c59c6458bb9bc8910c91ecb/src/cmd/go/internal/test/testflag.go#L90).

I guess either us or `dlv` will need to pickup this logic to properly separate and rewrite the test arguments as done when running `go test`. Or otherwise separate the test build/test run flags in our configuration and pass them appropriately, which feels less "clean".

brycekahle:
> The debug test codelens always had its limitations, especially when the test relied on arguments being passed to it. This is such a case and the workaround is to fallback to using a debug configuration in the launch.json file and pass `-ok` as part of the `args` property.

@ramya-rao-a  Could we add a new config value `go.testArgs` that is specifically for `args` during tests? That would solve this particular issue.

[...]

PR #2795, opened 2019-09-26
Fix testFlags not being applied correctly to debug CodeLens

quoctruong:
Fix a bug where the testFlags are not being applied correctly to the debug CodeLens (https://github.com/Microsoft/vscode-go/issues/2115)


ramya-rao-a:
@quoctruong With this solution, all flags get passed to delve as arguments and the `--build-flags` is ignored. Have you tested this change both with known and custom flags?

As an extra fyi, #1248 was related to how we pass build flags

ramya-rao-a:
@quoctruong There is more at play here including supporting the command `Go: Debug test at cursor`. I have pushed a few fixes to refactor the code around here which should help.

The commit that would fix the issue here is https://github.com/microsoft/vscode-go/commit/9ab7b8bff49c38830ae4625718f605ba73fec0dc

PR #2799, opened 2019-09-30
Show correct placeholder when extracting method

romanlevin:
Currently the placeholder text when extracting a function or a variable is

> Please enter a name for the extracted variable.

With this change, in the case of extracting a function, the placeholder text will be

> Please enter a name for the extracted method.


romanlevin:
@ramya-rao-a thanks!

ramya-rao-a:
The latest update to this extension (0.11.8) has this fix in.

Thanks @romanlevin and Happy Coding!

PR #2802, opened 2019-10-01
Fix and reenable linux in travis runs

### LINKED ISSUE #2779: Not able to run tests on Linux in Travis CI
June 4th, 2019 was the last time there was a successful run in Travis on Linux.
There has been no code change in the extension that is causing the failures.
This is most likely due to some changes upstream.

For now, I have [disabled the Linux runs](https://github.com/microsoft/vscode-go/commit/551b7cc6bd35905575573d948732a0bec715473e) on travis to reduce noise in PRs
This issue is to track the root cause for this issue, fix it and enable the Linux tests again

ToothlessGear:
After travis-ci moved their default linux os from trusty to xenial, `xvfb` has to be started differently.

Closes #2779

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,13 @@ sudo: false
 
 os:
   - osx
-  # - linux
+  - linux
 
 before_install:
   - if [ $TRAVIS_OS_NAME == "linux" ]; then
     export DISPLAY=:99.0;
-    sh -e /etc/init.d/xvfb start;
+    Xvfb :99 &
+    sleep 3; # give xvfb some time to start 
     sudo apt-get update && sudo apt-get install -y libsecret-1-0;
     fi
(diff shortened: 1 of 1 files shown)


ToothlessGear:
Per [documentation](https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-directly), to give Xvfb some time to start up. 
It's probably not needed, but better be safe than sorry. 😄

ramya-rao-a:
lol, can we please add that as a comment next to the sleep command?

ToothlessGear:
Sure, done!

ToothlessGear:
@ramya-rao-a I just saw, that the inline comment broke the yaml parsing for the entire file.
We'll have to move it outside.

PR #2803, opened 2019-10-02
Sort import packages by std lib first

### LINKED ISSUE #2683: Add Import could prioritize stdlib packages
Currently the "Go: Add Import" command doesn't prioritize core packages, so you often get a bunch of random packages over-powering the standard ones.  Here's a few examples:

Screen Shot 2019-08-03 at 11 10 51 AM

Screen Shot 2019-08-03 at 11 15 24 AM

It wasn't a big deal when autocompletion of core packages worked without imports, but I can't seem to get that working anymore `go.autocompleteUnimportedPackages` enabled — maybe a gopls issue?

karthikraobr:
This PR partially resolves #2683. The list is sorted by std lib packages first when there is no user input. But this ordering doesn't work when the user starts entering characters to the import box.
@ramya-rao-a any pointers on how to fix this after the user starts entering characters?


src/goImport.ts
```diff
@@ -20,10 +20,18 @@ export async function listPackages(excludeImportedPkgs: boolean = false): Promis
 		? await getImports(vscode.window.activeTextEditor.document)
 		: [];
 	const pkgMap = await getImportablePackages(vscode.window.activeTextEditor.document.fileName, true);
-
+	const stdLib = await getStandardLibraryPackages();
```
ramya-rao-a: Instead of making another call to `go list` which we know to be expensive, why not update `getImportablePackages`  itself return the information on whether a package is standard or not?

Right now, we use `{{.ImportPath}}` and `{{.Dir}}` when calling `go list` in the `goPackages.ts` file.
We can add `{{.Standard}}` to the same call and get a boolean representing what we need

src/goImport.ts
```diff
@@ -20,10 +20,18 @@ export async function listPackages(excludeImportedPkgs: boolean = false): Promis
...
 	return Array.from(pkgMap.keys())
 		.filter(pkg => !importedPkgs.some(imported => imported === pkg))
-		.sort();
+		.sort((a,b)=> {
+			if (stdLib.get(a)){
+				return -1
+			}
+			if (stdLib.get(b)){
+				return 1
+			}
```
ramya-rao-a: This sorting will only ensure that the standard libraries appear before the others.
But among the standard libraries themselves, we lose the sorting
Similarly, we also lose the sorting among the non standard libraries

src/goImport.ts
karthikraobr: If I am not wrong I guess we are using [gopkgs](https://github.com/uudashr/gopkgs) to list the packages and that doesn't support the `{{.Standard}}` flag. That's the reason I wrote another function that uses `go list`.

ramya-rao-a: Ah! fair point.

In that case, I would prefer if we can update `gopkgs` to also support `{{.Standard}}`
I am guessing that whatever it does under the hood to fetch the packages, it already knows if a package is standard or not and exposing this info is a valid feature request.

Elsewhere in this project in [goSuggest.ts](https://github.com/microsoft/vscode-go/blob/0.11.7/src/goSuggest.ts#L552-L553) file, a use a very crude way to identify the standard packages which is seeing if the import path contains a ".". 

So, maybe until `gopkgs` is fixed, we can use a similar approach?

karthikraobr: I will create a PR to support`{{.Standard}}` in `gopkgs` and then continue with this PR. That seems to be the cleanest approach right?

karthikraobr: @ramya-rao-a  fyi uudashr/gopkgs#17

ramya-rao-a: I see that you have used GOROOT in the Dir to determine if the package is standard or not.
We do have access to the {{.Dir}}, so maybe we can use that in the extension, compare with `process.env['GOROOT']`?

src/goPackages.ts
```diff
@@ -81,13 +80,23 @@ function gopkgs(workDir?: string): Promise<Map<string, string>> {
...
+					otherPkg.set(pkgPath, pkgName);
+				}
 			});
-
+			Array.from(stdPkgs.keys()).forEach((val)=>{
+				pkgs.set(val, stdPkgs.get(val));
+			})
+			Array.from(otherPkg.keys()).forEach((val)=>{
+				pkgs.set(val, otherPkg.get(val));
+			})
```
karthikraobr: @ramya-rao-a This looks very crude to me :) esp maintaining two maps and merging them. Could you suggest a way to improve it? Thanks!

ramya-rao-a: Instead of maintaining 2 different maps, we can go back to the single map as before.
That map uses string -> string at the moment
We can change that to string -> PackageInfo where PackageInfo is an interface
```
export interface PackageInfo {
   name string;
   isStd boolean;
}
```

ramya-rao-a: This makes it extendable such that in the future if we want to extract more info from the `gopkgs` tool, we can do so.

src/goPackages.ts
```diff
@@ -75,17 +81,21 @@ function gopkgs(workDir?: string): Promise<Map<string, string>> {
 					}
 					const index = pkgPath.lastIndexOf('/');
 					const pkgName = index === -1 ? pkgPath : pkgPath.substr(index + 1);
-					pkgs.set(pkgPath, pkgName);
+					pkgs.set(pkgPath, {
+						name: pkgName,
+						isStd: pkgPath.includes(process.env['GOROOT'])
```
ramya-rao-a: perhaps .startsWith is a better function to use here. We expect the path to start with the goroot right?

```suggestion
						isStd: pkgPath.startsWith(process.env['GOROOT'])
```

Also, do we want to take care of the case where `process.env['GOROOT']` doesn't have any value set? It shouldn't happen so, but if it does, do we might want to default to false here right?

src/goPackages.ts
```diff
@@ -75,17 +81,21 @@ function gopkgs(workDir?: string): Promise<Map<string, string>> {
...
 			}
-
 			output.split('\n').forEach((pkgDetail) => {
 				if (!pkgDetail || !pkgDetail.trim() || pkgDetail.indexOf(';') === -1) return;
-				const [pkgName, pkgPath] = pkgDetail.trim().split(';');
-				pkgs.set(pkgPath, pkgName);
+				const [pkgName, pkgPath, pkgDir] = pkgDetail.trim().split(';');
+				pkgs.set(pkgPath, {
+					name: pkgName,
+					isStd: pkgDir.includes(process.env['GOROOT'])
```
ramya-rao-a: perhaps .startsWith is a better function to use here. We expect the path to start with the goroot right?

```suggestion
						isStd: pkgPath.startsWith(process.env['GOROOT'])
```

Also, do we want to take care of the case where `process.env['GOROOT']` doesn't have any value set? It shouldn't happen so, but if it does, do we might want to default to false here right?

src/goPackages.ts
ramya-rao-a: This is the case of the old `gopkgs`. In that case, we will have to treat the output differently right?

karthikraobr: I am not entirely sure how the output of `{{.Dir}}` would look with an earlier version?

ramya-rao-a: https://github.com/tpng/gopkgs is the older tool. Can you give it a try?

ramya-rao-a: It most likely ignores the entire formatting input

karthikraobr: Oh sorry. I forgot to try this. Will do it tomorrow.

src/goPackages.ts
```diff
@@ -75,19 +81,23 @@ function gopkgs(workDir?: string): Promise<Map<string, string>> {
 					}
 					const index = pkgPath.lastIndexOf('/');
 					const pkgName = index === -1 ? pkgPath : pkgPath.substr(index + 1);
-					pkgs.set(pkgPath, pkgName);
+					pkgs.set(pkgPath, {
+						name: pkgName,
+						isStd: goroot === null ? false : pkgPath.startsWith(goroot)
```
ramya-rao-a: `pkgPath` is not the absolute path. It is the import path. So `pkgPath.startsWith(goroot)` here for the old gopkgs tool will always return false.

So, we might as well set this to `false` directly.

I'll soon retire this part of the code as we don't expect anyone to be using the old `gopkgs` anymore

ramya-rao-a: Or another option is to use this very crude logic I currently use for the auto completion widget

```suggestion
						isStd: !pkgPath.includes('.')
```

ramya-rao-a:
I pushed a few commits to this PR @karthikraobr 
- Simplify the sort logic. Your approach required accessing the map for every 2 entries that gets compared in the sort() algorithm. So pulled that part out.
- Tweaked the case of the old gopkgs scenario to use a crude way of figuring out standard packages that I use elsewhere in `goSuggest.ts` file
- Fixed linting errors

PR #2812, opened 2019-10-05
support Go 1.11 when installing tools in module mode

stamblerre:
Go 1.11 still requires a `go.mod` file to download in module mode. Support it by adding a fake `go.mod` file to the temporary directory for tool installation.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -129,6 +129,11 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 	if (!fs.existsSync(toolsTmpDir)) {
 		fs.mkdirSync(toolsTmpDir);
 	}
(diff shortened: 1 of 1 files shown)


src/goInstallTools.ts
```diff
@@ -129,6 +129,11 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 	if (!fs.existsSync(toolsTmpDir)) {
 		fs.mkdirSync(toolsTmpDir);
 	}
+	// Write a temporary go.mod file to support Go 1.11.
```
ramya-rao-a: If this is needed only for 1.11 and not newer versions, should we perhaps add a check for the Go version?

stamblerre: It shouldn't be a problem for the later Go versions, but I added a check anyway.

PR #2814, opened 2019-10-07
Move comment outside multiline command in .travis.yml

ToothlessGear:
Yaml doesn't support inline comments on multiline strings (#2802), unfortunately. 
Travis can't parse the file correctly and stops running builds. 
I saw it too late, after #2802 already got merged.
This PR should amend this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,11 @@ os:
   - linux
 
 before_install:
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Ah! This was the reason why travis wasnt running any tests when I was taking a look over the PRs in the weekend.

Thanks for looking out!

Issue #2806, opened 2019-10-02
Tests using go doc fail on Go tip due to difference in new lines


ramya-rao-a:
The below tests are failing when used on Go tip due to new line differences in the actual and expected output.
- Test SignatureHelp Provider using godoc
- Test Hover Provider using godoc
- Test Hover Provider using gogetdoc

These tests on previous versions of Go run fine.
One way to make the tests more resilient is to trim the text being compared, so that the starting and ending newlines or spaces dont result in tests failing.

PRs are most welcome
Code pointers:
- [Set up local dev env](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code)
- Find the above tests in the extension.test.ts file
- Update the `test('Test` to `test.only('Test ` so that you can focus only on the single test you care about
- Use the `Launch extension tests` to start debugging the tests

ramya-rao-a:
@stamblerre Turns out this is due to the difference in behavior of the `go doc` command in Go 1.13

For example, using one of the [test cases](https://github.com/microsoft/vscode-go/blob/master/test/fixtures/gogetdocTestData/test.go#L24), the command `go doc -c -cmd -u github.com/golang/example/hello print` gives different output

Go 1.12

<img width="917" alt="Screen Shot 2019-10-20 at 11 14 01 AM" src="https://user-images.githubusercontent.com/16890566/67163998-bdccc380-f32a-11e9-8cf9-4512de2c16e8.png">

Go 1.13

<img width="833" alt="Screen Shot 2019-10-20 at 10 35 03 AM" src="https://user-images.githubusercontent.com/16890566/67164001-c2917780-f32a-11e9-9a78-4941bc7b0b94.png">

As you can see, in Go 1.13 the package statement is also included in the output.

Do you have any insights around this?

ramya-rao-a:
Fixed with https://github.com/microsoft/vscode-go/commit/0839658c8d2dd95969d5463da9877dfa716d5fc6 for the moment.

PR #2818, opened 2019-10-07
Add Go 1.13 to Travis configuration

stamblerre:


ramya-rao-a:
@stamblerre Would you consider fixing #2806 as part of this PR? It applies to both tip and 1.13

stamblerre:
Sure thing. I'll comment here when it has been fixed.

ramya-rao-a:
@stamblerre Was the refactoring done in https://github.com/microsoft/vscode-go/pull/2818/commits/ecf0dbcd03a05de3fe3fb3445995011b7f9397a6 meant to be part of this PR which is about enabling tests for Go 1.13?

stamblerre:
Yes, I've found that when I run the tests locally, it picks up my VSCode configurations, so I am trying to make sure the tests run with the default settings, rather than those of the user.

src/util.ts
```diff
@@ -126,7 +126,11 @@ let telemtryReporter: TelemetryReporter;
 let toolsGopath: string;
 
 export function getGoConfig(): vscode.WorkspaceConfiguration {
-	return vscode.workspace.getConfiguration('go', vscode.window.activeTextEditor ? vscode.window.activeTextEditor.document.uri : null);
+	return vscode.window.activeTextEditor ? getGoConfigForUri(vscode.window.activeTextEditor.document.uri) : null;
```
ramya-rao-a: The code here has been tweaked to return null when there is no active text editor. Is this intentional? We need a valid go config to be returned at all times.

src/util.ts
```diff
@@ -126,7 +126,11 @@ let telemtryReporter: TelemetryReporter;
...
+}
+
+export function getGoConfigForUri(uri: vscode.Uri): vscode.WorkspaceConfiguration {
```
ramya-rao-a: Instead of a separate function that takes uri, you can tweak the original utility to take an optional param

```javascript
export function getGoConfig(uri?: vscode.Uri): vscode.WorkspaceConfiguration {
     if (vscode.window.activeTextEditor) {
        uri = vscode.window.activeTextEditor.document.uri;
     }
    return vscode.workspace.getConfiguration('go', uri);
}
```

test/integration/extension.test.ts
```diff
@@ -146,21 +150,20 @@ suite('Go Extension Tests', () => {
...
 
-	test('Test Definition Provider using godoc', (done) => {
-		const config = Object.create(vscode.workspace.getConfiguration('go'), {
-			'docsTool': { value: 'godoc' }
-		});
-		testDefinitionProvider(config).then(() => done(), done);
+	test('Test Definition Provider using godoc', async (done) => {
+		const config = getDefaultConfig();
+		config['docsTool'] = 'godoc';
+		testDefinitionProvider(config);
```
ramya-rao-a:
```suggestion
		await testDefinitionProvider(config);
```

src/goCover.ts
```diff
@@ -239,7 +238,7 @@ export function applyCodeCoverage(editor: vscode.TextEditor) {
 		return;
 	}
 
-	const cfg = vscode.workspace.getConfiguration('go', editor.document.uri);
+	const cfg = getGoConfig();
```
ramya-rao-a:
```suggestion
	const cfg = getGoConfig(editor.document.uri));
```

src/goDebugConfiguration.ts
```diff
@@ -69,7 +69,7 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 			debugConfiguration['env']['GOPATH'] = gopath;
 		}
 
-		const goConfig = vscode.workspace.getConfiguration('go', folder ? folder.uri : null);
+		const goConfig = folder ? getGoConfig(folder.uri) : null;
```
ramya-rao-a:
```suggestion
		const goConfig = getGoConfig(folder && folder.uri);
```

src/goExtraInfo.ts
```diff
@@ -36,7 +37,7 @@ export class GoHoverProvider implements HoverProvider {
 			const hoverTexts = new vscode.MarkdownString();
 			hoverTexts.appendCodeblock(text, 'go');
 			if (definitionInfo.doc != null) {
-				hoverTexts.appendMarkdown(definitionInfo.doc);
+				hoverTexts.appendMarkdown(definitionInfo.doc.trimLeft());
```
ramya-rao-a: Why is this change part of this PR?

src/goSignature.ts
```diff
@@ -50,7 +50,7 @@ export class GoSignatureHelpProvider implements SignatureHelpProvider {
 				const sigStart = nameEnd + 5; // ' func'
 				const funcName = declarationText.substring(0, nameEnd);
 				sig = declarationText.substring(sigStart);
-				si = new SignatureInformation(funcName + sig, res.doc);
+				si = new SignatureInformation(funcName + sig, res.doc.trimLeft());
```
ramya-rao-a: Is this part of a separate bug fix?

src/goSuggest.ts
```diff
@@ -97,7 +97,7 @@ export class GoCompletionItemProvider implements vscode.CompletionItemProvider,
 		}
 
 		return runGodoc(path.dirname(item.fileName), item.package || path.dirname(item.fileName), item.receiver, item.label, token).then(doc => {
-			item.documentation = new vscode.MarkdownString(doc);
+			item.documentation = new vscode.MarkdownString(doc.trimLeft());
```
ramya-rao-a: Same question here, can we have all the doc trim fixes as part of a separate fixes.
Helps a lot with git blame in the future

src/util.ts
```diff
@@ -80,7 +80,7 @@ export class GoVersion {
 
 	constructor(version: string) {
 		const matchesRelease = /go version go(\d.\d+).*/.exec(version);
-		const matchesDevel = /go version devel \+(.\d+).*/.exec(version);
+		const matchesDevel = /go version devel \+(.[a-zA-Z0-9]+).*/.exec(version);
```
ramya-rao-a: What does this fix?

test/integration/extension.test.ts
```diff
@@ -248,16 +254,16 @@ It returns the number of bytes written and any write error encountered.
 			[new vscode.Position(40, 23), 'package math', 'Package math provides basic constants and mathematical functions.\n\nThis package does not guarantee bit-identical results across architectures.\n'],
 			[new vscode.Position(19, 6), 'func Println(a ...interface{}) (n int, err error)', printlnDoc],
 			[new vscode.Position(27, 14), 'type ABC struct {\n    a int\n    b int\n    c int\n}', 'ABC is a struct, you coudn\'t use Goto Definition or Hover info on this before\nNow you can due to gogetdoc and go doc\n'],
-			[new vscode.Position(28, 6), 'func CIDRMask(ones, bits int) IPMask', 'CIDRMask returns an IPMask consisting of `ones\' 1 bits\nfollowed by 0s up to a total length of `bits\' bits.\nFor a mask of this form, CIDRMask is the inverse of IPMask.Size.\n']
+			[new vscode.Position(28, 6), 'func IPv4Mask(a, b, c, d byte) IPMask', 'IPv4Mask returns the IP mask (in 4-byte form) of the\nIPv4 mask a.b.c.d.\n']
 		];
-		const config = Object.create(vscode.workspace.getConfiguration('go'), {
+		const config = Object.create(getDefaultConfig(), {
```
ramya-rao-a: We should update this to follow the same format as other edits in this file

```javascript
const config = getDefaultConfig();
config['docsTool'] = 'gogetdoc';
```

ramya-rao-a:
Pushed a commit to fix the vendor related test that was failing for all Go versions

I am little concerned about trimming the docs in hover, signature help and completion widgets. Not sure how that will affect the experience. Was the issue not fixable by trimming the values before comparison in the assert statement?

src/util.ts
stamblerre: Go developer versions have a Git commit hash in them, for example:

`go version devel +868de9a111 Tue Oct 8 17:30:14 2019 +0000 linux/amd64`

This contains not just numbers, but letters as well.

ramya-rao-a:
@stamblerre,

I have fixed the failing tests for 1.13 via https://github.com/microsoft/vscode-go/commit/0839658c8d2dd95969d5463da9877dfa716d5fc6. The culprit happens to be https://github.com/microsoft/vscode-go/issues/2806#issuecomment-544277211

In this PR, I have reverted your changes to the tests and kept all other changes.

PR #2839, opened 2019-10-17
Distinguish between nil and zero length slices and maps

### LINKED ISSUE #2813: Delve shows difference between nil and zero-len/zero-cap slice, Go-extension does not
Even though delve is showing the difference between nil slices, and zero-len/zero-cap slices, the VSCode interface to the debugger is not doing so, and I appear to have the recent 0.11.7 release.
I also only have a single dlv.exe on my system of ver 1.2.0

 Delve showning nil vs zero-len/zero-cap slice
image

VScode Go extension not showing the difference
image

Test code:
```go
package main


import "fmt"

func main() {
	var x []string
	y:=[]string{}
	fmt.Println(x,y)

}
```

karthikraobr:
Resolves #2813

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -168,6 +168,7 @@ interface DebugVariable {
 	children: DebugVariable[];
 	unreadable: string;
 	fullyQualifiedName: string;
+	base: number;
 }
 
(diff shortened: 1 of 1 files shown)


karthikraobr:
@ramya-rao-a when you have some time could you please have a look :) Thanks!

Issue #2828, opened 2019-10-11
Debug test feature doesn't work for modules in gopath


SteelPhase:
Example is [here](https://github.com/SteelPhase/vscode-go-2828)

Running `Run Unit Test` while this module is outside of the gopath will result in the test debugging correctly.
Running `Run Unit Test` while this module is in the gopath will result in the test failing.

It appears that the execution of dlv test differs for modules outside of the gopath

- Module outside gopath `Running: /Users/steelphase/Code/Go/bin/dlv test --headless=true --listen=127.0.0.1:37823 --api-version=2 --log=true` 
- Module in gopath `Running: /Users/steelphase/Code/Go/bin/dlv test github.com/steelphase/vscode-go-2828/pkg/printer --headless=true --listen=127.0.0.1:47612 --api-version=2 --log=true`

The offending code appears to be here. I think there needs to be an additional check that the workspace we are in even if in gopath is a module. This is was changed in go 1.13 when the default for  `GO111MODULE` determines if project is a module differently.

https://github.com/microsoft/vscode-go/blob/9874b406352dbec2cb33648e9cc27fd1729bee06/src/debugAdapter/goDebug.ts#L433-L439

SteelPhase:
From #modules on the Gophers slack, it appears that relying on the output of `go env GOMOD` returning the go.mod file for the current working directory could be a decent solution to this. It would also support all the settings of `GO111MODULE`.

ramya-rao-a:
This only affects the time when GO111MODULE is set to `auto` right?

SteelPhase:
it does affect auto.  I'll check if the same behavior happens with on

ramya-rao-a:
> This is was changed in go 1.13 when the default for GO111MODULE determines if project is a module differently.

@SteelPhase Can you elaborate on what changed in go 1.13 for GO111MODULE?
cc @stamblerre

stamblerre:
I do believe the correct approach would be to use the result of `go env GOMOD`. This way, you don't have to worry about the changing meanings of `GO111MODULE`--you can rely on the `go` command to tell you that you are in a module.

https://golang.org/doc/go1.13#modules explains the changes to the meaning of `auto` for Go 1.13.

ramya-rao-a:
Please try the [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) of this extension has the fix for this

ramya-rao-a:
[...]

PR #2846, opened 2019-10-21
Use  mod file to determine module mode when debugging

ramya-rao-a:
This PR attempts to fix the issue in #2828 where code under GOPATH gets debugged under the assumption that we are not in module mode if GO111MODULE is not `on`.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -232,6 +232,7 @@ interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 
 	showGlobalVariables?: boolean;
 	currentFile: string;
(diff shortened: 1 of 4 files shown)


ramya-rao-a:
@SteelPhase I put something together that should help with #2828, but I haven't had the time to test it. Do you think you can take this for a spin? :)

SteelPhase:
I'll try it out in the morning, and report back

SteelPhase:
Tested it in outside of GOPATH and it works ok. When I test it inside GOPATH i get `Failed to continue: "TypeError: launchArgs.packagePathToGoModPathMap.get is not a function"`

ramya-rao-a:
Thanks @SteelPhase, can you try again?

SteelPhase:
This appears to be working as expected

PR #2586, opened 2019-06-20
Clear code coverage upon file save

alanlal:
PR for  [#2551](https://github.com/microsoft/vscode-go/issues/2551https://github.com/microsoft/vscode-go/issues/2551)

Update `goMain.ts` 

Added a new function to `goCover.ts` , `removeCodeCoverageOnFileSave` to handle File Save for clearing code coverage.

PR #2853, opened 2019-10-22
Fix #2551 : Code coverage disappears when typing a character

### LINKED ISSUE #2551: Go code coverage disappears when typing a character
The Go code coverage disappears when typing a character. This is implemented in removeCodeCoverageOnFileChange in goCover.ts. I find this very annoying, since I want to run the code coverage and then go through it and update the unit tests to add the missing tests, but as soon as I type a character in the unit test, the coverage information disappears.

I suggest that the code coverage should be removed when a Go file is saved, rather than when a character is typed. After all, the comment in the source saying that the code coverage is stale when a file is changed is false. The code coverage only becomes stale when the change is committed to disk (and would thus affect the result of compilation).

Alternately, please provide an option to disable removeCodeCoverageOnFileChange.

My current workaround is to change the language of the unit test file(s) from Go to plain text so it doesn't trigger the problem and then change back to Go before running the tests. It's not very nice.

Kegsay:
This also preserves an earlier fix to keep coverage when the modifications
are only due to comment edits.

Supersedes #2586


Kegsay:
The failing test is:
```
1) Test SignatureHelp Provider using gogetdoc
```
only on Go 1.12 on Mac which appears unrelated to these changes.

PR #2859, opened 2019-10-22
Activate extension when workspace contains go files

### LINKED ISSUE #2821: extension does not eager load when workspace contains .go files
The extension does not load on activationEvent  ""workspaceContains:**/.go"

This means users loading a workspace containing go files need to open a .go file, or run specific go commands before they can build the workspace.

Steps to Reproduce:
1. close all vscode instances
2. open a workspace that has .go files, ensure that the default state does not focus on a .go code file
3. try to ctrl-shift-p->"Go: Build Workspace"
4. Notice that it fails because the extension is not loaded

Possible solution: Modify package.json to include the workspaceContains event:
	"activationEvents": [
		"workspaceContains:**/.go",
		"onLanguage:go",
		"onCommand:go.gopath",
		"onCommand:go.tools.install",
		"onDebugResolve:go"
	],

zmb3:
Fixes #2821

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "vscode": "^1.30.0"
   },
   "activationEvents": [
+    "workspaceContains:**/*.go",
     "onLanguage:go",
     "onCommand:go.gopath",
     "onCommand:go.tools.install",


package.json
```diff
@@ -66,6 +66,7 @@
     "vscode": "^1.30.0"
   },
   "activationEvents": [
+    "workspaceContains:**/.go",
```
ToothlessGear: Shouldn't this be:
```suggestion
    "workspaceContains:**/*.go",
```
Otherwise it will only activate, if it finds a literal `.go` file.
I wanted to open a PR myself and this happened to me during testing, when omitting the asterisk in the globbing pattern.

zmb3: 🤦‍♂️oops, thanks @ToothlessGear

Issue #2380, opened 2019-03-15
go.json: review and update snippets


stamblerre:
From discussion with @ramya-rao-a on Slack:

- We should review the snippets in the go.json file and remove the ones that don’t make sense when one is in the middle of the file coding
- VS Code shouldnt have suggested this. `pkgm` only matches the `pkg` of `f.pkg`

Context: 
<img width="703" alt="screen_shot_2019-03-15_at_12 45 04_pm" src="https://user-images.githubusercontent.com/5856771/54448743-d20c6a80-4722-11e9-9ed9-42fa26975a85.png">

Ultimately, the language server (gopls) should return snippets, rather than having VSCode provide them.

mattouille:
Moving the snippets to gopls sounds like the right thing to do.

Before we do that, while I was assessing some of the snippets I started thinking about _what_ people use snippets for. In my experience I use them to:
1. Align myself to common patterns
2. To speed up development (things like anonymous function or method structure, etc...)

Would it be beneficial to establish how we envision snippets to be used to streamline the acceptance and maintenance workflow?

ramya-rao-a:
@mattouille Are you looking for an answer between the 2 options you have provided? I would imagine both to be applicable

mattouille:
I was more trying to determine what we want the standards for a snippet to be. The two points I supplied are what I use snippets for, but I was leaving it open ended for anyone to add on what they use snippets for.

hyangah:
We currently have a couple of open PRs to add snippets and we have already 50+ snippets.

I understand the desire for more snippets, but I think we need to stop adding more snippets to this extension. My reasoning is that

- due to the current limitation in VS Code's snippets (https://github.com/microsoft/vscode/issues/10565 etc), it's hard to selectively disable snippets included in an extension.

- there is clear overlap with some autocompletion features language servers can provide.

- different opinions/sentiment toward snippets exist.

- many of the existing snippets are shortcuts (rather than for encouraging common patterns). For those, personal snippets https://code.visualstudio.com/docs/editor/userdefinedsnippets#_create-your-own-snippets may serve better for the purpose.

stamblerre:
I agree that we should not add any more snippets without very good reason. I'm going to close all of the current outstanding snippet issues, but before I do so, I want to make sure we don't forget about these issues when we review snippets in the future. Here are the current snippet-related issues:

- Add struct and struct field snippets (https://github.com/microsoft/vscode-go/pull/2867)
[...]

PR #2867, opened 2019-10-29
Add struct and struct field snippets

CristianCurteanu:
Added following code snippets to improve struct code generation:
- Empty struct code snippet;
- Struct field code snippet, with typed tag mapping;
- Struct field code snippet, with empty tag body;
- Struct field code snippet, with JSON tag mapping;
- Empty struct, with a constructor that returns an empty instance of the same struct;


stamblerre:
Thank you for this contribution, @CristianCurteanu! Unfortunately, we won't be able to accept it at this time. We are currently reevaluating this extension's snippets and will not be accepting any new snippets for the time being. Please see more details here: https://github.com/microsoft/vscode-go/issues/2380#issuecomment-622009130.

PR #2877, opened 2019-10-31
run `go mod tidy` between tool installations

stamblerre:
This prevents modifications to the `go.mod` file from affecting other tool installations.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -155,7 +155,9 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 			const callback = (err: Error, stdout: string, stderr: string) => {
 				// Make sure to run `go mod tidy` between tool installations.
 				// This avoids us having to create a fresh go.mod file for each tool.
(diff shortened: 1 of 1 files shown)


src/goInstallTools.ts
```diff
@@ -148,7 +148,14 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
...
 
 		return res.then(sofar => new Promise<string[]>((resolve, reject) => {
+			const opts = {
+				env: envForTools,
+				cwd: toolsTmpDir,
+			};
 			const callback = (err: Error, stdout: string, stderr: string) => {
+				// Make sure to run `go mod tidy` between tool installations.
+				// This avoids us having to create a fresh go.mod file for each tool.
+				cp.execFileSync(goRuntimePath, ['mod', 'tidy'], opts);
```
ramya-rao-a: Shouldn't we do this only if `modulesOff` is false?

stamblerre: Done.

Issue #2822, opened 2019-10-09
Generate test function from a test file results in error opening file


pot-code:
Steps to Reproduce:

1. create a test file like `xxx_test.go`
2. move cursor to a function name(not a function prefixed with Test) and generate a test function, a error message then appears in the corner

What impressed me is that the desired test function is appended to file, without overwriting the current content.

Is there any way to remove the error message? Every time I generate a test function, it will appear.

Though it may be uncommon to do it, but it's useful sometimes, would you mind supporting it?

ramya-rao-a:
Thanks for reporting @pot-code, can you share the error message you are seeing?

pot-code:
Created a file named `demo_test.go` **without** the associated `demo.go` file

```
Unable to open 'demo.go': Unable to read file (Error: File not found (/Users/xxx/Documents/sandbox/playground/go/src/xxx.lang/spec/demo.go)).
```

It seems that it tries to open the `demo.go` file after creating the test function

ramya-rao-a:
Ah! understood.

The assumption always was that the "generate unit tests" command would be run on a non test file.
So, when the corresponding test file was updated, there was a need to open it.
For this, I re-used the `Toggle Test File` command.

In your case, the `Toggle Test File` command fails because the non test file doesnt exist :)

PRs are welcome to fix this bug
- [Set up local dev env](https://github.com/microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code)
- https://github.com/microsoft/vscode-go/blob/master/src/goGenerateTests.ts#L176 is where we attempt to toggle the file. You will have to add a check to not do this if we are operating on a test file already
- You can pass a flag via the [config passed to the generateTests](https://github.com/microsoft/vscode-go/blob/master/src/goGenerateTests.ts#L121) function to determine if the file from where the command was run was a test file or not

drakegens:
I will try to take this one on!

aykejriw:
@ramya-rao-a If no one is working here, I can do that.

ramya-rao-a:
I believe @drakegens has a PR out for this that had some feedback they had to react to.
@drakegens Are you planning on continuing your work in #2883?

drakegens:
Yes I will continue my work. Sorry work has been crazy.




On Mon, Nov 18, 2019 at 10:44 PM -0600, "Ramya Rao" <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #2883, opened 2019-11-02
Fixes Issue #2822

drakegens:
https://github.com/microsoft/vscode-go/issues/2822

### DIFF
--- a/src/goGenerateTests.ts
+++ b/src/goGenerateTests.ts
@@ -71,7 +71,10 @@ export function generateTestCurrentPackage(): Promise<boolean> {
 		return;
 	}
 	return generateTests(
-		{ dir: path.dirname(editor.document.uri.fsPath), isTestFile: editor.document.fileName.endsWith('_test.go') },
+		{
+			dir: path.dirname(editor.document.uri.fsPath),
+			isTestFile: editor.document.fileName.endsWith('_test.go')
+		},
 		getGoConfig(editor.document.uri)
 	);
 }
@@ -83,7 +86,10 @@ export function generateTestCurrentFile(): Promise<boolean> {
 	}
 
 	return generateTests(
-		{ dir: editor.document.uri.fsPath, isTestFile: editor.document.fileName.endsWith('_test.go') },
+		{
+			dir: editor.document.uri.fsPath,
+			isTestFile: editor.document.fileName.endsWith('_test.go')
+		},
 		getGoConfig(editor.document.uri)
 	);
 }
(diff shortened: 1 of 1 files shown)


src/goGenerateTests.ts
```diff
@@ -101,7 +101,14 @@ export async function generateTestCurrentFunction(): Promise<boolean> {
 	if (funcName.includes('.')) {
 		funcName = funcName.split('.')[1];
 	}
-	return generateTests({ dir: editor.document.uri.fsPath, func: funcName }, getGoConfig(editor.document.uri));
+
+	const currentFilePath = editor.document.fileName;
+	if (!currentFilePath.endsWith('.go')) {
```
ramya-rao-a: This check is already done in the `checkActiveEditor()` helper function which is called a few lines above. So we dont need it here.

src/goGenerateTests.ts
```diff
@@ -116,6 +123,11 @@ interface Config {
 	 * Specific function names to generate tests skeleton.
 	 */
 	func?: string;
+
+	/**
+	 * Whether or not the file to generate test functions for is a test file.
+	 */
+	testFile?: boolean;
```
ramya-rao-a: I would prefix this with an `is` i.e `isTestFile`

src/goGenerateTests.ts
```diff
@@ -101,7 +101,14 @@ export async function generateTestCurrentFunction(): Promise<boolean> {
...
 	}
-	return generateTests({ dir: editor.document.uri.fsPath, func: funcName }, getGoConfig(editor.document.uri));
+
+	const currentFilePath = editor.document.fileName;
+	if (!currentFilePath.endsWith('.go')) {
+		vscode.window.showInformationMessage('Cannot toggle test file. File in the editor is not a Go file.');
+		return;
+	}
+
+	return generateTests({ dir: editor.document.uri.fsPath, func: funcName, testFile: currentFilePath.endsWith('_test.go') }, getGoConfig(editor.document.uri));
```
ramya-rao-a: shouldn't we set this boolean for the other 2 cases where tests are generated for a file/package? i.e in `generateTestCurrentFile` and `generateTestCurrentPackage`?

src/goGenerateTests.ts
```diff
@@ -172,7 +184,10 @@ function generateTests(conf: Config, goConfig: vscode.WorkspaceConfiguration): P
 
 				vscode.window.showInformationMessage(message);
 				outputChannel.append(message);
-				if (testsGenerated) {
+
+				const isTestFile = (conf.testFile !== undefined && conf.testFile);
```
ramya-rao-a: Javascript is fun in a way that we can skip the undefined check :) Because null or undefined values would be "falsy"
And so your if check can directly be `if (testsGenerated && conf.testFile)`

drakegens:
Sorry for the delay. I have been on vacation. I will address these comments :)

drakegens:
@ramya-rao-a  I addressed your feedback! Please let me know if there's anything else.

PR #2885, opened 2019-11-05
Exclude prerelease tagged versions when fetching latest gopls version

stamblerre:
This change makes sure that we don't update users to pre-release versions of `gopls`.

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -518,15 +518,8 @@ async function latestGopls(tool: Tool): Promise<semver.SemVer> {
 	}
 	versions.sort(semver.rcompare);
 
-	let latestNonPreRelease: semver.SemVer;
-	for (const version of versions) {
-		// The first version in the sorted list without a prerelease tag.
-		if (version.prerelease.length === 0) {
-			latestNonPreRelease = version;
-			break;
-		}
-	}
-	return latestNonPreRelease;
+	// The first version in the sorted list without a prerelease tag.
+	return versions.find(version => !version.prerelease || !version.prerelease.length);
 }
 
 async function goplsVersion(goplsPath: string): Promise<string> {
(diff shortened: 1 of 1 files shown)


src/goLanguageServer.ts
```diff
@@ -507,13 +507,26 @@ async function latestGopls(tool: Tool): Promise<semver.SemVer> {
...
+
+	var latestNonPreRelease: semver.SemVer;
+	for (const version of versions) {
+		// The first version in the sorted list without a prerelease tag.
+		if (version.prerelease.length == 0) {
+			latestNonPreRelease = version;
+			break;
+		}
+	}
+	return latestNonPreRelease;
```
ramya-rao-a: Looks good

Pro tip: `const latestNonPreRelease = versions.find(version => !version.prerelease.length)`

src/goLanguageServer.ts
```diff
@@ -507,13 +507,26 @@ async function latestGopls(tool: Tool): Promise<semver.SemVer> {
...
 	if (versions.length === 0) {
 		return null;
 	}
 	versions.sort(semver.rcompare);
-	return versions[0];
+
+	let latestNonPreRelease: semver.SemVer;
+	for (const version of versions) {
+		// The first version in the sorted list without a prerelease tag.
+		if (version.prerelease.length === 0) {
```
ramya-rao-a: Is there a case where `version.prerelease` could be null/undefined or is it always initialized to an empty array?

src/goLanguageServer.ts
stamblerre: Done. Thanks!

src/goLanguageServer.ts
stamblerre: I've only seen it be initialized to an empty array in the cases I've tested.

Issue #1307, opened 2017-10-28
debug broken: vscode breakpoint set but not hit


suntong:
Using the `Go-latest.vsix` as instructed at
https://github.com/Microsoft/vscode-go/issues/1203#issuecomment-337772833
has caused my debugging not working any more -- I'm setting the break point at the first line in my main, but this is the only thing that I get (in the debug console):

```
ConfigurationDoneRequest
ContinueRequest
ContinueResponse
ThreadsRequest
2017/10/28 10:46:25 debugger.go:499: continuing
continue state {"Threads":null,"NextInProgress":false,"exited":true,"exitStatus":1,"When":""}
TerminatedEvent
DisconnectRequest
```

I.e., no break points works, and I simply cannot debug any more. Not using this Go-latest at least I can still do debugging. Is there any way I can go back?

Once again, all my own code are _not_ under GOPATH, as I want to distinct my own code from those I get from `go get`. That was the reason `vscode-go` was confused about GOPATH before and it might still be the reason that my debugging is not working, because after carefully examine the debug console log, I saw:

```
Error on CreateBreakpoint: unknown file: /path/to/repo/gitwork/src/github.com/suntong/dnstools/cmd_probe.go
Error on CreateBreakpoint: unknown file: /path/to/repo/gitwork/src/github.com/suntong/dnstools/dnstools_main.go
All set:[null]
All set:[null]
SetBreakPointsResponse
...
```

However, under my normal file system, the files are just fine. E.g., both `ls /path/to/repo/gitwork/src/github.com/suntong/dnstools/dnstools_main.go` and `file /path/to/repo/gitwork/src/github.com/suntong/dnstools/dnstools_main.go` works just fine.

ramya-rao-a:
- What is the version of VS Code you are using? The `Go-latest.vsix` needs VS Code 1.17.0 or higher.
- Does debugging work when you pass the GOPATH in the env property in the debug configuration in the launch.json file?

> Once again, all my own code are not under GOPATH, as I want to distinct my own code from those I get from go get

Here are you referring to the tools the Go extension depends on and downloads using `go get`? If yes, then you can use the `go.toolsGopath` setting to have a separate location to store all the dependent tools.

To go back to the old version, uninstall the Go extension and install it again from the marketplace.

suntong:
Thanks ramya, I am able to go back to the old version as per your help and my debugging is now working. 

[...]

PR #2908, opened 2019-11-20
Relativize source files passed to delve if possible

arifogel:
- add 'sourceRoot' launch configuration property
- fixes failure to set breakpoint when binary is built with trim-paths, e.g. via bazel
- possibly related to #1307

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/vscode-go/2908)

### DIFF
--- a/package.json
+++ b/package.json
@@ -410,6 +410,11 @@
           "launch": {
             "required": [],
             "properties": {
+              "sourceRoot": {
+                "type": "string",
+                "description": "Path with which to relativize source file paths. Required to debug binaries built with trim-paths e.g. with Bazel",
+                "default": "${workspaceFolder}"
+              },
               "program": {
                 "type": "string",
                 "description": "Path to the program folder (or any file within that folder) when in 'debug' or 'test' mode, and to the pre-built binary file to debug in 'exec' mode.",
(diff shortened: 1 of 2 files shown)


arifogel:
Hmm. Breakpoint is getting hit, and variable inspection works. But there are some usability issues I need to clean up. WIll do over next couple days.

arifogel:
Seems to work now. Remote debugging is unaffected - so if it didn't work before, it still won't work now.

tschundler:
I'm looking forward to this hopefully helping with my Bazel Go debugging.
Sounds like this may resolve #3011

yalcinmelihyasin:
I have just briefly looked at the plugin code so I cannot say much but have you tested it with different OS? I remember there were some issues with path separators on different OS. See #3011 for detailed information.

yalcinmelihyasin:
Also, I don't know if this would help with generated go code e.g. proto files. Although, even if it would not, it is a step towards right direction.

ramya-rao-a:
Hello all,

My apologies for the late response.

We have made some improvements in the way the path separators are used and mapped, but also realize the need to respect relative paths. We want to explore a few other alternatives before resorting to adding more fields/properties to the debug configuration as being done in this PR.

Please take a look at https://github.com/golang/vscode-go/issues/45 where @quoctruong has shared the current state of things and potential improvements. There is a linked PR that you can check out as well. We would appreciate your input there.

As for this PR, we will have to close it as we are currently in the midst of a repo move, see [We are moving](https://github.com/microsoft/vscode-go#we-are-moving---) section in our readme for more details. At the moment, we are not accepting any PR unless it is related to the move or critical bug fixes. 

Thanks for your efforts here and we hope to see you engage in the discussion in https://github.com/golang/vscode-go/issues/45

PR #2917, opened 2019-11-26
upgrade https-proxy-agent from 2.2.2 to 2.2.4

hyangah:
https://github.com/nodejs/security-wg/blob/master/vuln/npm/505.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1654,13 +1654,13 @@
       "integrity": "sha512-lTmS6AlAlMHOvPQemVwo3CezxBp0sNB95KNPkqp3Nxd5VFEnuG1ByM0zlRWos0zjO3ZWtkvhal0COgiV1xIA4A=="
     },
     "vscode-test": {
-      "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


ramya-rao-a:
This is a good catch @hyangah 

The dependency on https-proxy-agent is added via the `vscode-test` package which in itself has released an update with the latest https-proxy-agent

The right approach here is to update `vscode-test` which in turn would update `https-proxy-agent`

I have pushed a commit for the same to your branch.

ramya-rao-a:
Also, looks like this is your first PR contribution to this project, Thanks & Welcome!

ramya-rao-a:
Can you sign the CLA as linked in https://github.com/microsoft/vscode-go/pull/2917#issuecomment-561325261 please?

ramya-rao-a:
Sorry @hyangah 
I found another PR (#2900) doing the same where the contributor had already signed the CLA, so am going with that for the moment.

hyangah:
Thanks!

Issue #2936, opened 2019-12-10
gotests fails to be installed


eraserix:
### What version of Go, VS Code & VS Code Go extension are you using?
```
% go version
go version go1.12.14 linux/amd64
```

### Describe the bug
When trying to install gotests, I always get an error:
go get github.com/cweill/gotests/...: no matching versions for query "latest"

### Steps to reproduce the behavior:
1. mkdir -p /tmp/tmpmod
2. echo module tmpmod > /tmp/tmpmod/go.mod
2. cd /tmp/tmpmod
3. `env GO111MODULE=on go get -u -v "github.com/cweill/gotests/..."`
  `go: finding github.com/cweill/gotests/... latest`
  `Fetching https://github.com?go-get=1`
  `Parsing meta tags from https://github.com?go-get=1 (status code 200)`
  `go get github.com/cweill/gotests/...: no matching versions for query "latest"`

4. Or alternatively via goproxy:
  `env GOPROXY=https://proxy.golang.org GO111MODULE=on go get -u -v "github.com/cweill/gotests/..."`
  `Fetching https://proxy.golang.org/github.com/cweill/gotests/@v/list`
  `Fetching https://proxy.golang.org/github.com/cweill/@v/list`
  `Fetching https://proxy.golang.org/github.com/@v/list`
  `go get github.com/cweill/gotests/...: malformed module path "github.com/cweill/gotests/...": double dot`

### What seems to works
I think the problem is caused by a combination of "..." in the path and modules support. Explicitly specifying the binary gotests works:

```
% env GOPROXY=https://proxy.golang.org GO111MODULE=on go get -u -v "github.com/cweill/gotests/gotests"
Fetching https://proxy.golang.org/github.com/cweill/gotests/gotests/@v/list
Fetching https://proxy.golang.org/github.com/cweill/gotests/@v/list
Fetching https://proxy.golang.org/github.com/cweill/gotests/@v/v1.5.3.mod
Fetching https://proxy.golang.org/golang.org/x/tools/imports/@v/list
Fetching https://proxy.golang.org/golang.org/x/tools/@v/list
go: finding golang.org/x/tools latest
Fetching https://proxy.golang.org/golang.org/x/tools/@latest
```

ramya-rao-a:
Thanks for reporting @eraserix 

@stamblerre Can you take a look at this? It might be related to the changes we made around the tool installation code.

stamblerre:
[...]

PR #2948, opened 2019-12-16
delete and recreate go.mod instead of running go mod tidy

stamblerre:
Running `go mod tidy` is more error-prone than just deleting and recreating the file.

Fixes https://github.com/microsoft/vscode-go/issues/2936.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -134,10 +134,6 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 	// Install tools in a temporary directory, to avoid altering go.mod files.
 	const toolsTmpDir = fs.mkdtempSync(getTempFilePath('go-tools-'));
 
-	// Write a temporary go.mod file.
-	const tmpGoModFile = path.join(toolsTmpDir, 'go.mod');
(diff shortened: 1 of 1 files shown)


src/goInstallTools.ts
```diff
@@ -148,16 +144,19 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 			envForTools['GO111MODULE'] = 'on';
 		}
 
+		// Write a temporary go.mod file. This shouldn't affect anything if we've explicitly disabled modules.
+		const tmpGoModFile = path.join(toolsTmpDir, 'go.mod');
+		fs.writeFileSync(tmpGoModFile, 'module tools');
```
ramya-rao-a: With this, we are creating a new go.mod file for each tool being installed right?
In that case, should we move it inside the `else` block of the `if (modulesOffForTool)` check a few lines above?

ramya-rao-a: Also, did you check if `fs.writeFileSync` fails gracefully if the file already exists?

stamblerre: It shouldn't matter if we write the file even if modules are off, but I moved the check in case. `fs.writeFileSync` replaces the file if it already exists.

PR #2971, opened 2020-01-07
support showing diagnostics only on save

### LINKED ISSUE #2907: Help with live errors when using language server
Building up on https://github.com/microsoft/vscode-go/issues/2707
I would reopen it and add more there, but it is locked.

I ended up disabling the language server and I never ran into the issue again, although as of late gocode is taking up more resources than it should so I either can turn it off and not have a functional IDE, or try the language server.

This time around, I noticed that yeah, it's the language server (gopls) making the live errors come back, and I noticed that I cannot turn them off with the `go.liveErrors` config. The next solution would be to turn `go.LanguageServerExperimentalFeatures.diagnostics` to false, but that disables all error checking. 

I noticed that I can pass flags with `go.languageServerFlags`, but `gopls` doesn't seem to have any for live errors? 

The heck am I supposed to do to turn them off with the language server, _without_ disabling my error checking on save?

stamblerre:
This change extends the language server's diagnostic configuration to allow diagnostics to only be shown on save. This is done by ignoring diagnostics when a file is unsaved and then applying them when the file is saved.

Fixes #2907


GrimTheReaper:
Have you tested this? My question is does this also then disable code assist such as code completion until the file is saved?

GrimTheReaper:
Also, won't this break compatibility? You should probably add a `false` option to the enum list or check so it can be checked for people who turn it off.

stamblerre:
Yes, I have tested it. It won't disable any other features because it just hides diagnostics from the user's view (it doesn't actually change anything that gets sent to `gopls`). 

It does break compatibility with the settings, but because this is all experimental I think it should be fine. @ramya-rao-a can speak more to how settings should be transitioned. I would be happy to add some warning that pops up so that the user knows to change their settings.

GrimTheReaper:
Alright, sounds good. I look forward to seeing these changes 👍

ramya-rao-a:
> It does break compatibility with the settings, but because this is all experimental I think it should be fine. @ramya-rao-a can speak more to how settings should be transitioned. I would be happy to add some warning that pops up so that the user knows to change their settings.

Eventually, we want to move away from the `go.languageServerExperimentalFeatures` altogether and use the official settings for gopls. I'd rather have a transition plan for that point than a transition plan for intermediate steps. Let's consider the value `false` as `off` in the code, but not support it in the enum for the `diagnostics` setting. This way, `false` won't appear as an option when user is editing this setting. But if `false` was already set, it will be respected

src/goLanguageServer.ts
```diff
@@ -88,18 +94,38 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
 				},
 				handleDiagnostics: (uri: vscode.Uri, diagnostics: vscode.Diagnostic[], next: HandleDiagnosticsSignature) => {
-					if (!config.features.diagnostics) {
+					if (config.features.diagnostics == DiagnosticFrequency.Off) {
 						return null;
 					}
+					// If the user only wants diagnostics updated on save, 
+					// and this document is open in the editor, don't show the diagnostics.
+					if (config.features.diagnostics == DiagnosticFrequency.OnSave) {
+						latestDiagnostics.set(uri.toString(), diagnostics);
```
ramya-rao-a: Why not use the `uri` as key for the map and avoid the string conversions when setting and getting the values?

src/goLanguageServer.ts
```diff
@@ -88,18 +94,38 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
 			},
 			revealOutputChannelOn: RevealOutputChannelOn.Never,
 			middleware: {
-				provideDocumentFormattingEdits: (document: vscode.TextDocument, options: FormattingOptions, token: vscode.CancellationToken, next: ProvideDocumentFormattingEditsSignature) => {
-					if (!config.features.format) {
-						return [];
+				didSave: (data: vscode.TextDocument, next: NextSignature<vscode.TextDocument, void>) => {
+					// If the user only wants diagnostics on save, show the diagnostics now.
+					if (config.features.diagnostics == DiagnosticFrequency.OnSave) {
+						const diagnostics = latestDiagnostics.get(data.uri.toString());
```
ramya-rao-a: Why not use the `uri` as key for the map and avoid the string conversions when setting and getting the values?

src/goLanguageServer.ts
```diff
@@ -71,6 +75,8 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
 		promptForUpdatingTool(toolName);
 	}
 
+	const latestDiagnostics = new Map<string, vscode.Diagnostic[]>();
```
ramya-rao-a: Why not use the `uri` as key for the map and avoid the string conversions when setting and getting the values?

src/goLanguageServer.ts
```diff
@@ -88,18 +94,38 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+				didSave: (data: vscode.TextDocument, next: NextSignature<vscode.TextDocument, void>) => {
+					// If the user only wants diagnostics on save, show the diagnostics now.
+					if (config.features.diagnostics == DiagnosticFrequency.OnSave) {
+						const diagnostics = latestDiagnostics.get(data.uri.toString());
+						c.diagnostics.set(data.uri, diagnostics);
 					}
-					return next(document, options, token);
+					return next(data, null);
 				},
 				handleDiagnostics: (uri: vscode.Uri, diagnostics: vscode.Diagnostic[], next: HandleDiagnosticsSignature) => {
```
ramya-rao-a: I am not familiar as to when exactly is `handleDiagnostics` called? Is it for every file change or file open?

stamblerre: It's called whenever a diagnostics notification comes in - for `gopls`, we send diagnostics when needed for any file event (open, change, close, save, create, delete).

src/goLanguageServer.ts
stamblerre: The URI that comes into the `didSave` request has a lot more fields set than the one returned by the language server, so they can't be used as map keys. I did change it to use the `path` field instead of `toString()`.

src/goLanguageServer.ts
hyangah: Is it not necessary to clear the entry in latestDiagnostics - at least when files are moved, deleted, or closed?

stamblerre: I think it should be ok, because `gopls` should send those "clearing" diagnostics back to the client.

src/goLanguageServer.ts
ramya-rao-a: In that case, isn't it better to use the "gopls" settings such that `gopls` avoids calculating the diagnostics on file change if the user wants them only on file save?

stamblerre: `gopls` always parses and type checks on file changes, as that information is necessary for other `gopls` features, so there will be no benefits to pushing this configuration into `gopls` itself. That's the reasoning for this PR - to avoid displaying diagnostics if users don't want to see them without affecting `gopls`.

ramya-rao-a: understood

src/goLanguageServer.ts
```diff
@@ -88,18 +94,38 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
-					if (!config.features.diagnostics) {
+					if (config.features.diagnostics === DiagnosticFrequency.Off) {
 						return null;
 					}
+					// If the user only wants diagnostics updated on save,
+					// and this document is open in the editor, don't show the diagnostics.
+					if (config.features.diagnostics === DiagnosticFrequency.OnSave) {
+						latestDiagnostics.set(uri.path, diagnostics);
```
ramya-rao-a: Are we assuming that when a file is saved, the `handleDiagnostics` for that file is called first and then the `didSave`?

Because otherwise, `latestDiagnostics` would not have the latest diagnostics right?

stamblerre: `gopls` sends diagnostics on every file change, so diagnostics shouldn't change between a `didChange` event and a `didSave` event.

PR #2978, opened 2020-01-13
Refactor telemetry handling

stamblerre:
This change moves all of the telemetry-related code into a separate file, in order centralize it and make it clear what telemetry events the extension sends.

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -47,6 +47,7 @@ interface LanguageServerConfig {
 // registerLanguageFeatures registers providers for all the language features.
 // It looks to either the language server or the standard providers for these features.
 export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
+	console.log('hi~');
 	// Subscribe to notifications for changes to the configuration of the language server.
 	ctx.subscriptions.push(vscode.workspace.onDidChangeConfiguration(e => watchLanguageServerConfiguration(e)));
 
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -47,7 +47,6 @@ interface LanguageServerConfig {
 // registerLanguageFeatures registers providers for all the language features.
 // It looks to either the language server or the standard providers for these features.
 export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
-	console.log('hi~');
 	// Subscribe to notifications for changes to the configuration of the language server.
 	ctx.subscriptions.push(vscode.workspace.onDidChangeConfiguration(e => watchLanguageServerConfiguration(e)));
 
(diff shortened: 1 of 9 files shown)


src/goLanguageServer.ts
```diff
@@ -47,6 +47,7 @@ interface LanguageServerConfig {
 // registerLanguageFeatures registers providers for all the language features.
 // It looks to either the language server or the standard providers for these features.
 export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
+	console.log('hi~');
```
hyangah: remove this

src/util.ts
```diff
@@ -83,7 +83,7 @@ export class GoVersion {
 			this.isDevel = true;
 			this.commit = matchesDevel[0];
 		}
-		sendTelemetryEventForGoVersion(goVersion);
+		sendTelemetryEventForGoVersion(goVersion.format());
```
hyangah: The CI failure is real.

     Uncaught TypeError: Cannot read property 'format' of null
  	at new GoVersion (out/src/util.js:89:62)
  	at /Users/travis/gopath/src/github.com/microsoft/vscode-go/out/src/util.js:280:29
  	at ChildProcess.exithandler (child_process.js:294:7)
  	at maybeClose (internal/child_process.js:1021:16)
  	at Socket.<anonymous> (internal/child_process.js:430:11)
  	at Pipe.<anonymous> (net.js:586:12)

 Is "this.format" available by now?

stamblerre: Fixed

src/telemetry.ts
```diff
@@ -0,0 +1,183 @@
...
+ *--------------------------------------------------------*/
+
+import vscode = require('vscode');
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export const extensionId: string = 'ms-vscode.Go';
+const extensionVersion: string = vscode.extensions.getExtension(extensionId).packageJSON.version;
+const aiKey: string = 'AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217';
+
+export function sendTelemetryEventForModulesUsage() {
```
ramya-rao-a: What is the motivation for having a separate function for every kind of telemetry event that needs to be logged instead of moving only the `sendTelemetryEvent` to this file and leave all the files other than the utilities untouched?

stamblerre: My thinking was that it would be helpful to see all of the different telemetry events clearly in one file as opposed to having them throughout. It makes it really obvious what telemetry is being collected, and I liked the idea of being able to look in one file and quickly see all of the different metrics being collected. I also thought the GDPR comments made the code harder to read, so this might be a little simpler. But if you don't like this, I'm happy to just do the most minimal change.

PR #2979, opened 2020-01-13
Add a script to enable running tests locally

hyangah:
It's not obvious how to run the vscode extension tests - there are many tools and environments involved to run tests reliably in addition to the node dependencies. This PR adds a script that runs the tests in a docker container.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -7,3 +7,4 @@ node_modules/fs-extra
 test/
 **/*.map
 **/tslint.json
+build/**/*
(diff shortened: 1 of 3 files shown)


ramya-rao-a:
The npm script `test` should run all the tests assuming that all the Go tools are already installed. Have you tried running `npm run test`?

hyangah:
@ramya-rao-a  Currently `npm run test` passes and that's what this script calls. The goal of this script is not require the plugin developers to install all the tools on their machine but still allow to run the test. Moreover, the container provides a clean environment for the testing. I have a conflicting plugin for my personal vscode dev environment, and that plugin interferes with the testing and caused failure. We can also use this script when setting up in ci. Much of the code in the script was copied from the current travis configuration of this project.

Issue #2282, opened 2019-01-28
Unit test generation should consider the receiver type


Guitarbum722:
**Description:**
The Go Plugin doesn't seem to consider the receiver type for a method.  So, if there are `>=` 1 method names in the same file, then clicking "generate unit tests for function" creates test functions for _all_ of them.  

[This](https://github.com/Microsoft/vscode-go/issues/1880) references an issue with `gotests`.  However, that same [gotests issue](https://github.com/cweill/gotests/issues/35) explains that the receiver type is in fact supported.  

I tried this by running the following (as an example) in the command line using `gotests` directly.
`$ gotests -w -only ^PersonValidate$ .  `
This works as expected, and there is only a test function generated for the `Person` type.

**Version of VS Code, the Go plugin and Go:**
VSCode: ```1.30.2
61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8
x64```
Go Plugin: ```0.8.0```
Go: 1.11.2 darwin/amd64

**OS/Platform you are working on**
MacOS / AMD64

**Reproducible steps (1... 2... 3...) that cause the issue**
1. right click a method name 
2. select "Generate Unit Tests for Function
3. all methods that share the same name but are on different types have a test function generated

**What you expected to see, versus what you actually saw**

I would expect that the Go extension uses `gotests` with the `-only` flag and prepends the type name with the regexp.  Therefore, one single test function is generated for the respective method that was selected.

Instead, any methods that have the same name in the file have a test function generated.

Great extension though! 👍 Hopefully this is enough info.

ramya-rao-a:
Thanks for reporting @Guitarbum722 

So looks like all we need to do is identify the receiver type, convert it to title case, prepend it to the function name and then pass the resulting string to `gotests`

PRs are welcome to fix this
Code Pointers:
- See [Build and Debug Extension](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code) to set up local dev environment
- https://github.com/Microsoft/vscode-go/blob/0.8.0/src/goGenerateTests.ts#L86 is the function that is called for generating unit tests for a function, so all changes would go here.

jamesgeorge007:
@ramya-rao-a I'm new here. I would like to give it a try :clap: 
Are the required workarounds to be made within `getFunctions()` methods where the current function is being figured out before `generateTests()` is being invoked.
[...]

PR #2987, opened 2020-01-17
Considering receiverType on function test creation issue-2282

marcel-basel:
The receiver type for a method is now considered. So, if there are >= 1 method names in the same file, then clicking "generate unit tests for function" creates test functions only for the selected method.

funcName like "(*Point).Move" will results in "gotests -w -only ^PointMove$ calls.

### DIFF
--- a/src/goGenerateTests.ts
+++ b/src/goGenerateTests.ts
@@ -98,17 +98,13 @@ export async function generateTestCurrentFunction(): Promise<boolean> {
 		return Promise.resolve(false);
 	}
 	let funcName = currentFunction.name;
-	if (funcName.includes('.')) {  // receiverType specified
-		let funcNameParts = funcName.match(/^\(\*?(.*)\)\.(.*)$/);
-		if(funcNameParts != null && funcNameParts.length === 3){
-			let rType = funcNameParts[1].replace(/^\w/, c => c.toUpperCase());
-			let fName = funcNameParts[2].replace(/^\w/, c => c.toUpperCase());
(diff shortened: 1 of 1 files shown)


src/goGenerateTests.ts
```diff
@@ -98,8 +98,16 @@ export async function generateTestCurrentFunction(): Promise<boolean> {
 		return Promise.resolve(false);
 	}
 	let funcName = currentFunction.name;
-	if (funcName.includes('.')) {
-		funcName = funcName.split('.')[1];
+	if (funcName.includes('.')) {  // receiverType specified
+		let funcNameParts = funcName.match(/^\(\*?(.*)\)\.(.*)$/);
```
ramya-rao-a: If we are going to use a regular expression, then the `if (funcName.includes('.'))` check is redundant right?

Do we foresee the function name to include `.` in a non receiverType case?

marcel-basel: Hi Ramya, thanks for your fantastic work on the extension!

I removed the `if (funcName.includes('.'))` block. I don't see an other case where the function name would include `.` Do you?

ramya-rao-a: Nope, change looks good. I have pushed a commit to fix the linting errors, we are good to merge now

marcel-basel:
Thanks @ramya-rao-a, do you consider #2282 as closable?

ramya-rao-a:
I'll be closing it once I release an update with the fix.

PR #2988, opened 2020-01-17
Update to latest vscode-languageclient

stamblerre:
The next version of the vscode-languageclient has been released.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "vscode-debugadapter": "^1.36.0",
     "vscode-debugprotocol": "^1.36.0",
     "vscode-extension-telemetry": "^0.1.2",
-    "vscode-languageclient": "~6.0.1",
+    "vscode-languageclient": "^6.0.1",
     "web-request": "^1.0.7"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -38,29 +38,29 @@
   },
   "extensionDependencies": [],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


stamblerre:
There are also a few major version updates there too, and I actually used the https://www.npmjs.com/package/npm-check-updates tool to generate this PR. I find it simplifies the process, but if you'd prefer to only update the single dependency I am happy to revert the other changes.

ramya-rao-a:
Major version updates implies that there can be breaking changes in the product.
Since I don't usually have the time to confirm that none of the breaking changes affect the extension, I avoid updating major versions unless I absolutely have to.

stamblerre:
Sounds good. I edited the PR to only update the vscode-languageclient dependency.

package.json
```diff
@@ -46,7 +46,7 @@
     "vscode-debugadapter": "^1.36.0",
     "vscode-debugprotocol": "^1.36.0",
     "vscode-extension-telemetry": "^0.1.2",
-    "vscode-languageclient": "~5.2.1",
+    "vscode-languageclient": "~6.0.1",
```
ramya-rao-a: Do you want to consider changing the ~ here to ^?

stamblerre: Done. I wasn't sure if there was some reason that was a ~.

ramya-rao-a: There was a time when a new feature got added which wasnt supported yet, caused some problems, so we moved to getting only patch updates.

PR #2998, opened 2020-01-22
remove obsolete enableBreakpointsFor

hyangah:
This suppresses the lint warning.
There is already the "languages" field.

Context: See github.com/microsoft/vscode/issues/9037

### DIFF
--- a/package.json
+++ b/package.json
@@ -325,11 +325,6 @@
       {
         "type": "go",
         "label": "Go",
-        "enableBreakpointsFor": {
-          "languageIds": [
-            "go"
-          ]
-        },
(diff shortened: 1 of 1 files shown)


hyangah:
> Have you tested the debugging feature after making this change?
> I was not able to add any breakpoints after this change.
> 
> After reading through [microsoft/vscode#9037](https://github.com/microsoft/vscode/issues/9037), it looks like we need to have the `breakpoints` under `contributes` as a replacement for `enableBreakpointsFor`

You are right. I was confused by debuggers.languages field. Added. Observed I could add breakpoints and debugger obeyed. Thanks!

PR #2999, opened 2020-01-23
Show warning if 'program' attribute is used in 'remote'

quoctruong:
When using the debugger in remote attach mode, users have to use `cwd` instead of `program`. This can be confusing so we should show a warning message.

### DIFF
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -118,7 +118,9 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 				`Request type of 'launch' with mode 'remote' is deprecated, please use request type 'attach' with mode 'remote' instead.`);
 		}
 
-		if (debugConfiguration.request === 'launch' && debugConfiguration['mode'] === 'remote') {
+		if (debugConfiguration.request === 'attach'
+			&& debugConfiguration['mode'] === 'remote'
+			&& debugConfiguration['program']) {
 			this.showWarning(
 				'ignoreUsingRemotePathAndProgramWarning',
 				`Request type of 'attach' with mode 'remote' does not work with 'program' attribute, please use 'cwd' attribute instead.`);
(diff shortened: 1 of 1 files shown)


src/goDebugConfiguration.ts
```diff
@@ -1,136 +1,147 @@
...
+			debugConfiguration.request === 'attach' && debugConfiguration['mode'] === 'remote' && debugConfiguration['program']);
+
+		return debugConfiguration;
+	}
+
+	/**
+	 * Show warningMessage to the user if warningCondition is met.
+	 * If the user chooses not to see the warning again, sets it to the global state.
+	 */
+	private showWarning(warningKey: string, warningMessage: string, warningCondition: boolean) {
```
ramya-rao-a: If `warningCondition` is true, then the entire code path in this method is a no-op. It would be easier to read the code if this condition was checked outside and this function called only if it were true

ramya-rao-a: Also would suggest renaming `warningKey` to `ignoreWarningKey`

src/goDebugConfiguration.ts
```diff
@@ -112,25 +112,31 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
...
-						updateGlobalState(ignoreWarningKey, true);
-					}
-				});
+		if (debugConfiguration.request === 'launch' && debugConfiguration['mode'] === 'remote') {
+			this.showWarning(
+				'ignoreDebugLaunchRemoteWarning',
+				`Request type of 'launch' with mode 'remote' is deprecated, please use request type 'attach' with mode 'remote' instead.`);
+		}
+
+		if (debugConfiguration.request === 'launch' && debugConfiguration['mode'] === 'remote') {
```
ramya-rao-a: Shouldn't the request here be `attach`?

quoctruong:
@ramya-rao-a PTAL

PR #3004, opened 2020-01-27
Update description for useLanguageServer setting

ramya-rao-a:
@stamblerre Would you like any other tweaks here?


ramya-rao-a:
cc @hyangah

stamblerre:
Looks good to me - thanks!

Issue #2607, opened 2019-06-26
VSCode needs to be reloaded when gopls is updated


SteelPhase:
Would there be any interest in replacing the message after installing gopls with a call to `Go: Restart Language Server`, or a similar method instead.

```
Installing 1 tool at /Users/ebajumpa/Code/Go/bin
  gopls

Installing golang.org/x/tools/cmd/gopls SUCCEEDED

Reload VS Code window to use the Go language server
All tools successfully installed. You're ready to Go :).
```

ramya-rao-a:
I guess now that we have a command to restart the language server, we don't need to reload VS Code window? It needs to be tested though.

If so, then we can update the code that prints this message to instead just restart the language server.

ramya-rao-a:
cc @stamblerre

stamblerre:
I think we could probably go as far as running the command ourselves without asking the user to, right?

ramya-rao-a:
What is the behavior when neither VS Code is reloaded, nor the language server is restarted after an update?

stamblerre:
I would imagine that the server wouldn't start, since we only start it on extension activation or if the restart command is executed.

stamblerre:
This is fixed in the nightly extension by https://github.com/golang/vscode-go/pull/30.

PR #3012, opened 2020-01-29
gopls: restart language server automatically instead of asking user to reload

stamblerre:
Updates https://github.com/microsoft/vscode-go/issues/2607

Minimize the number of instances where we ask the user to reload their VS Code window.
There are still a few more places where this command could be improved.


stamblerre:
Closing in favor of https://github.com/golang/vscode-go/pull/3.

PR #3014, opened 2020-01-29
Logic for gopls updater to accept a pre-release version

hyangah:
acceptGoplsPrerelease is a const set to false.
We are currently discussing what's the best way to expose this knob to users. 
Meanwhile, developers can flip the var to test whether Gopls picks up the newest version including prerelease.

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -8,6 +8,7 @@
 import cp = require('child_process');
 import fs = require('fs');
 import path = require('path');
+import { SemVer } from 'semver';
 import vscode = require('vscode');
 import { getLanguageServerToolPath } from './goLanguageServer';
 import { envPath, getToolFromToolPath } from './goPath';
@@ -18,6 +19,7 @@ import {
 	disableModulesForWildcard,
 	getConfiguredTools,
 	getImportPath,
+	getImportPathWithVersion,
 	getTool,
 	hasModSuffix,
 	isGocode,
@@ -89,11 +91,21 @@ export async function installAllTools(updateExistingToolsOnly: boolean = false) 
 }
 
 /**
+ * ToolAtVersion is a Tool with version annotation.
+ * Lack of version implies the latest version
+ */
+export interface ToolAtVersion extends Tool {
+	version?: SemVer;
+}
+
+/**
  * Installs given array of missing tools. If no input is given, the all tools are installed
  *
- * @param string[] array of tool names to be installed
+ * @param missing array of tool names and optionally, their versions to be installed.
+ *                If a tool's version is not specified, it will install the latest.
+ * @param goVersion version of Go that affects how to install the tool. (e.g. modules vs legacy GOPATH mode)
  */
-export function installTools(missing: Tool[], goVersion: GoVersion): Promise<void> {
+export function installTools(missing: ToolAtVersion[], goVersion: GoVersion): Promise<void> {
 	const goRuntimePath = getBinPath('go');
 	if (!goRuntimePath) {
 		vscode.window.showErrorMessage(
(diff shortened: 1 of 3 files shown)


hyangah:
@stamblerre @ramya-rao-a

src/goLanguageServer.ts
```diff
@@ -368,29 +369,30 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
...
+		return null;
 	}
 
 	// First, run the "gopls version" command and parse its results.
 	// If "gopls" is so old that it doesn't have the "gopls version" command,
 	// or its version doesn't match our expectations, prompt the user to download.
 	const usersVersion = await goplsVersion(languageServerToolPath);
 	if (!usersVersion) {
-		return true;
+		return null;
```
ramya-rao-a: In this case, we do want to prompt user to install the latest version. By returning `null`, we end up not prompting at all

src/goInstallTools.ts
```diff
@@ -82,12 +84,22 @@ export async function installAllTools(updateExistingToolsOnly: boolean = false)
 		});
 }
 
+/**
+ * ToolAtVersion is a Tool with version annotation.
+ * Lack of version implies the latest version
+ */
+export interface ToolAtVersion extends Tool {
+	version?: SemVer;
```
ramya-rao-a: Why not add `version` as an optional property to `Tool` itself?

src/goInstallTools.ts
```diff
@@ -163,6 +175,10 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 
 	outputChannel.appendLine(installingMsg);
 	missing.forEach((missingTool) => {
+		let toolName = missingTool.name;
+		if (missingTool.version) {
+			toolName += '@' + missingTool.version;
```
ramya-rao-a: `toolName` is updated but not used

hyangah: good catch. fixed.

src/goInstallTools.ts
hyangah: Tool is used in various places, especially those in the allToolsInformation map, or other places worry me. Without properly cloning, the version information may be stored or not stored accidentally and misused. So, I decided to create an explicit versioned type.

src/goLanguageServer.ts
hyangah: nice catch. Fixed.

src/goInstallTools.ts
ramya-rao-a: I am a little hesitant in introducing new interfaces unless absolutely required. Based on the changes in this PR, it looks like the code path for the prompt to update `gopls` is the only one that would use this `version` field. Do we foresee this feature to be extended to other tools? If not, then I would recommend having a separate function to prompt the update for `gopls` say `promptForUpdatingGopls` whose input parameter can be the version

ramya-rao-a: Nevermind, looks like the problem is going to be to plumb that information all through `installTools`

src/goLanguageServer.ts
```diff
@@ -421,29 +422,25 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
 	vscode.workspace.onDidChangeTextDocument(parseLiveFile, null, ctx.subscriptions);
 }
 
+const acceptGoplsPrerelease = false;
```
ramya-rao-a, line 425: Am guessing you want to expose this via a setting? Have we decided if this will go inside the `gopls` setting or will we add an official setting?

src/goInstallTools.ts
stamblerre: I think that, in general, @hyangah was hoping to introduce the ability to pin tools to specific versions in the extension. That would fix the issue of tools breaking in certain versions, etc.

src/goLanguageServer.ts
stamblerre: This is actually a switch for the Go Nightly extension - one of the patches converts the value to true for Go Nightly, false for the regular Go extension. It's not something that could be a `gopls` setting because it's about which version of `gopls` to install rather than something `gopls` can use.

ramya-rao-a: got it

PR #3016, opened 2020-01-30
Considering if launch.json's program attribute resolves to a file in debug mode

marcel-basel:
Fixing #1229 

Implemented intial proposal by @ramya-rao-a 

When the `program` attribute points to a file (${file} or absolute path), the filepath is now passed to the delve debug command. Thus debugging also works if multiple main files exist in the package. Launching debug from active editor file with no main function (Program = ${file}) will result in following error "Can not debug non-main package". 

Program ${fileDirname} configuration works as before. 

Updated the default value for program attribute in goDebugConfiguration.ts file to be {$file} instead of ${fileDirname}

Documentation [https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code] (url) should be updated e.g. first launch.json example. Since I'm not a native English speaker so not the best person to help.

### DIFF
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -24,7 +24,7 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 				type: 'go',
 				request: 'launch',
 				mode: 'auto',
-				program: '${file}',
+				program: '${fileDirname}',
 				env: {},
 				args: []
 			}
--- a/src/goDebugConfiguration.ts
+++ b/src/goDebugConfiguration.ts
@@ -24,7 +24,7 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 				type: 'go',
 				request: 'launch',
 				mode: 'auto',
-				program: '${fileDirname}',
+				program: '${file}',
 				env: {},
 				args: []
 			}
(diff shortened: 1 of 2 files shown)


src/goDebugConfiguration.ts
```diff
@@ -24,7 +24,7 @@ export class GoDebugConfigurationProvider implements vscode.DebugConfigurationPr
 				type: 'go',
 				request: 'launch',
 				mode: 'auto',
-				program: '${fileDirname}',
+				program: '${file}',
```
ramya-rao-a: I understand this change is based on my comment https://github.com/microsoft/vscode-go/issues/1229#issuecomment-473731132

But I now realize that this change can break the workflow for existing users who might 
start debugging from a file that does not have the main function. Having `${fileDirname}` ensured that such scenarios are supported.

I recommend we not make this change. Folks wanting to have multiple files with main function in the same folder, can create a debug configuration with ${file} and have the debugging feature work for them

ramya-rao-a:
@marcel-basel I tested the changes here with the below set up and I see failures

- Have a folder with 2 go files
- Have the below in the first file
```go
package main

import "fmt"

func main() {
	fmt.Println(Bye())
}
```
- Have the below in the second file
```go
package main

func Bye() string {
	return "bye"
}
```
- Have a debug configuration where `program` is set to `${file}`
- Open the first file in the editor and press F5

I get the below error:
`undefined: Bye`

Looks like when we pass a single file to `dlv debug`, delve does not expect any other file to exist in the package

ramya-rao-a:
cc @quoctruong

ramya-rao-a:
cc @jhendrixMSFT

ramya-rao-a:
I discussed with @quoctruong and @jhendrixMSFT offline and it looks like the above behavior matches the `go run filepath` behavior, so we should be good

PR #3018, opened 2020-01-30
update README.md mostly where gopls is mentioned

hyangah:
@stamblerre @ramya-rao-a

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,16 +36,16 @@ Read the [Changelog](https://github.com/Microsoft/vscode-go/blob/master/CHANGELO
 
 ### IntelliSense
 
-- Auto Completion of symbols as you type (using `gocode`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,16 +36,16 @@ Read the [Changelog](https://github.com/Microsoft/vscode-go/blob/master/CHANGELO
 
 ### IntelliSense
 
-- Auto Completion of symbols as you type (using `gocode`)
-- Signature Help for functions as you type (using `gogetdoc` or `godef`+`go doc`)
-- Quick Info on the symbol as you hover over it (using `gogetdoc` or `godef`+`go doc`)
+- Auto Completion of symbols as you type (using `gopls` or `gocode`)
```
ramya-rao-a: How about the below instead?

> using language server or `gocode`

Otherwise, I worry that the message is that `gopls` is just another tool

hyangah: Done.

PR #3020, opened 2020-01-31
checking setting 'go.buildTags' to contain build tags from file fixing #1497

marcel-basel:
If the current file has build tags which are not present in `go.buildtags` the user is now prompted to update the `go.buildtags` setting. "Don't show again" and  "Open settings" were implemented on the prompt.

**Example**
```
// +build linux,386 darwin,!cgo
// +build development
``` 
translates to condition `((linux AND 386) OR (darwin AND (NOT cgo))) AND development` therefore go.buildtags should contain values `linux 386 development` OR `darwin cgo development`

Implementation is based on previous work of @urohit001

### DIFF
--- a/src/goCheck.ts
+++ b/src/goCheck.ts
@@ -29,11 +29,16 @@ export function removeTestStatus(e: vscode.TextDocumentChangeEvent) {
 	statusBarItem.text = '';
 }
 
-export function notifyIfGeneratedFile(this: void, e: vscode.TextDocumentChangeEvent) {
+export function checksOnFileEdit(this: void, e: vscode.TextDocumentChangeEvent) {
 	const ctx: any = this;
 	if (e.document.isUntitled || e.document.languageId !== 'go') {
 		return;
 	}
+	notifyIfGeneratedFile(ctx, e);
+	notifyOnBuildTagMismatch(ctx, e);
+}
+
+function notifyIfGeneratedFile(ctx: any, e: vscode.TextDocumentChangeEvent) {
 	if (
 		ctx.globalState.get('ignoreGeneratedCodeWarning') !== true &&
 		e.document.lineAt(0).text.match(/^\/\/ Code generated .* DO NOT EDIT\.$/)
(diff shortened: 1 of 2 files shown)


src/goCheck.ts
```diff
@@ -46,6 +51,66 @@ export function notifyIfGeneratedFile(this: void, e: vscode.TextDocumentChangeEv
...
+
+	let areInSetting = false;
+	const goConfig = vscode.workspace.getConfiguration('go', e.document.uri);
+	let settBldTags = goConfig['buildTags'];
+
+	if (settBldTags) {
+		const splChr = settBldTags.includes(',') ? ',' : ' ';
+		settBldTags = settBldTags.split(splChr);
+		areInSetting = fileBuildTags.every((tag: string) => {
+			// build tags OR array containing AND array
```
hyangah: I thought go.buildTags are for -tags flag. Usually architecture and os related build constraints are supposed to be specified using environment variables (GOOS/GOARCH) instead of -tags flag.

marcel-basel: You are right! There are many implicitly provided words like os and arch specified in GOOS & GOARCH. Also the compiler, CGO, version of Go and the "ignore" word. So only the additional words have to be satiesfied via -tags. ([https://golang.org/pkg/go/build/](url)).

So I would not check for: Currently supported os/arch, gc. gccgo, cgo, Go version gox.xx, "ignore" and still catch missing custom build tag like 'development'. 

Is this still wanted or does it smell like 'over engineering'?

src/goCheck.ts
```diff
@@ -29,11 +29,16 @@ export function removeTestStatus(e: vscode.TextDocumentChangeEvent) {
...
 }
 
-export function notifyIfGeneratedFile(this: void, e: vscode.TextDocumentChangeEvent) {
+export function checksOnFileEdit(this: void, e: vscode.TextDocumentChangeEvent) {
 	const ctx: any = this;
 	if (e.document.isUntitled || e.document.languageId !== 'go') {
 		return;
 	}
+	notifyIfGeneratedFile(ctx, e);
+	notifyOnBuildTagMismatch(ctx, e);
```
ramya-rao-a, line 38: What is the reason to have this notification be shown as part of the file change event?
I would expect the check on build tags to happen right before the features mentioned in #1497 are triggered.
Otherwise one would get prompted way more times and even in situations where they might be doing simple edits like renaming or formatting and not really looking to build the code.

src/goCheck.ts
```diff
@@ -46,6 +51,66 @@ export function notifyIfGeneratedFile(this: void, e: vscode.TextDocumentChangeEv
...
+function notifyOnBuildTagMismatch(ctx: any, e: vscode.TextDocumentChangeEvent) {
+	if (ctx.globalState.get('ignoreBuildTagMismatchWarning') === true) {
+		return;
+	}
+
+	const fileBuildTags = [];
+	for (let i = 0; i < e.document.lineCount; i++) {
+		const line = e.document.lineAt(i);
+		if (line.isEmptyOrWhitespace) {
+			break;
```
ramya-rao-a, line 63: Say I have a file with a line comment preceding the build constraing
```go
// my comment

// +build hello
```

This is a valid use of build constraints, but you will end up exiting the loop based on the current code.

Based on the [docs for build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints), line comments and empty lines can preceed the build constraint

src/goCheck.ts
```diff
@@ -46,6 +51,66 @@ export function notifyIfGeneratedFile(this: void, e: vscode.TextDocumentChangeEv
...
+		}
+		const bldTagLine = line.text.match(/^\/\/\s*\+build\s+(.+)$/);
+		if (bldTagLine) {
+			if (i === e.contentChanges[0].range.start.line) {
```
ramya-rao-a, line 67: Why check for only the first entry in `e.contentChanges` ?

ramya-rao-a:
Hey @marcel-basel,

Thanks for all your efforts here, but as explained in https://github.com/microsoft/vscode-go/issues/1497#issuecomment-619631149, we are planning not to go ahead with this feature. If you have any thoughts on this, please reply on the issue #1497.

Issue #3028, opened 2020-02-04
settings: create one setting to specify an alternate language server


stamblerre:
Right now, users can specify alternative tools for VS Code Go to use through the `go.alternateTools` setting. Both `"gopls"` and `"go-langserver"` can be replaced, which makes for a confusing user experience. What is the expected behavior if both of these are configured?

I propose that we create one setting for the language server specifically, something like:

```json5
"go.alternateTools": {
    "language-server": "gopls"
}
```

Where `"gopls"` might be the default, `"go-langserver"` is another option, and any other text is also allowed.

@ramya-rao-a, @hyangah: What do you think?

hyangah:
+1 for refactoring.

I wonder if `go.useLanguageServer` should be enum instead of a bool.
 
`go.alternateTools` can be used to specify a wrapper script or an absolute path to the corresponding binary (for example, you can utilize this to test or ensure the use of a different version of gopls without uninstalling or installing the gopls binary from the PATH) so depending on go.alternateTools['...'] to be `gopls`  and `go-langserver` should be done carefully.

ramya-rao-a:
> Both "gopls" and "go-langserver" can be replaced, which makes for a confusing user experience. What is the expected behavior if both of these are configured?

Today, if the "alternate tool" set for `gopls` exists, then that takes precedence over the alternate tool set for `go-langserver`.

We are well beyond the point of supporting `go-langserver`. It was supported in the absence of `gopls` during a time when we were making the transition to using `gopls`. I am ok removing the support for `go-langserver`. If users still want to use it, they can always map `gopls` to `go-langserver`.

Regarding using something generic `language-server` as the "key" in `go.alternateTools`, I don't believe that gives us a lot of benefits for the amount of overhead supporting it would add

ramya-rao-a:
> I wonder if go.useLanguageServer should be enum instead of a bool.

What would the enum support?

stamblerre:
I agree that adding an enum might encourage users to try `go-langserver`, which we shouldn't do. I think that my PR (https://github.com/microsoft/vscode-go/pull/3029) will achieve what we want for now, which is to only run `go-langserver` if the user sets it as an alternate tool for `gopls`. I didn't make the change to using the generic `"language-server"` configuration, so I am fine leaving it as is.

hyangah:
@ramya-rao-a 

> I wonder if go.useLanguageServer should be enum instead of a bool.
[...]

PR #3029, opened 2020-02-04
refactor handling for go-langserver installation

stamblerre:
Fixes https://github.com/microsoft/vscode-go/issues/3028.

I thought that we actually supported the "go.alternateTools" settings for the "go-langserver", but it seems it's been deprecated. Remove the code that handled this in the language server installation, and be careful to only install the tool specified by the user.

Also, remove some unused import lines in the `goLanguageServer.ts` file.

/cc @hyangah

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -307,7 +307,8 @@ export function getLanguageServerToolPath(): string {
 		// Check if the user has set the deprecated "go-langserver" setting.
 		const golangserverAlternate = goConfig['alternateTools']['go-langserver'];
 		if (golangserverAlternate) {
-			vscode.window.showErrorMessage(`The "go.alternateTools" setting for "go-langserver" has been deprecated. Please use "gopls" instead.`);
+			vscode.window.showErrorMessage(`The "go.alternateTools" setting for "go-langserver" has been deprecated.
+Please set "gopls" instead, and then reload the VS Code window.`);
 			return;
 		}
 		if (goplsAlternate) {
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -15,21 +15,10 @@ import {
 	FormattingOptions,
 	HandleDiagnosticsSignature,
 	LanguageClient,
-	ProvideCompletionItemsSignature,
-	ProvideDefinitionSignature,
 	ProvideDocumentFormattingEditsSignature,
-	ProvideDocumentHighlightsSignature,
 	ProvideDocumentLinksSignature,
-	ProvideDocumentSymbolsSignature,
-	ProvideHoverSignature,
-	ProvideReferencesSignature,
-	ProvideRenameEditsSignature,
-	ProvideSignatureHelpSignature,
-	ProvideWorkspaceSymbolsSignature,
 	RevealOutputChannelOn
 } from 'vscode-languageclient';
-import { ProvideImplementationSignature } from 'vscode-languageclient/lib/implementation';
-import { ProvideTypeDefinitionSignature } from 'vscode-languageclient/lib/typeDefinition';
 import WebRequest = require('web-request');
 import { GoDefinitionProvider } from './goDeclaration';
 import { GoHoverProvider } from './goExtraInfo';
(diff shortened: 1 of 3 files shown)


src/goLanguageServer.ts
```diff
@@ -311,39 +298,43 @@ export function getLanguageServerToolPath(): string {
...
+	// gopls is the default choice.
 	let languageServerOfChoice = 'gopls';
 	if (goConfig['alternateTools']) {
 		const goplsAlternate = goConfig['alternateTools']['gopls'];
+
+		// Check if the user has set the deprecated "go-langserver" setting.
 		const golangserverAlternate = goConfig['alternateTools']['go-langserver'];
-		if (typeof goplsAlternate === 'string') {
+		if (golangserverAlternate) {
+			vscode.window.showErrorMessage(`The "go.alternateTools" setting for "go-langserver" has been deprecated. Please use "gopls" instead.`);
```
hyangah: Do you mean "language-server", or are you suggesting "go-langserver" is not supported at all and switch to use "gopls"? 

And, the user needs to reload the window, right? Mention that in the error message.

stamblerre: Mentioned reloading in the error message.

I realized that the `"go-langserver"` setting was actually not supported in the `package.json` at all, so I figured I'd first start by fully removing it in the code. In a follow-up, I will change the setting to be `"language-server"`, if @ramya-rao-a is okay with that.

src/goLanguageServer.ts
```diff
@@ -291,13 +280,11 @@ export function parseLanguageServerConfig(): LanguageServerConfig {
 }
 
 /**
- * Get the absolute path to the language server to be used.
- * If the required tool is not available, then user is prompted to install it.
- * This supports the language servers from both Google and Sourcegraph with the
- * former getting a precedence over the latter
+ * If the user has enabled the language server, return the absolute path to the
+ * correct binary. If the required tool is not available, prompt the user to
+ * install it. Only gopls (Google) and go-langserver (Sourcegraph) are supported.
```
hyangah: nit: time to remove mention of `go-langserver` in this comment.

stamblerre: Done.

src/goLanguageServer.ts
```diff
@@ -311,39 +298,44 @@ export function getLanguageServerToolPath(): string {
...
-	// Only gopls and go-langserver are supported.
+	// Get the path to the language server binary.
+	const languageServerBinPath = getBinPath(languageServerOfChoice);
+	if (path.isAbsolute(languageServerBinPath)) {
+		return languageServerBinPath;
+	}
+
+	// Installation of gopls and go-langserver is supported.
+	// Other language servers must be installed manually.
 	if (languageServerOfChoice !== 'gopls' && languageServerOfChoice !== 'go-langserver') {
```
ramya-rao-a: Shouldn't we remove `go-langserver` in this check?

src/goTools.ts
```diff
@@ -121,9 +121,14 @@ export function getConfiguredTools(goVersion: GoVersion): Tool[] {
 	// Add the linter that was chosen by the user.
 	maybeAddTool(goConfig['lintTool']);
 
-	// Add the language server for Go versions > 1.10 if user has choosen to do so
+	// Add the language server for Go versions > 1.10 if user has choosen to do so.
+	// Respect the go.alternateTools setting.
 	if (goConfig['useLanguageServer'] && goVersion.gt('1.10')) {
-		maybeAddTool('gopls');
+		if (goConfig['alternateTools']['gopls']) {
+			maybeAddTool(goConfig['alternateTools']['gopls']);
```
ramya-rao-a: Are we trying to show the "alternate" tool in the `Go: Install/Update Tools` command?
At present this would only work if the alternate tool was set to `go-langserver` right?

src/goLanguageServer.ts
stamblerre: Done. I wasn't sure if we still wanted to support installation of go-langserver but I'm happy to remove support for that as well.

src/goTools.ts
stamblerre: That's true. Removed this. I like the idea of completely removing support for go-langserver.

stamblerre:
Closing in favor of https://github.com/golang/vscode-go/pull/2.

PR #3031, opened 2020-02-06
update thirdpartynotices.txt

hyangah:
It was generated based on what license-checker(https://www.npmjs.com/package/license-checker)
detected.

- rm -rf node_modules
- npm ci
- license-checker -json --direct --production --excludePrivatePackages

The generation program is https://gist.github.com/hyangah/26938becbdef9d7ced0d046c7677f1d0

Note: there are modules that do not have valid license files or notices in
the released versions we are depending on. Those dependencies are:

 assert-plus@1.0.0(https://github.com/mcavage/node-assert-plus) MIT 
 better-curry@1.6.0(https://github.com/pocesar/js-bettercurry): MIT
 emitter-listener@1.1.2(https://github.com/othiym23/emitter-listener): BSD-2-Clause
 uri-js@4.2.2(https://github.com/garycourt/uri-js): BSD-2-Clause
 json-schema@0.2.3(https://github.com/kriszyp/json-schema): AFLv2.1 or BSD

### DIFF
--- a/thirdpartynotices.txt
+++ b/thirdpartynotices.txt
@@ -1,15 +1,1305 @@
 THIRD-PARTY SOFTWARE NOTICES AND INFORMATION
 For Microsoft vscode-go
- 
-This project incorporates material from the projects listed below.  The original copyright notice 
-and the license under which Microsoft received such material are set out below. Microsoft reserves 
+
+This project incorporates material from the projects listed below.  The original copyright notice
+and the license under which Microsoft received such material are set out below. Microsoft reserves
 all other rights not expressly granted, whether by implication, estoppel or otherwise.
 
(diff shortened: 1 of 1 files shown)


hyangah:
@ramya-rao-a If there is any standard tool vscode team is using to generate thirdpartynotices file, feel free to ignore this and use that instead. I was trying to mimic the format of vscode's ThirdPartyNotices.txt.

ramya-rao-a:
@roblourens Do you have any recommendations here?

roblourens:
We have a tool that generates this for vscode but it's specific to vscode. I don't know what other extensions do, and I'm not sure what you are required to do...

PR #3041, opened 2020-02-11
Include the link to release note/package overview in the update prompt, and update gopls default version

hyangah:
@stamblerre

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -361,8 +361,8 @@ export async function promptForUpdatingTool(toolName: string) {
 	}
 	const goVersion = await getGoVersion();
 	const updateMsg = `Your version of ${tool.name} appears to be out of date. Please update for an improved experience.`;
-	var choices: string[] = ['Update'];
-	if (toolName == `gopls`) {
+	const choices: string[] = ['Update'];
+	if (toolName === `gopls`) {
 		choices.push('Release Notes');  // TODO(hyangah): pass more info such as version, release note location.
 	}
 	vscode.window.showInformationMessage(updateMsg, ...choices).then((selected) => {
@@ -371,7 +371,7 @@ export async function promptForUpdatingTool(toolName: string) {
 				installTools([tool], goVersion);
 				break;
 			case 'Release Notes':
-				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://github.com/golang/go/issues/33030#issuecomment-510151934'));			
+				vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://github.com/golang/go/issues/33030#issuecomment-510151934'));
 				break;
 			default:
 				declinedUpdates.push(tool);
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -363,19 +363,15 @@ export async function promptForUpdatingTool(toolName: string) {
 	const updateMsg = `Your version of ${tool.name} appears to be out of date. Please update for an improved experience.`;
 	var choices: string[] = ['Update'];
 	if (toolName == `gopls`) {
-		choices.push('Release Note');  // TODO(hyangah): pass more info such as version, release note location.
+		choices.push('Release Notes');  // TODO(hyangah): pass more info such as version, release note location.
 	}
 	vscode.window.showInformationMessage(updateMsg, ...choices).then((selected) => {
 		switch (selected) {
 			case 'Update':
 				installTools([tool], goVersion);
 				break;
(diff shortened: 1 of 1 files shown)


stamblerre:
Can you add a screenshot of how this will look?

src/goLanguageServer.ts
```diff
@@ -376,8 +376,8 @@ function registerUsualProviders(ctx: vscode.ExtensionContext) {
 	vscode.workspace.onDidChangeTextDocument(parseLiveFile, null, ctx.subscriptions);
 }
 
-const defaultLatestVersion = semver.coerce('0.1.7');
-const defaultLatestVersionTime = moment('2019-09-18', 'YYYY-MM-DD');
+const defaultLatestVersion = semver.coerce('0.3.1');
+const defaultLatestVersionTime = moment('2020-02-04', 'YYYY-MM-DD');
```
stamblerre: @ramya-rao-a: Can we get away from having this hardcoded default as a backup?

hyangah:
> Can you add a screenshot of how this will look?

image

And once you click the Release Note, it will open a browser with the provided link. I wish I knew a way to let the prompt stay after clicking the link, so users can come back and click the Update button.

stamblerre:
Thanks! That looks really good.
Maybe to avoid the disappearing window, we can have the link to the release notes pop up after the user has already clicked update? Or @ramya-rao-a might know a better way.

src/goInstallTools.ts
```diff
@@ -355,11 +355,22 @@ export async function promptForUpdatingTool(toolName: string) {
 	}
 	const goVersion = await getGoVersion();
 	const updateMsg = `Your version of ${tool.name} appears to be out of date. Please update for an improved experience.`;
-	vscode.window.showInformationMessage(updateMsg, 'Update').then((selected) => {
+	var choices: string[] = ['Update'];
+	if (toolName == `gopls`) {
+		choices.push('Release Note');  // TODO(hyangah): pass more info such as version, release note location.
```
stamblerre: I think this should probably be "Release Notes" - I checked, and when there is a new version of VS Code, the title of the page that pops up is "Release Notes: <version>".

src/goLanguageServer.ts
ramya-rao-a: We had this for the case when user doesnt want the extension to keep checking with the proxy server to determine if an update is needed. Removing the default would mean that such users will never get the prompt to update `gopls`

src/goInstallTools.ts
ramya-rao-a: I agree

src/goInstallTools.ts
```diff
@@ -355,11 +355,22 @@ export async function promptForUpdatingTool(toolName: string) {
...
+	}
+	vscode.window.showInformationMessage(updateMsg, ...choices).then((selected) => {
 		switch (selected) {
 			case 'Update':
 				installTools([tool], goVersion);
 				break;
+			case 'Release Note':
+				if (toolName == `gopls`) {
+					vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://github.com/golang/go/issues/33030#issuecomment-510151934'));			
+				} else {
```
ramya-rao-a: A few lines above, we add the `Release Note` option only when the tool is `gopls`. Therefore, this else block will never be reached

ramya-rao-a:
The update to gopls default version is ported to master branch via https://github.com/microsoft/vscode-go/commit/36f0b5815f4000db94797ca36741ae91411d9f55 and is part of 0.13.1 version of the extension

src/goInstallTools.ts
hyangah: done.

src/goInstallTools.ts
hyangah: you're right. removed.

ramya-rao-a:
There are a few linting errors, can you fix them?
Try `npm run fix-lint`

Screen Shot 2020-02-28 at 5 33 56 AM

src/goInstallTools.ts
```diff
@@ -361,8 +361,8 @@ export async function promptForUpdatingTool(toolName: string) {
 	}
 	const goVersion = await getGoVersion();
 	const updateMsg = `Your version of ${tool.name} appears to be out of date. Please update for an improved experience.`;
-	var choices: string[] = ['Update'];
-	if (toolName == `gopls`) {
+	const choices: string[] = ['Update'];
+	if (toolName === `gopls`) {
```
hyangah: Ooops, embarrassing.
Thank you for the fix.

PR #3045, opened 2020-02-13
goImports: avoid collapsing new imports into pseudo import line

### LINKED ISSUE #1701: automatic import adder should not separate first import from its comment
0. Relevant Settings:

       "go.autocompleteUnimportedPackages": true

1. Type the following program

```
package main

//#include <stdlib.h>
import "C"

func main() {
    fmt
}
```

2. Place cursor behind "fmt"
3. Type "."  The Intellisense window should pop up with suggestions.
4. Accept any suggestion (e.g. "Errorf")
5. The import statement will be rewritten to

```
//#include <stdlib.h>
import (
	"fmt"
	"C"
)
```

However this breaks the magic of import "C" which uses the preceding comment.

When converting import lines to a parenthesized import block, vscode-go should insert the "import (" before the comment lines immediately preceding (i.e. with no empty lines in between) the first import (if there are any).

hyangah:
getTextEditForAddImport tries to insert a newly added package
into an existing import group. If no import group exists, it
tries to merge single line import statements and create a group.

In this process, import statements like

  import "C"

are pseudo imports and shouldn't be grouped with other imports.
This CL changes parseFilePrelude to detect such pseudo imports
and excludes such imports from the grouping logic.

Fixes #1701


hyangah:
the test failures on platforms other than mac are real - I messed up with file name (case sensitivity). Will fix it when I can access my linux tomorrow.

src/util.ts
```diff
@@ -156,11 +156,12 @@ export function parseFilePrelude(text: string): Prelude {
 		}
 		if (line.match(/^(\s)*import(\s)+\(/)) {
 			ret.imports.push({ kind: 'multi', start: i, end: -1, pkgs: [] });
-		}
-		if (line.match(/^(\s)*import(\s)+[^\(]/)) {
+		} else if (line.match(/^\s*import\s+"C"/)) {
```
ramya-rao-a, line 159: Is C the only import that falls under this pseudo category?

hyangah: Yes. "C" is the only one in this category (psuedo import whose placement in the import affects the correctness/functionality)

PR #3047, opened 2020-02-13
travis-ci: prepare for 1.14 release

hyangah:
remove 1.11


hyangah:
do not review yet - 1.14 is still in rc, and it is not available in travis.

hyangah:
1.14 is released, so it should be working now.

.travis.yml
```diff
@@ -47,6 +47,7 @@ install:
   - go get -u -v github.com/davidrjenni/reftools/cmd/fillstruct
   - go get -u -v github.com/haya14busa/goplay/cmd/goplay
   - go get -u -v github.com/mdempsky/gocode
+  - go get -u -v github.com/stamblerre/gocode
```
ramya-rao-a: Won't installing both the `gocode` result in one overriding the other?

Issue #3050, opened 2020-02-14
Cannot install gopkgs tool


jedebarbieri:
Hello, I can't install gopkgs and the "auto-completion" and "go to definition" feature are not working.

- go version go1.13.7 windows/amd64
- VS Code : 1.42.1
c47d83b293181d9be64f27ff093689e8e7aed054
x64
- VS Code Go extension : v0.13.0
- windows / amd64
- settings.json: 
`"go.formatTool": "goimports",`

1. Go to "GO: Install/Update Tools"
2. Select "gopkgs"
3. Wait until install finish. It shows me this error:
```
go.toolsGopath setting is not set. Using GOPATH C:\GoPath
Installing 1 tool at C:\GoPath\bin in module mode.
  gopkgs

Installing github.com/uudashr/gopkgs/cmd/gopkgs FAILED

1 tools failed to install.

gopkgs:
Error: Command failed: C:\Go\bin\go.exe get -v github.com/uudashr/gopkgs/cmd/gopkgs
go: finding github.com/uudashr/gopkgs/cmd/gopkgs latest
go get: github.com/uudashr/gopkgs/cmd/gopkgs@v0.0.0-20191024034442-58e9141cd7d6 requires
	github.com/uudashr/gopkgs/v2@v2.1.0 requires
	github.com/uudashr/gopkgs@v0.0.0: reading github.com/uudashr/gopkgs/go.mod at revision v0.0.0: unknown revision v0.0.0
go: finding github.com/uudashr/gopkgs/cmd/gopkgs latest
go get: github.com/uudashr/gopkgs/cmd/gopkgs@v0.0.0-20191024034442-58e9141cd7d6 requires
	github.com/uudashr/gopkgs/v2@v2.1.0 requires
	github.com/uudashr/gopkgs@v0.0.0: reading github.com/uudashr/gopkgs/go.mod at revision v0.0.0: unknown revision v0.0.0
```
And a message "Analysis Tools Missing" is shown in the right bottom side:
![image](https://user-images.githubusercontent.com/49592738/74538073-67067d80-4f09-11ea-99df-7a730ae1c126.png)
And when I press "Install" button the process is repeated.

stamblerre:
I think you're running into this issue: https://github.com/uudashr/gopkgs/issues/25. You can try downloading manually with modules turned off (`GO111MODULE=off go get github.com/uudashr/gopkgs`), but you may not get the correct version.

nezorflame:
[...]

PR #3057, opened 2020-02-18
Update gopkgs

nezorflame:
Closes https://github.com/microsoft/vscode-go/issues/3050
See https://github.com/uudashr/gopkgs/issues/25


nezorflame:
This import path works with modules being both on and off since the repo adheres to the [Major subdirectory](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher) approach.

nezorflame:
@ramya-rao-a hi! Any chance this can be merged? I assume the pipeline failure is due to some previous change.

PR #3064, opened 2020-02-24
Update to latest vscode-languageclient

hyangah:
This reverts commit 5f15c237c71e527d983d29d351597838868c4b1b.

This effectively brings in the commit 3b02de0 back again.
Originally reverted because it requires vscode 1.41.x,
which was just released at the time of Go extension 0.13.0
release. Currently 1.42.x is the latest version, and we expect
another cycle of vscode release when the next version of Go
extension is out.

We need more up-to-date version of vscode-languageclient to
bring in various fixes to problems such as
https://github.com/microsoft/vscode-languageserver-node/issues/442

Considered only minor version upgrade but I couldn't find a suitable
newer version than 5.2.1
(https://www.npmjs.com/package/vscode-languageclient)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1480,39 +1480,32 @@
       }
     },
     "vscode-jsonrpc": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/vscode-jsonrpc/-/vscode-jsonrpc-4.0.0.tgz",
-      "integrity": "sha512-perEnXQdQOJMTDFNv+UF3h1Y0z4iSiaN9jIlb0OqIYgosPCZGYh/MCUlkFtV2668PL69lRDO32hmvL2yiidUYg=="
+      "version": "5.0.1",
+      "resolved": "https://registry.npmjs.org/vscode-jsonrpc/-/vscode-jsonrpc-5.0.1.tgz",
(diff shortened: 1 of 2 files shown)


stamblerre:
I think you will need to increase the value of the VS Code engine [here](https://github.com/microsoft/vscode-go/blob/master/package.json#L69) to `^1.42.0`.

ramya-rao-a:
The latest version of the language client seems to need 1.42 where as the previous version needed only 1.41

Am ok changing our min supported version of VS Code to 1.41, but am hesitant to change it to 1.42 without giving some time for users to upgrade. https://github.com/microsoft/vscode-go/commit/7d0069a130f11c837b59311afd225cc041703fba reflects the change to do the above.

Side note: https://github.com/microsoft/vscode-languageserver-node/commit/f5542d9f912e5a54b6bc81ae2df0d12cf7ee9082#r37530736

ramya-rao-a:
Closing this PR for now
In the next month, we can go ahead and change the semver for the language client to `^6.0.1` and our min VS Code version to 1.42 if needed

ramya-rao-a:
The [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) 0.14.0-beta.1 is using version 6.1.0 of the language client.

Please use this to test out the fixes from the language client

PR #3079, opened 2020-02-28
Map remote go module cache to local module cache

fnmunhoz:
This is similar to another scenario: https://github.com/microsoft/vscode-go/pull/1179

When using the remote debug feature, the paths to remote go module cache might not match the paths to local go module cache.

The idea here is to check if the remote path includes the go module cache sub string, for example: 

`/go/pkg/mod/github.com/account/package`

In this case, it would be a match for the go module cache substring: 

`/pkg/mod/`

And in this case, the path would be replaced by: 

`/Users/someuser/go/pkg/mod/github.com/account/package`

Considering a local gopath like this: 

`$GOPATH="/Users/someuser/go"`.

By doing that, vscode-go can find all source files during debug, including cached modules.

I've made the minimal required changes to make it work for the scenario that I was facing, but I'm happy to make improvements based on feedback.

If I missed any guidelines, please let me know.

Also, I'm curious to know if this is the right approach to deal with this, or there is a better alternative.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -854,6 +854,21 @@ class GoDebugSession extends LoggingDebugSession {
 			if (goroot && index > 0) {
 				return path.join(goroot, pathToConvert.substr(index));
 			}
+
+			const indexGoModCache = pathToConvert.indexOf(
+				`${this.remotePathSeparator}pkg${this.remotePathSeparator}mod${this.remotePathSeparator}`
+			);
+			const gopath = process.env['GOPATH'];
+
+			if (gopath && indexGoModCache > 0) {
+				return path.join(
+					gopath,
+					pathToConvert
+						.substr(indexGoModCache)
+						.split(this.remotePathSeparator)
+						.join(this.localPathSeparator)
+				);
+			}
 		}
 		return pathToConvert
 			.replace(this.delve.remotePath, this.delve.program)
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -846,9 +846,9 @@ class GoDebugSession extends LoggingDebugSession {
 			return this.convertDebuggerPathToClient(pathToConvert);
 		}
 
-		// Fix for https://github.com/Microsoft/vscode-go/issues/1178
-		// When the pathToConvert is under GOROOT, replace the remote GOROOT with local GOROOT
+		// When the pathToConvert is under GOROOT or Go module cache, replace path appropriately
 		if (!pathToConvert.startsWith(this.delve.remotePath)) {
+			// Fix for https://github.com/Microsoft/vscode-go/issues/1178
 			const index = pathToConvert.indexOf(`${this.remotePathSeparator}src${this.remotePathSeparator}`);
 			const goroot = process.env['GOROOT'];
 			if (goroot && index > 0) {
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Thanks for the PR @fnmunhoz 

@quoctruong, Can you verify the fix here?

quoctruong:
LGTM

src/debugAdapter/goDebug.ts
```diff
@@ -854,6 +854,21 @@ class GoDebugSession extends LoggingDebugSession {
...
 			}
+
+			const indexGoModCache = pathToConvert.indexOf(
+				`${this.remotePathSeparator}pkg${this.remotePathSeparator}mod${this.remotePathSeparator}`
+			);
+			const gopath = process.env['GOPATH'];
+
+			if (gopath && indexGoModCache > 0) {
+				return path.join(
+					gopath,
```
ramya-rao-a: `gopath` can potentially have multiple paths with the path separator specific to the OS platform. In that case, we can't use it in `path.join` directly this way.

ramya-rao-a: Does /pkg/mod apply to the first path in the gopath? Is there a rule around this?
Also, we need to ensure the right path separators are being used like we did in #3108

cc @quoctruong

stamblerre: > Does /pkg/mod apply to the first path in the gopath? Is there a rule around this?

Yes. I'm not sure where this is specified, but you can see `goimports` relying on this fact [here](https://github.com/golang/tools/blob/0d839f3cf2ed7d63fd98875ea3589e5bf4b64de7/internal/imports/mod.go#L85).

ramya-rao-a: Thanks @stamblerre 

@fnmunhoz, @quoctruong I have pushed a commit to use the first go path, can either of you test this one more time?

ramya-rao-a:
@fnmunhoz Thanks for your first PR contribution to this project. Can you please sign the CLA as requested in the above comment?

src/debugAdapter/goDebug.ts
```diff
@@ -846,14 +846,29 @@ class GoDebugSession extends LoggingDebugSession {
...
+			);
+			const gopath = (process.env['GOPATH'] || '').split(path.delimiter)[0];
+
+			if (gopath && indexGoModCache > 0) {
+				return path.join(
+					gopath,
+					pathToConvert
+						.substr(indexGoModCache)
+						.split(this.remotePathSeparator)
+						.join(this.localPathSeparator)
```
quoctruong: I just thought of one problem with this logic. Even if you are replacing the local path separator by the remote path separator, `path.join` will [normalize the path ](https://nodejs.org/api/path.html#path_path_join_paths) according to the platform. This means that you probably have to do the replacing part AFTER joining the path?

ramya-rao-a: We will have to do similar changes like you did to `toDebuggerPath` here as well. I was hoping to do that outside of this PR and have this PR focus only on the module cache part. Because the change has to applied to the goroot workaround and to the return statement a few lines below

quoctruong: That sounds good.

fnmunhoz:
Hey @ramya-rao-a sorry for the really long delay.

I've signed the CLA. Not sure if this PR is still valid based on the comments above, but feel free to go ahead with it if that's the case.

Thanks!

src/debugAdapter/goDebug.ts
fnmunhoz: @ramya-rao-a sure I'll try to do it today and let you know. Thanks.

fnmunhoz: Hey @ramya-rao-a I just double checked the extension after the change you made and it works fine for me, I was able to debug all files, including libraries inside go mod cache.

Thanks!

ramya-rao-a:
Thanks @fnmunhoz!

fnmunhoz:
Yay! Thank you @ramya-rao-a

ramya-rao-a:
The latest update 0.14.2 has this change, thanks @fnmunhoz!

fnmunhoz:
Good to know! Thanks @ramya-rao-a 😃

PR #3084, opened 2020-03-03
goLanguageServer: set completion follow up command from middleware

### LINKED ISSUE #3075: How to disable parameter hints post completion when using gopls
### What version of Go, VS Code & VS Code Go extension are you using?
- Run `go version` to get version of Go
	- 1.14
- Run `code -v` or `code-insiders -v` to get version of VS Code or VS Code Insiders
	- 1.42.1
- Check your installed extensions to get the version of the VS Code Go extension 
	- 0.13.1
- Run `go env GOOS GOARCH` to get the operating system and processor arhcitecture details
	- linux amd64

### Share the Go related settings you have added/edited

```
{
    "editor.parameterHints.enabled": false,
    "go.useLanguageServer": true
}
```

### Describe the bug
editor.parameterHints.enabled does not work,I want to disable parameter hints.
QQ截图20200228143509

hyangah:
Middleware checks both `editor.parameterHints.enabled` and
`[go].editor.parameterHints.enabled`. The language specific
config takes precedence as long as it is defined.

vscode.workspace.getConfiguration behaves differently when
querying the language-specific configuration section ('[go]')
so I couldn't query [go].editor.parameterHints directly using
getConfiguration.

Tested manually. (sorry)

Fixes #3075

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -12,6 +12,7 @@ import semver = require('semver');
 import util = require('util');
 import vscode = require('vscode');
 import {
+	Command,
 	FormattingOptions,
 	HandleDiagnosticsSignature,
 	LanguageClient,
@@ -26,8 +27,7 @@ import {
 	ProvideRenameEditsSignature,
 	ProvideSignatureHelpSignature,
 	ProvideWorkspaceSymbolsSignature,
-	RevealOutputChannelOn,
-	Command
+	RevealOutputChannelOn
 } from 'vscode-languageclient';
 import { ProvideImplementationSignature } from 'vscode-languageclient/lib/implementation';
 import { ProvideTypeDefinitionSignature } from 'vscode-languageclient/lib/typeDefinition';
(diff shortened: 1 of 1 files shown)


hyangah:
@stamblerre @ramya-rao-a

src/goLanguageServer.ts
```diff
@@ -143,6 +144,44 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
 					return next(document, token);
+				},
+				provideCompletionItem: (
+					document: vscode.TextDocument,
+					position: vscode.Position,
+					context: vscode.CompletionContext,
+					token: vscode.CancellationToken,
+					next: ProvideCompletionItemsSignature
+				) => {
+						const editorParamHintsEnabled = vscode.workspace.getConfiguration('editor.parameterHints', document.uri)['enabled'];
```
stamblerre: Are these both necessary? I would think that the purpose of the `document.uri` parameter is to decide if the setting applies to the current file.

src/goLanguageServer.ts
```diff
@@ -143,6 +144,44 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+						const goParamHintsEnabled = vscode.workspace.getConfiguration('[go]', document.uri)['editor.parameterHints.enabled'];
+
+						var paramHintsEnabled: boolean = false;
+						if (typeof goParamHintsEnabled === 'undefined') {
+							paramHintsEnabled = editorParamHintsEnabled;
+						} else {
+							paramHintsEnabled = goParamHintsEnabled;
+						}
+						var cmd: Command = undefined;
```
stamblerre: is the `= undefined` part necessary?

src/goLanguageServer.ts
```diff
@@ -143,6 +144,44 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+							r: vscode.CompletionItem[] | vscode.CompletionList | null | undefined):
+							vscode.CompletionItem[] | vscode.CompletionList | null | undefined {
+							if (r) {
+								(Array.isArray(r) ? r : r.items).forEach((i: vscode.CompletionItem) => {
+									i.command = cmd;
+								});
+							}
+							return r;
+						}
+						const isThenable = <T>(obj: vscode.ProviderResult<T>): obj is Thenable<T> => obj && (<any>obj)['then'];
```
stamblerre: What does this do? I'm totally confused at this point :smile:

hyangah: vscode.ProviderResult can be a simple T or Thenable<T>. So, if the result of `next` call is a promise (thenable that has 'then' property), we chain configureCommands call using 'then'.
If the result of `next` is not thenable, it's not a promise and there is no 'then'. So, simply invoke configureCommands and pass the result of the next.

src/goLanguageServer.ts
hyangah: That's what I thought but it didn't work that way. Good news is that [vscode 1.42 api](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42169/commits/15e3d9ed635760023f0cacdbe0d6710a4c814673) has ConfigurationScope that includes {languageId, uri} pair. If we use it, getConfiguration does the proper job (applying the language specific config on top). Added TODO.

src/goLanguageServer.ts
```diff
@@ -143,6 +144,55 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+					let cmd: Command;
+					if (paramHintsEnabled) {
+						cmd = { title: 'triggerParameterHints', command: 'editor.action.triggerParameterHints' };
+					}
+
+					function configureCommands(
+						r: vscode.CompletionItem[] | vscode.CompletionList | null | undefined):
+						vscode.CompletionItem[] | vscode.CompletionList | null | undefined {
+						if (r) {
+							(Array.isArray(r) ? r : r.items).forEach((i: vscode.CompletionItem) => {
```
ramya-rao-a: The `configureCommands()` function is typed in such a way that it takes null and undefined as well.
In those cases, you will get a TypeError as `Array.isArray(r)` will be false and there is an attempt to get to the `items` property

src/goLanguageServer.ts
```diff
@@ -143,6 +144,55 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+								i.command = cmd;
+							});
+						}
+						return r;
```
ramya-rao-a: Since all we want to do is to update a property on the completion items, we shouldn't need to return anything. The array passed in should be updated and ready to use

src/goLanguageServer.ts
```diff
@@ -143,6 +144,55 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
...
+					}
+					const ret = next(document, position, context, token);
+
+					const isThenable = <T>(obj: vscode.ProviderResult<T>): obj is Thenable<T> => obj && (<any>obj)['then'];
+					if (isThenable<vscode.CompletionItem[] | vscode.CompletionList | null | undefined>(ret)) {
```
ramya-rao-a: Why are we expecting `ret` to be a thenable resolving to null or undefined?

src/goLanguageServer.ts
hyangah: but that's guarded by the above if (r), isn't it?

src/goLanguageServer.ts
hyangah: vscode.ProviderResult (the return type of next) allows null or undefined.

src/goLanguageServer.ts
ramya-rao-a: Ah! yes. Sorry, missed that

ramya-rao-a:
If I understood the upstream conversations, it looks like `gopls` already calls `editor.action.triggerParameterHints` command. Won't we end up calling this command twice with the changes in this PR? 

Are we planning to remove the call to the command from `gopls`?

stamblerre:
> Are we planning to remove the call to the command from `gopls`?

Yes. This will handle the fact that we can't respect user's configurations in `gopls`.

ramya-rao-a:
My concern is for users that have this change in the extension, but uses the gopls that is also running the `editor.action.triggerParameterHints` command. So, wouldn't this command get triggered twice?

hyangah:
> My concern is for users that have this change in the extension, but uses the gopls that is also running the `editor.action.triggerParameterHints` command. So, wouldn't this command get triggered twice?

There is only one command in each completion item and this PR is overriding it. 
So, basically, it's undoing what `gopls` did in line 183.

PR #3091, opened 2020-03-06
telemetry.ts: send telemetry only if aiKey is not empty

hyangah:
Change-Id: I7157be16784884cb2b6f5ad1a4f82846cba642c6

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -183,6 +183,9 @@ function sendTelemetryEvent(
 	properties?: { [key: string]: string },
 	measures?: { [key: string]: number }
 ): void {
+	if (!aiKey) {
+		return;  // cannot enable telemetry
+	}
 	telemtryReporter = telemtryReporter
 		? telemtryReporter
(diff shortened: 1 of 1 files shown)


hyangah:
@stamblerre

stamblerre:
LGTM - thanks!

src/telemetry.ts
```diff
@@ -183,6 +183,9 @@ function sendTelemetryEvent(
 	properties?: { [key: string]: string },
 	measures?: { [key: string]: number }
 ): void {
+	if (!aiKey) {
```
ramya-rao-a: The `aiKey` is initialized in line 12 and is a constant.
Therefore, in its current state, the extension will not reach a state where this will be empty..

stamblerre: https://github.com/golang/vscode-go/commit/ada08587033b59b41f093fc39f8890350d2dbacd#diff-499709399145585d79806a7415935b15L12 is the reasoning behind this.

PR #3070, opened 2020-02-26
Improve the bug report workflow

hyangah:
1) update the issue template to provide extra information about `go env` and `gopls` 
2) a new report issue command that captures necessary information and open an issue filled with the information.

PR #3095, opened 2020-03-12
bug_report.md: Fix "architecture" typo

darkvertex:
It's a tiny thing, I know, but I couldn't help noticing it. lol


stamblerre:
Thanks for the contribution! LGTM, though I think it might be more helpful if we asked users for the full output of `go env`. @ramya-rao-a: What do you think?

ramya-rao-a:
We can discuss using the full out of `go env` in #3070

Issue #2010, opened 2018-10-15
Breakpoints set in VS Code not hit (docker remote debugging)


peteat:
I can't seem to hit breakpoints set in VS Code when remotely debugging my go app (which is running in Docker).

I've set up my launch.json:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Remote debug in Docker",
            "type": "go",
            "request": "launch",
            "mode": "remote",
            "program": "${workspaceRoot}",
            "env": {},
            "args": [],
            "remotePath": "${workspaceRoot}",
            "port": 40000, // Port 
            "host": "127.0.0.1", // Docker IP 
            "showLog": true,
            "trace": "verbose"
        }
    ]
}
```

This is my `go.delveConfig` (user settings in VS Code):
```
{
    "useApiV1": false,
    "dlvLoadConfig": {
      "followPointers": true,
      "maxVariableRecurse": 3,
      "maxStringLen": 400,
      "maxArrayValues": 400,
      "maxStructFields": -1
    }
}
```

It attaches fine, and even hits breakpoints set via `runtime.Breakpoint()` but will not hit any breakpoints set via VS Code. Debugging the same project in GoLand works fine, I can set and hit breakpoints with no issue.

I have done a lot of googling, and checking documentation, playing with configurations (launch.json, dockerfile, dlv etc) but can't seem to find a fix. Not sure what's wrong, but I have noticed a difference in the VS Code CreateBreakpoint request compared to the GoLand one by looking at the dlv log output (by adding `--log --log-output=rpc`).

Here's the dlv log (via docker) when debugging via GoLand:
```
testapp_1  | time="2018-10-15T04:59:46Z" level=debug msg="<- RPCServer.CreateBreakpoint(rpc2.CreateBreakpointIn{\"Breakpoint\":{\"id\":0,\"name\":\"\",\"addr\":0,\"file\":\"C:/Users/<username>/Dev/go/src/testApp/main.go\",\"line\":7,\"Cond\":\"\",\"continue\":false,\"goroutine\":false,\"stacktrace\":0,\"LoadArgs\":null,\"LoadLocals\":null,\"hitCount\":null,\"totalHitCount\":0}})" layer=rpc
[...]

PR #3108, opened 2020-03-17
Fix remote path bug (#2010)

quoctruong:
This is to fix https://github.com/microsoft/vscode-go/issues/2010. When we are remote debugging cross-compiled binary on Windows, we are not respecting the separator that users use in `remotePath` field.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -835,10 +835,9 @@ class GoDebugSession extends LoggingDebugSession {
 		if (this.delve.remotePath.length === 0) {
 			return this.convertClientPathToDebugger(filePath);
 		}
-		// When the filePath has a different path separator
-		// than the local path separator (cross-compilation),
-		// the split and join logic won't work.
-		// See github.com/microsoft/vscode-go/issues/2010.
+		// The filePath may have a different path separator than the localPath
+		// So, update it to use the same separator as the remote path to ease
+		// in replacing the local path in it with remote path
 		filePath = filePath.replace(/\/|\\/g, this.remotePathSeparator);
 		return filePath
 			.replace(this.delve.program.replace(/\/|\\/g, this.remotePathSeparator), this.delve.remotePath);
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -754,13 +761,6 @@ class GoDebugSession extends LoggingDebugSession {
 		log('InitializeResponse');
 	}
 
-	protected findPathSeperator(filePath: string) {
-		if (/^(\w:[\\/]|\\\\)/.test(filePath)) {
-			return '\\';
-		}
-		return filePath.includes('/') ? '/' : '\\';
```
hyangah: Is there a possibility that the path has both / and \? The new change selects / in that case while the old findPathSeparator picks \  if it matches the above pattern (in line 758). Wonder if that would be potentially a problem.

src/debugAdapter/goDebug.ts
```diff
@@ -835,10 +835,13 @@ class GoDebugSession extends LoggingDebugSession {
 		if (this.delve.remotePath.length === 0) {
 			return this.convertClientPathToDebugger(filePath);
 		}
+		// When the filePath has a different path separator
+		// than the local path separator (cross-compilation),
+		// the split and join logic won't work.
+		// See github.com/microsoft/vscode-go/issues/2010.
+		filePath = filePath.replace(/\/|\\/g, this.remotePathSeparator);
```
hyangah: wonder if it's doable to write some unit tests to check these path manipulation functions.

src/debugAdapter/goDebug.ts
```diff
@@ -835,10 +835,13 @@ class GoDebugSession extends LoggingDebugSession {
 		if (this.delve.remotePath.length === 0) {
 			return this.convertClientPathToDebugger(filePath);
 		}
+		// When the filePath has a different path separator
+		// than the local path separator (cross-compilation),
+		// the split and join logic won't work.
+		// See github.com/microsoft/vscode-go/issues/2010.
```
ramya-rao-a: The below manipulation is also required when the user provided `program` uses a different path separator than what VS Code passes to the debug adapter.

```suggestion
		// The filePath may have a different path separator than the localPath
		// So, update it to use the same separator as the remote path to ease
		// in replacing the local path with remote path
```

PR #3117, opened 2020-03-20
test/integration: use default GOPATH when the env var is not set

hyangah:
If the env var GOPATH is not set when the test runs,
now it runs `go env GOPATH` to retrieve the default value
the underlying go commands will use. I wished I could make
this command run asynchronously, but I am not familiar with
the test framework and failed to make it working with
async functions. So, like many other file operations used in
the test setup phase, I am using execFileSync.

Update  golang/vscode-go#6

Change-Id: Id075827ab8ebd955f7a86460aaa7ea4e0e7cd183
GitHub-Last-Rev: 62221580ddaae0062ec79a3ab3efc6258c912117
GitHub-Pull-Request: golang/vscode-go#8
Reviewed-on: https://go-review.googlesource.com/c/vscode-go/+/223177
Reviewed-by: Andrew Bonventre <andybons@golang.org>

### DIFF
--- a/test/integration/extension.test.ts
+++ b/test/integration/extension.test.ts
@@ -19,6 +19,7 @@ import {
 	generateTestCurrentPackage
 } from '../../src/goGenerateTests';
 import { getTextEditForAddImport, listPackages } from '../../src/goImport';
+import { updateGoPathGoRootFromConfig } from '../../src/goInstallTools';
 import { documentSymbols, GoDocumentSymbolProvider, GoOutlineImportsOptions } from '../../src/goOutline';
 import { getAllPackages } from '../../src/goPackages';
 import { goPlay } from '../../src/goPlayground';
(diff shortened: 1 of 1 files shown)


test/integration/extension.test.ts
```diff
@@ -36,9 +36,29 @@ import {
...
+	return gopath;
+}
+
 suite('Go Extension Tests', function() {
 	this.timeout(10000);
-	const gopath = getCurrentGoPath();
+	let gopath = getCurrentGoPath();
+	if (!gopath) {
+			gopath = queryDefaultGopathSync();
+	}
```
ramya-rao-a: Try the below after updating the callback function to be async

```suggestion
	await updateGoPathGoRootFromConfig();
```

hyangah: That's what I tried before but failed to convince the test suite to wait - the test just terminates 
 without running actual tests and declares the success. Need help with the Mocha suite setup.

zmb3: `suite()` cannot accept an async function, so you'll need to declare `gopath`, `repoPath`, `fixturePath`, etc. with `let` instead of `const`, and assign them in `suiteSetup()`, which can take async functions.

hyangah: Thanks @zmb3. Made the suggested change.

PR #3119, opened 2020-03-22
Keep the last used cover profile path

### LINKED ISSUE #3118: Keep last used path when using coverprofile
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I use the "Apply Cover Profile" prompt:
- Each time I want to show the coverage of my source code, I have to type the whole path to the cover profile file.
- If I make a typo when typing the path to my cover profile, I get the error `Cannot find the file [path with typo]` and I have to type the whole path another time.

**Describe the solution you'd like**
It could be really cool to keep the last entered value even if the user entered a wrong value to allow him to fix the path without having to retype it again. 

**Describe alternatives you've considered**
Keeping the path in Alfred ... but that's not really an alternative.

**Additional context**
I'm using **Visual Studio Code Remote - SSH** that's why I need to type the full path of the cover profile.

alexandrevilain:
Signed-off-by: Alexandre Vilain <alexandre.vilain@me.com>

Fixes #3118

### DIFF
--- a/src/goMain.ts
+++ b/src/goMain.ts
@@ -480,19 +480,22 @@ export function activate(ctx: vscode.ExtensionContext): void {
 				return;
 			}
 			const lastCoverProfilePathKey = 'lastCoverProfilePathKey';
+			const lastCoverProfilePath = getFromWorkspaceState(lastCoverProfilePathKey, '');
 			vscode.window
 				.showInputBox({
 					prompt: 'Enter the path to the coverage profile for current package',
-					value: getFromWorkspaceState(lastCoverProfilePathKey, ''),
+					value: lastCoverProfilePath,
 				})
 				.then((coverProfilePath) => {
 					if (!coverProfilePath) {
 						return;
 					}
-					updateWorkspaceState(lastCoverProfilePathKey, coverProfilePath);
 					if (!fileExists(coverProfilePath)) {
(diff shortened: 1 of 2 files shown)


src/goMain.ts
```diff
@@ -471,14 +479,17 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
 				.showInputBox({
-					prompt: 'Enter the path to the coverage profile for current package'
+					prompt: 'Enter the path to the coverage profile for current package',
+					value: getFromWorkspaceState(lastCoverProfilePathKey, ''),
 				})
 				.then((coverProfilePath) => {
 					if (!coverProfilePath) {
 						return;
 					}
+					updateWorkspaceState(lastCoverProfilePathKey, coverProfilePath);
```
ramya-rao-a: We can skip updating the state if the same value was used

ramya-rao-a:
The changes here are available in the [latest beta version](https://github.com/microsoft/vscode-go/wiki/Use-the-beta-version-of-the-latest-Go-extension) of this extension.

alexandrevilain:
Hi @ramya-rao-a ! Thanks :)

ramya-rao-a:
This change is now available in the latest version of this extension.

Thanks @alexandrevilain!

PR #3127, opened 2020-03-25
Remove support for github.com/sourcegraph/go-langserver

stamblerre:
PR in Go Nightly: https://github.com/golang/vscode-go/pull/2

I improved the error messages for the `go-langserver` alternate tools setting, but I figured that it might be too much effort to add buttons to make automatic changes for users. I would imagine that most people have already made the switch, and if not, it's quite a small amount of manual work to justify the extra code for different buttons. Once a user changes the setting and reloads their editor, they will be prompted to install `gopls` anyway.

@ramya-rao-a: If you disagree, I'd be happy to add the buttons anyway. If I do that, could you point me to an API reference for how the buttons would work? I got as far as adding items to `vscode.window.showErrorMessage`, but I couldn't figure out how to connect them to functions when they are clicked or how to programmatically modify a user's settings.

Fixes https://github.com/microsoft/vscode-go/issues/3028

/cc @hyangah

### DIFF
--- a/src/goInstallTools.ts
+++ b/src/goInstallTools.ts
@@ -277,7 +277,7 @@ export function installTools(missing: Tool[], goVersion: GoVersion): Promise<voi
 			outputChannel.appendLine(''); // Blank line for spacing
 			const failures = res.filter((x) => x != null);
 			if (failures.length === 0) {
-				if (containsString(missing, 'go-langserver') || containsString(missing, 'gopls')) {
+				if (containsString(missing, 'gopls')) {
 					outputChannel.appendLine('Reload VS Code window to use the Go language server');
 				}
 				outputChannel.appendLine('All tools successfully installed. You are ready to Go :).');
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -344,10 +344,10 @@ export function parseLanguageServerConfig(): LanguageServerConfig {
 }
 
 /**
- * Get the absolute path to the language server to be used.
- * If the required tool is not available, then user is prompted to install it.
- * This supports the language servers from both Google and Sourcegraph with the
- * former getting a precedence over the latter
+ *
+ * If the user has enabled the language server, return the absolute path to the
+ * correct binary. If the required tool is not available, prompt the user to
+ * install it. Only gopls is officially supported.
  */
 export function getLanguageServerToolPath(): string {
 	// If language server is not enabled, return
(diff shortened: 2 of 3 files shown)


ramya-rao-a:
On second thoughts, I am wondering if we should skip the trouble and simplify as below

```js
// Get the path to gopls or any alternative that the user might have set for gopls
	const goplsBinaryPath = getBinPath('gopls');
	if (path.isAbsolute(goplsBinaryPath)) {		
		return goplsBinaryPath;		
	}
	
	// Otherwise, prompt the user to install the language server.
	promptForMissingTool('gopls');
```

None of the other tools get to show any error when the alternate tool that is defined is not found.
So, am fine removing similar check for the language server

Uers who have been relying on the alternate tool for go-langserver, will now see the prompt for installing `gopls` which is what we want anyway.
And I believe the number of users falling under this category is going to be a very small

stamblerre:
I agree that it's a small number of users, but I still would prefer to show high-quality error messages for topics relating to the language server. There's been a lot of confusion around it, and people might have old settings configured that they forgot about. I'm happy to delete this code for next release or something like that, but I prefer to err on the side of more error messages.

ramya-rao-a:
In that case I propose the below:

```javascript
// Get the path to gopls
const goplsBinaryPath = getBinPath('gopls');
if (path.isAbsolute(goplsBinaryPath)) {
	return goplsBinaryPath;
}

if (goConfig['alternateTools']) {
	// Show error message for alternate tool configured for gopls that cannot be found
	const goplsAlternate = goConfig['alternateTools']['gopls'];
	if (goplsAlternate) {
		vscode.window.showErrorMessage(
			`Cannot find the alternate tool ${goplsAlternate} configured for gopls. 
			Please install it and reload this VS Code window.`
		);
		return;
	}

	// Show error message for alternate tool configured for go-langserver which is no longer respected
	const goLangserverAlternate = goConfig['alternateTools']['go-langserver'];
	if (goLangserverAlternate) {
		vscode.window.showErrorMessage(`Support for "go-langserver" has been deprecated in favor of gopls.
The alternate tool configuration for "go-langserver" can be deleted.`);
	}
}

// Prompt the user to install gopls.
promptForMissingTool('gopls');
```

stamblerre:
Done. Sorry about the confusion with the PR getting closed - managing 3 remotes has proven to be a bit confusing, and I'm not the best at git :)

ramya-rao-a:
Thanks for your patience @stamblerre :)
I have pushed a commit for a tiny change to replace "this setting" with "alternate tool setting" as the user won't be aware of what "this" refers to

Issue #3044, opened 2020-02-12
performance: limit concurrent subprocesses


stamblerre:
There are a number of issues filed about the number of processes started by the extension, and in particular, processes that get started by various actions on save. These include `golint`, `go vet`, `go list`, `go doc`, etc. We should make sure that only one process of each "kind" is running at a given instant, meaning that we need to logic to cancel these processes when another is spawned.

This issue tracks this work, and as such encompasses https://github.com/microsoft/vscode-go/issues/2276, https://github.com/microsoft/vscode-go/issues/1916, https://github.com/microsoft/vscode-go/issues/1015, and https://github.com/microsoft/vscode-go/issues/2421 (a few of the issues I found with this type of report). Let's centralize discussion here.

zmb3:
Did some investigation here.  There's two items worth discussing.

1. What do we mean when we say only one of each _kind_ of process should run at any given time.  A simple heuristic is to just look at the command, but does that fail us for things like the `go` tool  with many subcommands?  Should we be able to run a `go build` concurrently with a `go list`?

2. I put together a simple implementation of the above, and the tests fail because each suite kicks off a lot of concurrent runs of the same tool, which then get killed.  Is it worth rewriting the tests to serialize everything, or can we detect when we're running under the test suite and disable this behavior?

ramya-rao-a:
We have code in place that ensures that existing child processes are killed before a new process is started for the same feature. We make use of the cancellation token for this.

You can see this in action for most of the features. For example, in case of the build on save feature, we create a new cancellation token for each invocation of the feature and cancel the token for the previous invocation:

https://github.com/microsoft/vscode-go/blob/master/src/goBuild.ts#L78-L85

When the token is cancelled, the corresponding child process is killed

https://github.com/microsoft/vscode-go/blob/6d1bf5c7999d5af0adcc15141ca012c34a9ca71c/src/util.ts#L674-L680

So, what we need is an investigation to see which features don't have the above guards. Or if the guards are triggered in the right places.

For example, in #2956, we can see that multiple `go doc` processes are being triggered. The guards are indeed present for this, so we need to investigate why they guards didn't work. See https://github.com/microsoft/vscode-go/blob/6d1bf5c7999d5af0adcc15141ca012c34a9ca71c/src/util.ts#L1029-L1032

zmb3:
There's a few problems with the following: 

[...]

PR #3131, opened 2020-03-27
Fix process cancelation

zmb3:
Processes were not being killed due to the path to the external scripts being incorrect.  Fix this by relying on the tree-kill package rather than shelling out to an external script to kill a process and its children.

Also add sinon as a dev dependency for testing purposes.

Updates #3044


quoctruong:
That sounds good. I’ll close this pr then.

src/debugAdapter/goDebug.ts
```diff
@@ -36,6 +36,7 @@ import {
 	getInferredGopath,
 	parseEnvFile
 } from '../goPath';
+import { killTree } from '../util';
```
ramya-rao-a: We cannot have the debug adapter take a dependency on util which depends on the `vscode` module. This is because the debug adapter is run in a process different from the rest of the extension. The process in which the rest of the extension runs has access to `vscode`, but the process running the debug adapter doesnt. That's why previously, the kill code is copied here in this file instead of re-using from util

ramya-rao-a: We can however move the killTree out of the util into a separate file and then re-use it from there

zmb3: Ahh, got it.

PR #3132, opened 2020-03-27
Switch to using treekill

quoctruong:
Switch to using [tree-kill package](https://www.npmjs.com/package/tree-kill) instead of using our own implementation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1365,6 +1365,11 @@
         }
       }
     },
+    "tree-kill": {
+      "version": "1.2.2",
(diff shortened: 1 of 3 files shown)


stamblerre:
Thanks for putting this PR together, @quoctruong! Looks like @zmb3 just beat you to it though: https://github.com/microsoft/vscode-go/pull/3131. I'd suggest we go with Zac's PR since it includes tests, if that's ok with you.

ramya-rao-a:
Closing in favor of #3131

Issue #3130, opened 2020-03-26
Test for build tags failing on Go tip


ramya-rao-a:
The test with title `Build Tags checking` is failing consistently, but only on Go tip

https://travis-ci.org/github/microsoft/vscode-go/jobs/666658067?utm_medium=notification&utm_source=github_status

cc @stamblerre, @hyangah

hyangah:
Probably https://go-review.googlesource.com/c/go/+/185345
I see the test look for the specific error message. Does the Go extension's behavior depend on the  specific error message?

ramya-rao-a:
> Does the Go extension's behavior depend on the specific error message?

No, it doesnt. Am fine with removing checks on the exact contents of the error message

PR #3135, opened 2020-03-27
test/integration: fix Build Tags checking test

hyangah:
The error message when there are no Go files to build is changed
in go1.15. That broke this test that includes the diagnostics
message string checking. Adds the message from go1.15 in the
match set.

Fixes microsoft/vscode-go#3130


ramya-rao-a:
Thanks @hyangah!

Issue #3096, opened 2020-03-12
Debug: add "go run ." support


chapati:
After changes made in the issue #2085 'Start without debugging' always defaults to the current go file.  The rationale behind the change was the following: 

> Debug: Start without Debugging is the command to use to simply run the current file rather than debugging.

Let's consider the following program.

main. go

package main
func main () {
   second()
}

second.go
package main
func second() {
}

If you open main.go in your active editor and Ctrl+F5 you would get 'undefined: second'
If you open second.go in your active editor and Ctrl+F5 you would get 'runtime.main_main·f: function main is undeclared in the main package'

While I agree that cmd should not start debugging I do not see any reasons why I'm unable to 'just run' the program that has several go files. Currently I'm forced to debug, i.e. switch to the debug view, stop at each breakpoint, switch back to the editor when debug is finished, while in 90% of cases I just want to run, check output and continue editing.

The current behaviour is pretty useless. It works only for programs that have only one file with the main in it. This is pretty rare. It would be much more useful to run the whole package ("go run packagename" or "go run .") and not the current file or at least allow to override the current behaviour from launch.json. Currently this is forced in

https://github.com/microsoft/vscode-go/blob/master/src/goDebugConfiguration.ts

debugConfiguration['currentFile'] =	activeEditor && activeEditor.document.languageId === 'go' && activeEditor.document.fileName;

and in  https://github.com/microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts

and there is no way to run the whole package. Let's add an option to allow running the whole package / execute "go run ." or "go run packagename".  Take a look at https://github.com/golang/go/issues/22726 for more details about "go run ."

hyangah:
I think this is a valid feature request. I arrived late so I am missing the original motivation. Currently `go run` takes list of `.go` files, package import path, `.`, or relative path. So accepting them seem more reasonable. What do you think? @ramya-rao-a

chapati:
> I think this is a valid feature request. I arrived late so I am missing the original motivation. Currently `go run` takes list of `.go` files, package import path, `.`, or relative path. So accepting them seem more reasonable. What do you think? @ramya-rao-a

Originally Ctrl+F5 started debugging as well. In #2085 they mad it to run the file currently opened in editor. I suggest to extend change made in #2085 and support not only simple the-whole-program-is-one-file case but make Ctrl+F5 more general.

ramya-rao-a:
[...]

PR #3147, opened 2020-04-02
replace . by current file when running without debugging

simpleapples:
i have read the issue https://github.com/microsoft/vscode-go/issues/3096 about adding 'go run .' support and i think the commit https://github.com/microsoft/vscode-go/commit/78518d7e9736670d75fe2cd648c7c3eb23413157 has some stuff to work on @ramya-rao-a :

1. the owner of https://github.com/microsoft/vscode-go/issues/3096 just want to solve the problem which can not run a go file without debugging, i think there is no need to pass '.' to `go run` to solve this problem, pass current dir to `cwd` is fine.

2. if pass the '.' to `go run`, users will not check the process, because the executable file will by named as the directory name by go compiler, it's not consistent.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -427,15 +427,15 @@ class Delve {
 					if (mode === 'debug') {
 						this.noDebug = true;
 						const runArgs = ['run'];
-						const runOptions: { [key: string]: any } = { env };
+						const runOptions: { [key: string]: any } = { cwd: dirname, env };
 						if (launchArgs.buildFlags) {
 							runArgs.push(launchArgs.buildFlags);
 						}
-						runArgs.push(program);
 						if (isProgramDirectory) {
-							runOptions.cwd = program;
-						} else {
-							runOptions.cwd = dirname;
+							runArgs.push("*.go");
+						}
+						else {
+							runArgs.push(program);
 						}
 						if (launchArgs.args) {
 							runArgs.push(...launchArgs.args);
@@ -472,7 +472,7 @@ class Delve {
 				if (!existsSync(launchArgs.dlvToolPath)) {
 					log(
 						`Couldn't find dlv at the Go tools path, ${process.env['GOPATH']}${
-							env['GOPATH'] ? ', ' + env['GOPATH'] : ''
+						env['GOPATH'] ? ', ' + env['GOPATH'] : ''
 						} or ${envPath}`
 					);
 					return reject(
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -431,12 +431,11 @@ class Delve {
 						if (launchArgs.buildFlags) {
 							runArgs.push(launchArgs.buildFlags);
 						}
+						runArgs.push(program);
 						if (isProgramDirectory) {
 							runOptions.cwd = program;
```
hyangah, line 436: in both cases (isProgramDirectory is true or false) `runOptions.cwd` can be `dirname`.

src/debugAdapter/goDebug.ts
```diff
@@ -431,12 +431,11 @@ class Delve {
 						if (launchArgs.buildFlags) {
 							runArgs.push(launchArgs.buildFlags);
 						}
+						runArgs.push(program);
```
hyangah, line 434: from src/goDebugConfiguration.ts, the program can be set to the active editor's document file name. If the program is set to the file name, I'm afraid it's basically reverting the intended fix for #3096 (supporting the main package with multiple files). What am I missing?

src/debugAdapter/goDebug.ts
simpleapples: you're right, i have moved the `cwd` to initialization of `runOptions` https://github.com/microsoft/vscode-go/pull/3147/commits/763acb860eb30a36e786250a6805e474e8756949

src/debugAdapter/goDebug.ts
simpleapples: my mistake, it turely reverts the intended fix for #3096 , but i found a win-win solution, if the program is set to `*.go` instead of `.`, the executable filename will be named as the filename of main function, and it can support multiple go files in a folder.

hyangah: Thank you for addressing the comment. In that case, the executable filename will be the first file name. 

Naming the binary based on the directory name is the de-facto standard of the go command and users who'd get the command using `go get` command will depend on the behavior. For example, golang.org/x/tools/gopls has main.go as the only file but we expect the final binary name to be the directory name 'gopls', not 'main'.  Why is it a problem?

simpleapples: i agree with your opinion, actually, i thought the executable name should be same in both debug and run mode. but it's not a standard.

src/debugAdapter/goDebug.ts
```diff
@@ -427,15 +427,14 @@ class Delve {
 					if (mode === 'debug') {
 						this.noDebug = true;
 						const runArgs = ['run'];
-						const runOptions: { [key: string]: any } = { env };
+						const runOptions: { [key: string]: any } = { cwd: dirname, env };
```
ramya-rao-a, line 430: @simpleapples, @hyangah I think I am missing something... I am not seeing a net change here.
Given that `dirname` is the same as `program` when `isProgramDirectory` is true, setting the `cwd` upfront here and the code before would essentially result in the same behavior

Of course, for simplicity sake, the change in this PR is good. I just want to confirm that this change doesnt change any behavior to the end user

simpleapples: i reviewd the change again and fix some typo. i think it won't change any behavior to the end user.

ramya-rao-a:
For anyone landing here, the net change in this PR is only code refactoring and no change to functionality

PR #3156, opened 2020-04-05
src/goLanguageServer.ts: remove fallbacks for features that cannot be opted-out of

stamblerre:
We should only have fallbacks for the features that we support opting out of.
Also, remove a few unused imports.

/cc @hyangah

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -17,20 +17,10 @@ import {
 	HandleDiagnosticsSignature,
 	LanguageClient,
 	ProvideCompletionItemsSignature,
-	ProvideDefinitionSignature,
 	ProvideDocumentFormattingEditsSignature,
-	ProvideDocumentHighlightsSignature,
 	ProvideDocumentLinksSignature,
-	ProvideDocumentSymbolsSignature,
-	ProvideHoverSignature,
-	ProvideReferencesSignature,
-	ProvideRenameEditsSignature,
-	ProvideSignatureHelpSignature,
-	ProvideWorkspaceSymbolsSignature,
 	RevealOutputChannelOn
 } from 'vscode-languageclient';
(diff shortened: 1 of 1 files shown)


src/goLanguageServer.ts
```diff
@@ -281,12 +202,6 @@ export async function registerLanguageFeatures(ctx: vscode.ExtensionContext) {
 			ctx.subscriptions.push(languageServerDisposable);
 		})
 	);
-
-	// gopls is the only language server that provides live diagnostics on type,
-	// so use gotype if it's not enabled.
-	if (!(toolName === 'gopls' && config.features['diagnostics'])) {
-		vscode.workspace.onDidChangeTextDocument(parseLiveFile, null, ctx.subscriptions);
```
ramya-rao-a, line 288: Wont we need this if user has opted out of diagnostics?

stamblerre: I would prefer to avoid these types of fallbacks, as it can make debugging `gopls` issues difficult. I don't see a reason that someone would opt-out of `gopls` diagnostics on-type, but still want diagnostics on type from a different tool.

ramya-rao-a: > I don't see a reason that someone would opt-out of gopls diagnostics on-type, but still want diagnostics on type from a different tool.

Previously, it has been mainly due to the issues with the diagnostics feature from gopls. If we are saying that all of those issues are resolved, then my next question would be as to what are the cases where one doesnt want diagnostics on type

stamblerre: I think there is some set of people that prefer to have their diagnostics only show up on save, not on-type. I think at this point, if someone is having issues with `gopls` diagnostics, we should work with them to resolve the issue, or they can disable `gopls`, but we can't let people replace the diagnostics - it's too confusing for debugging purposes.

PR #3157, opened 2020-04-06
test/gopls: initial test setup for gopls and hover tests

hyangah:
This CL includes only HoverProvider tests, that use the same scenario
as the hover provider tests with godoc and gogetdoc in
integration/extension.test.ts.

Gopls test setup is complicated because gopls currently does not
work in a single-file mode nor support mono repos with multiple
modules. VS Code in test environment does not support dynamic
registration of workspace folders during testing.  So, the
workspace folder needs to be specified before the extension
host launches (by vscode-test.runTests or launch.json extensionHost
type tasks).

As a workaround, I use an empty directory and start the extension
host with the empty directory for a testing. Then, let the test
populate the directory with necessary files. This is not ideal;
file copy is slow and the use of the single workspace directory
prevents parallelized testing.

Commiting an empty directory to use as a scratch space is not
ideal either, but no worse than what vscode dev in my opinion.
https://github.com/microsoft/vscode/blob/master/.vscode/launch.json

I also considered shelling out the extension host launch
after creating a temp directory and using it as a workspace directory.
That may work inside test/runTest.ts, but complicates debugging
when sarting the test using launch.json.

Also passes '--user-data-dir' flag to prevent the tests from
using my personal vscode settings and interfere tests.

Change-Id: I7b064441720da5c89833afab35e8273de808cfad

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ install:
   - go get -u -v github.com/zmb3/gogetdoc
   - go get -u -v golang.org/x/lint/golint
   - go get -u -v golang.org/x/tools/cmd/gorename
+  - GO111MODULE=on go get golang.org/x/tools/gopls
 
 script:
   - npm run lint
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -44,7 +44,7 @@
 				"--extensionDevelopmentPath=${workspaceFolder}",
 				"--extensionTestsPath=${workspaceFolder}/out/test/integration/index",
 				"--timeout",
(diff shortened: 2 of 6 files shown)


stamblerre:
Thanks for working on this, @hyangah! It looks good to me, and I'm sorry you had to jump through so many hoops to get `gopls` to run in the test framework. I'll aim to support single-file mode for the next release so that we can avoid some of this, but I do still think it's good for us to have single-file, `GOPATH`, and Go modules-style tests.

PR #3167, opened 2020-04-09
Add note about needing WSL 2 on Windows 10

mpatnode:
delve will hang on the WSL 1 kernel due to trace system call issues.    Just adding a note to that effect.  Would be nice to put in the Wiki under troubleshooting as well.  @ramya-rao-a

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ A few of these are available in the editor context menu as an experimental featu
 
 ### _Optional_: Debugging
 
-To use the debugger, you must currently manually install `delve`.  See the [Installation Instructions](https://github.com/derekparker/delve/tree/master/Documentation/installation) for full details.  On MacOS it requires creating a self-signed cert to sign the `dlv` binary.
+To use the debugger, you must currently manually install `delve`.  See the [Installation Instructions](https://github.com/derekparker/delve/tree/master/Documentation/installation) for full details.  On MacOS it requires creating a self-signed cert to sign the `dlv` binary.  If using WSL on Windows, you will need the WSL 2 Linux kernel.  See [WSL 2 Installation](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install) and note the Window 10 build version requirements. 
 
 For more read [Debugging Go Code Using VS Code](https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code).
 
--- a/docs/Debugging-Go-code-using-VS-Code.md
+++ b/docs/Debugging-Go-code-using-VS-Code.md
@@ -234,7 +234,7 @@ You may see this in the debug console, while trying to run in the `test` mode. T
 
 **_Solution_**: Ensure that the `program` attribute points to the folder that contains the test files you want to run.
 
-#### delve/launch hangs with no messages
+#### delve/launch hangs with no messages when using WSL
 Try running ```delve debug ./main``` at the WSL command line and see if you get a prompt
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ A few of these are available in the editor context menu as an experimental featu
 
 ### _Optional_: Debugging
 
-To use the debugger, you must currently manually install `delve`.  See the [Installation Instructions](https://github.com/derekparker/delve/tree/master/Documentation/installation) for full details.  On MacOS it requires creating a self-signed cert to sign the `dlv` binary.
+To use the debugger, you must currently manually install `delve`.  See the [Installation Instructions](https://github.com/derekparker/delve/tree/master/Documentation/installation) for full details.  On MacOS it requires creating a self-signed cert to sign the `dlv` binary.  If using WSL on Windows, you will need the WSL 2 Linux kernel.  See [WSL 2 Installation](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install) and note the Window 10 build version requirements.
```
ramya-rao-a, line 210: > and note the Window 10 build version requirements.

Its not clear to me as to why we want to call this out. Can you elaborate?

mpatnode: The default WSL uses syscall mapping, hence delve just hangs on the trace() call.   You need to join the Windows Insiders program and get the latest version of the OS, and then you can install the WSL 2 kernel, which is a true virtualized machine.

mpatnode:
> We just copied over all the content from the wiki to the `docs` folder in this repo and will be updating the wiki to just point to these.
> 
> Can you make a similar update in the wiki like you mentioned?

👍 will do

docs/Debugging-Go-code-using-VS-Code.md
```diff
@@ -234,6 +234,11 @@ You may see this in the debug console, while trying to run in the `test` mode. T
 
 **_Solution_**: Ensure that the `program` attribute points to the folder that contains the test files you want to run.
 
+#### delve/launch hangs with no messages
+Try running ```delve debug ./main``` at the WSL command line and see if you get a prompt
+
+**_Solution_**: Ensure you are running the WSL 2 Kernel, which (as of 4/15/2020) requires an early release of the Windows 10 OS.  This is available to anyone via the Windows Insider program.  See [WSL 2 Installation](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install)
+
 #### could not launch process: could not fork/exec
```
mpatnode, line 242: @ramya-rao-a Wasn't sure what to do with this dangler.

ramya-rao-a: Sorry, which dangler?

mpatnode: The "could not launch process: could not fork/exec" line has no explanation or solution

ramya-rao-a: Ah! he solution to that is split for Mac vs Linux, looks like the lack of indentation isnt helping anyone :) I'll fix that

docs/Debugging-Go-code-using-VS-Code.md
```diff
@@ -234,6 +234,11 @@ You may see this in the debug console, while trying to run in the `test` mode. T
 
 **_Solution_**: Ensure that the `program` attribute points to the folder that contains the test files you want to run.
 
+#### delve/launch hangs with no messages
```
ramya-rao-a, line 237:
```suggestion
#### delve/launch hangs with no messages when using WSL
```

Issue #3148, opened 2020-04-02
Wiki: update FAQ, modules support, misc...


hyangah:
@ramya-rao-a Unfortunately, there is no easy way to  make a PR to the wiki repo like usual git repos. Github doesn't have UI support for easy PR or PR review for wiki repos.

Included is the patch created with `git diff`. Or you can directly pull the commit from https://github.com/hyangah/vscode-go.wiki.git

[wiki.diff.zip](https://github.com/microsoft/vscode-go/files/4423834/wiki.diff.zip)

cc @stamblerre

hyangah:
the commits were merged.

PR #3168, opened 2020-04-09
Documentation: move docs in wiki to Documentation directory

hyangah:
By having all docs under Documentation, we can use the GitHub UI for 
review and easy pr creation.

This commit does not change the file contents. All files here
are verbatim copies of the contents from
https://github.com/microsoft/vscode-go.wiki.git@12d2e8f5

Update microsoft/vscode-go#3148


hyangah:
Links in the documents still point to wiki pages. Let me know if you want me to fix them in this PR.

@ramya-rao-a

.vscodeignore
```diff
@@ -8,3 +8,4 @@ test/
 **/*.map
 **/tslint.json
 build/**/*
+Documentation/
```
ramya-rao-a, line 11: Can we use either `documentation` or `docs` as the folder name instead? JS projects usually use camel casing in names and short names in general

hyangah: Moved to docs. Sorry for the forced push (I messed up)

PR #3182, opened 2020-04-16
package.json: update all dependencies

stamblerre:
@ramya-rao-a: I know you typically don't like these big updates to all dependencies, but I figured there are only a few major version updates, and it would be nice to get these improvements now that we've upgraded the VS Code engine. Also, since its early in the release cycle, we can fix any bugs that come up in the nightly build. Happy to downgrade any of these if you think it's necessary.

/cc @hyangah


package.json
```diff
@@ -40,33 +40,33 @@
   },
   "extensionDependencies": [],
   "dependencies": {
-    "diff": "^4.0.1",
+    "diff": "^4.0.2",
```
ramya-rao-a, line 43: Note: The use of ^ here results in the package-lock.json using the latest version of `diff` with major version 4. So, this change (or any others where we havent updated the major version) doesnt make a net difference.

package.json
```diff
@@ -40,33 +40,33 @@
...
-    "semver": "^6.3.0",
+    "semver": "^7.3.2",
     "tree-kill": "^1.2.2",
-    "vscode-debugadapter": "^1.36.0",
-    "vscode-debugprotocol": "^1.36.0",
+    "vscode-debugadapter": "^1.40.0",
+    "vscode-debugprotocol": "^1.40.0",
     "vscode-extension-telemetry": "^0.1.2",
-    "vscode-languageclient": "6.1.0",
+    "vscode-languageclient": "^6.1.3",
```
ramya-rao-a, line 51: The reason I pinned the version of the language client was that the latest version 6.1.3 because the vscode engine for later versions was specified as 1.43 in vscode-languageclient

stamblerre: Ah, I see - makes sense, sorry about that!

package.json
stamblerre: Ah, I thought so but when I tried playing around with it, I wasn't sure if it was working like that. Either way, I used the `npm-check-updates` command to create this change, so I think it's probably easiest to leave it as is? If not I can remove all of the minor/patch version upgrades.

ramya-rao-a: sure, we can keep the change

package.json
```diff
@@ -40,33 +40,33 @@
...
-    "prettier": "^1.16.4",
-    "sinon": "^9.0.1",
-    "tslint": "^5.19.0",
-    "typescript": "^3.7.2",
-    "vscode-test": "^1.2.3"
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^13.11.1",
+    "@types/semver": "^7.1.0",
+    "@types/sinon": "^9.0.0",
+    "@types/vscode": "^1.44.0",
```
ramya-rao-a, line 61: Turns out min version specified for `@types/vscode` cannot be greater than what we have in engines, the vsce command begins to fail. Also, we missed getting the lock file updated in this PR. Fixed both in https://github.com/microsoft/vscode-go/commit/1c314d4b6094d0be7cd07ce718657d50f4636423

PR #3186, opened 2020-04-17
src/goLanguageServer.ts: respect configuration changes on server restart

stamblerre:
As a precursor to https://golang.org/cl/222418, I wanted to extract the logic to start the language server into the restart function. This way, server restarts will account for changes in the configs, flags, etc. Also, I think it simplifies the logic of starting the server because I made it handle the case when there is no server running yet.

Some possible issues are (1) I had to extract a number of global variables to make this work, and (2) I had to add a dependency on [deep-equal](https://www.npmjs.com/package/deep-equal) in order to correctly diff the environment.

Original change here: https://go-review.googlesource.com/c/vscode-go/+/227279.

/cc @hyangah


src/goLanguageServer.ts
```diff
@@ -6,91 +6,118 @@
...
+	languageServerDisposable = languageClient.start();
+	ctx.subscriptions.push(languageServerDisposable);
+
+	return true;
+}
+
+function rebuildLanguageClient(config: LanguageServerConfig) {
+	// Reuse the same output channel for each instance of the server.
+	if (!serverOutputChannel) {
+		serverOutputChannel = vscode.window.createOutputChannel(config.serverName);
```
ramya-rao-a, line 113: This looks new, are we creating a new output channel for gopls?
Doesnt gopls get a dedicated channel already?

stamblerre: It gets created automatically so there is one already, but I figured we should reuse it across different `gopls` instances if it gets restarted.

src/goLanguageServer.ts
```diff
@@ -6,92 +6,118 @@
...
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.languageserver.restart', () => {
+		return restartLanguageServer(ctx, parseLanguageServerConfig());
+	}));
 
-	// If the user has not enabled the language server,
-	// register the default language features and return.
-	if (!config.enabled) {
-		registerUsualProviders(ctx);
-		return;
+	const config = parseLanguageServerConfig();
```
ramya-rao-a, line 61: Pull this a few lines above and we can use it to register the restart language server command as well

src/goLanguageServer.ts
```diff
@@ -6,92 +6,118 @@
...
+	if (!config || !config.enabled) {
```
ramya-rao-a, line 62: `parseLanguageServerConfig` does not return null/undefined, so we can skip the !config check

```suggestion
	if (!config.enabled) {
```

src/goLanguageServer.ts
```diff
@@ -6,92 +6,118 @@
...
+
+// startLanguageServer starts the language server (if enabled), returning
+// true on success.
+export async function startLanguageServer(ctx: vscode.ExtensionContext): Promise<boolean> {
 	// Subscribe to notifications for changes to the configuration of the language server.
 	ctx.subscriptions.push(vscode.workspace.onDidChangeConfiguration((e) => watchLanguageServerConfiguration(e)));
 
-	const config = parseLanguageServerConfig();
+	// Support a command to restart the language server.
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.languageserver.restart', () => {
```
ramya-rao-a, line 57: Before this PR, we registered the restart command only if the language server was enabled by the user. Now, we are registering it regardless of whether user has enabled it or not. If we want it this way, then at the very least this command should throw an error if the user has not opted into using the language server

src/goLanguageServer.ts
```diff
@@ -6,92 +6,118 @@
...
+	if (!tool) {
+		return false;
 	}
+	const versionToUpdate = await shouldUpdateLanguageServer(tool, config.path, config.checkForUpdates);
+	if (versionToUpdate) {
+		promptForUpdatingTool(tool.name);
+	}
+	// This function handles the case when the server isn't started yet,
+	// so we can call it to start the language server.
+	return restartLanguageServer(ctx, config);
```
ramya-rao-a, line 75: It is odd that we are using a method named "restart.." to start the language server for the first time as well.
Another oddity is that `latestConfig` is never initialized. Meaning, the first time it gets a value set to itself is when the check `if (!deepEqual(latestConfig, config))` fails. 

I get that functionally there is nothing wrong with this, as you have ensured both the first start and subsequent restarts are doing the right things.

But, for the sake of readability, I would recommend the below refactoring

- Rename `startLanguageServer` to `registerLanguageFeatures`. Initialize `latestConfig` here.
- Rename `rebuildLanguageClient` to `startLanguageServer`. Skip building the language client if config has not changed and  start language server here.
- Update `restartLanguageServer` to call `startLanguageServer`

src/goLanguageServer.ts
ramya-rao-a: Can we not share the one that VS Code creates for it? My concern is that we will now have 2 output channels for gopls. One being created here, the other that we already have

src/goLanguageServer.ts
stamblerre: Done.

src/goLanguageServer.ts
stamblerre: Done.

src/goLanguageServer.ts
stamblerre: Fixed by addressing the comment above.

src/goLanguageServer.ts
stamblerre: I made the renaming changes, but I'm not sure I agree that the starting the language client in `rebuildLanguageClient` will make things more readable. Moving the language client start into `rebuildLanguageClient` would basically collapse it into `startLanguageServer`, and I think that having a separate function for building the very large language client object makes things cleaner.

My intention is to make the code in `startLanguageServer` more maintainable - any start of the language server follows the same code path, so we will never have any special cases to fix or reason about.

Similarly, I don't think it's necessary to explicitly initialize `latestConfig` - I don't see a reason why it is different from the other global variables which aren't explicitly initialized. Moreover, it will mean that we may have to remember to fix multiple places in the code if we want to make a change.

Of course I'm happy to change things as you prefer or add more documentation to make things easier to understand, but I just wanted to clarify my thinking before doing that.

src/goLanguageServer.ts
stamblerre: The VS Code language client creates one automatically if you don't set it, but by setting it in the language client object it won't, so we're fine.

ramya-rao-a: Ah! got it

src/goLanguageServer.ts
```diff
@@ -6,92 +6,119 @@
...
-		// should be presented by now.
-		return;
+	// Support a command to restart the language server, if it's enabled.
+	ctx.subscriptions.push(vscode.commands.registerCommand('go.languageserver.restart', () => {
+		return startLanguageServer(ctx, parseLanguageServerConfig());
+	}));
+
+	const tool = getTool(config.serverName);
+	if (!tool) {
+		return false;
```
ramya-rao-a, line 68: `getTool()` works for the predefined set of tools we work with.
`config.servername` can potentially have a value that is not in the predefined set of tools if the user has used `go.alternateTools` to point `gopls` to something else
In such case, we will end up returning false here

ramya-rao-a: I would revert the check here to
```js
if (config.serverName === 'gopls') {
    // call shouldUpdateLanguageServer and prompt to update if needed
}
```

src/goLanguageServer.ts
ramya-rao-a: > Moving the language client start into rebuildLanguageClient would basically collapse it into startLanguageServer, and I think that having a separate function for building the very large language client object makes things cleaner.

makes sense. The only change now I would suggest is to rename `rebuildLanguageClient` to `buildLanguageClient` since it gets used for the initial build as well

stamblerre: Done. Thanks!

src/goLanguageServer.ts
stamblerre: Done.

Issue #1903, opened 2018-09-01
Show panic error information, stack trace, and set editor to source line that panicked.


jpap:
When a debug session panics, Delve will show a code listing with a cursor at the panic with a "Command failed: XXX" message.  You can then issue commands to get a backtrace, etc.

In VSC, a cryptic message "Failed to continue - XXX" is shown in the Debug Console and nothing else.  It would be great if instead:

* Move the editor to the relevant source line that panicked.
* Show the call stack in the debugger pane.  If that's too hard, at the very least please show the panic details in the Debug Console, including stack trace.
* Instead of "Failed to continue", use a more descriptive "Debugger stopped after program failure: XXX".  (The first time I encountered this I thought it was a debugger issue, rather than my code.)

db47h:
@jpap This is the default behavior for dlv to stop at the beginning of [fatalpanic](https://github.com/golang/go/blob/b794ca64d29f3e584cbdf49bde7141d3c12dd2ab/src/runtime/panic.go#L681) when a Go program panics. While I agree that this might be confusing, panics can occur anywhere, even in the standard library (IIRC, some string to int conversion functions do), so it's not easy to guess which stack frame the user will really be interested in. In any case that specific issue should be taken to https://github.com/derekparker/delve/.

Also, in order to have vscode automatically switch to debug view on panic, you can set 

```
"debug.openDebug": "openOnDebugBreak",
```

or

```
"debug.openDebug": "openOnSessionStart"
```

in your user settings.

jpap:
@db47h I am not concerned with the debug view opening automatically.  I am more concerned with:
1. A cryptic error message is shown on the Debug Console, that isn't clearly obvious that the issue is a fault in the program being run.  For example, in #1853 the author probably thought it was their debugger setup, before realizing it was a defect in their code.  (It has happened to me before too.)
2. When a runtime error occurs, other than an explicit call to `runtime.panic`, the debugger also shows a cryptic error message ("Failed to continue - XXX") on the Debug Console, **and the Debug View is blank**.

This is what I get in Delve with the above code sample:

```
$ dlv debug oops.go
Type 'help' for list of commands.
(dlv) c
     1:	package main
     2:	
     3:	func oops() {
     4:		var a *int
=>   5:		*a++
     6:	}
     7:	
     8:	func main() {
     9:		oops()
    10:	}
Command failed: bad access
(dlv) 
```

[...]

PR #3196, opened 2020-04-20
Debug: Improve 'bad access' error message

polinasok:
Clarify the error message that is printed to the user when delve returns "bad access".
Updates https://github.com/microsoft/vscode-go/issues/1903.

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -1818,21 +1818,19 @@ class GoDebugSession extends LoggingDebugSession {
 		}
 
 		let errorMessage = err.toString();
-		// Handle unpropagated fatalpanic errors with a more user friendly message:
+		// Use a more user friendly message for an unpropagated SIGSEGV (EXC_BAD_ACCESS)
+		// signal that delve is unable to send back to the target process to be
+		// handled as a panic.
 		// https://github.com/microsoft/vscode-go/issues/1903#issuecomment-460126884
 		// https://github.com/go-delve/delve/issues/852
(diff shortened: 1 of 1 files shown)


src/debugAdapter/goDebug.ts
```diff
@@ -1818,21 +1818,19 @@ class GoDebugSession extends LoggingDebugSession {
...
 		if (errorMessage === 'bad access') {
+			// Reuse the panic message from the Go runtime.
 			errorMessage =
-				'unpropagated fatalpanic: signal SIGSEGV (EXC_BAD_ACCESS). This fatalpanic is not traceable on macOS, see https://github.com/go-delve/delve/issues/852';
+				`runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation]\nUnable to propogate EXC_BAD_ACCESS signal to target process and panic (see https://github.com/go-delve/delve/issues/852)`;
 		}
 
 		logError(message + ' - ' + errorMessage);
-
-		if (errorMessage === 'bad access') {
```
polinasok, line 1832: This never gets printed because by the time we get to this line, 'bad access' has already been wiped out by the previous statements. I don't think this message is necessary - the stack trace is printed with "last known" comment, which implies that it might not be the expected most recent state.

polinasok:
cc @lggomez
cc @jhendrixMSFT

lggomez:
LGTM

ramya-rao-a:
The latest update 0.14.2 has this change, Thanks @polinasok

Issue #3194, opened 2020-04-20
v0.14.1 does not seem to install gopls anymore


MorpheusXAUT:
### What version of Go, VS Code & VS Code Go extension are you using?
- Run `go version` to get version of Go
	- go version go1.14.2 linux/amd64
- Run `code -v` or `code-insiders -v` to get version of VS Code or VS Code Insiders
	- 1.44.2 ff915844119ce9485abfe8aa9076ec76b5300ddd x64 (Windows 10)
- Check your installed extensions to get the version of the VS Code Go extension 
	- 0.14.1
- Run `go env GOOS GOARCH` to get the operating system and processor architecture details
	- linux amd64

### Share the Go related settings you have added/edited

Run `Preferences: Open Settings (JSON)` command to open your settings.json file.
Share all the settings with the `go.` or `["go"]` prefixes.

```
{
    "go.useLanguageServer": true,
    "[go]": {
        "editor.formatOnSave": true,
        "editor.codeActionsOnSave": {
            "source.organizeImports": true,
        },
        "editor.snippetSuggestions": "none",
    },
    "[go.mod]": {
        "editor.formatOnSave": true,
        "editor.codeActionsOnSave": {
            "source.organizeImports": true,
        },
    },
    "gopls": {
        "usePlaceholders": true,
        "staticcheck": false,
    },
    "go.autocompleteUnimportedPackages": true,
    "go.lintTool": "golangci-lint",
    "go.lintFlags": [
        "--fast"
    ],
    "go.testFlags": [
        "-cover",
        "-race",
        "-count=1",
        "-v"
    ]
}
```

### Describe the bug
I am using a Devcontainer setup for Go, running an golang:1.14.2 Docker image on a Windows 10 host. After updating to the latest version of vscode-go (0.14.1), the extension doesn't seem to install `gopls` in the devcontainer anymore and no prompt to install/update it is shown.

Using the `"go.useLanguageServer": true` setting as described [here](https://github.com/microsoft/vscode-go/blob/master/docs/Go-modules-support-in-Visual-Studio-Code.md), I expected the extension to install `gopls` as required.

Manually downgrading vscode-go to version 0.13.1 and reloading the window prompts for a `gopls` install as it was detected missing.

### Steps to reproduce the behavior:
1. Install vscode-go 0.14.1
2. Setup devcontainer environment with golang:1.14.2 image
3. Re-open project in container, no prompt to install gopls appears

hyangah:
[...]

PR #3197, opened 2020-04-20
src/goLanguageServer: fix missing gopls detection logic

hyangah:
Also, update getMissingTools to check the tool's existence
by simply checking whether the resolved tool path is absolute.
getBinPath already ran the existence check before returning
the absolute path for the tool.

Fixes microsoft/vscode-go#3194

### DIFF
--- a/src/goLanguageServer.ts
+++ b/src/goLanguageServer.ts
@@ -298,9 +298,11 @@ Please install it and reload this VS Code window.`
 		if (alternateTools['go-langserver']) {
 			vscode.window.showErrorMessage(`Support for "go-langserver" has been deprecated.
 The recommended language server is gopls. Delete the alternate tool setting for "go-langserver" to use gopls, or change "go-langserver" to "gopls" in your settings.json and reload the VS Code window.`);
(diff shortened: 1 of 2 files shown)


src/goLanguageServer.ts
```diff
@@ -298,9 +298,11 @@ Please install it and reload this VS Code window.`
 		if (alternateTools['go-langserver']) {
 			vscode.window.showErrorMessage(`Support for "go-langserver" has been deprecated.
 The recommended language server is gopls. Delete the alternate tool setting for "go-langserver" to use gopls, or change "go-langserver" to "gopls" in your settings.json and reload the VS Code window.`);
+			return;
 		}
-		return;
 	}
+
+	console.log('promptForMissingTools(gopls)');
```
stamblerre, line 305: intentional?

stamblerre:
Thanks for doing this!

src/goLanguageServer.ts
hyangah: ooops. deleted

Issue #2850, opened 2019-10-21
Allow sub-test selection


smyrman:
If you have a question, please ask it on https://gitter.im/Microsoft/vscode-go or on the VSCode channel in Gophers Slack(https://invite.slack.golangbridge.org/)

**Is your feature request related to a problem? Please describe.**

VS-Code has a _code lens_ link that allows you to run individual tests. 

![image](https://user-images.githubusercontent.com/75205/67205997-20988a80-f411-11e9-9786-afa3ac1883e7.png)

This works by generating a paramater ala `-run ^(TestDeepEquals)$` for the `go test` command. This is great, but for large tests and table tests, it would be great if there was also a possibility to focus on a single _sub-test_ (as [introduced](https://blog.golang.org/subtests) in Go 1.7). This is something that Go already supports by providing a flag ala  `-run ^(TestDeepEquals/sub-test)$`.

A method of doing this has been suggested before in #1748, but closed due to the complexity of the solution. I want to introduce a hopefully much simpler solution to the problem.

**Describe the solution you'd like**

 I see two possible solutions, which would ideally be _combined_.

Solution 1:

In the test failure output, allow clicking on a failing sub-test to rerun it with the right "name" added to the `-run` parameter.

Solution 2 alternative A:

In the code lense, add a link to run a sub-test. I am a bit on thin ice here, but since go test results are _cached_, if we could easily access this cache, perhaps it's possible to issue a dialog with auto-complete to search for (and select) any previously run sub-test. If no sub-tests have previously been run, the dialog could either be hidden, or allow free-text typing.

![image](https://user-images.githubusercontent.com/75205/67206645-5e49e300-f412-11e9-8700-26a5ade5db7a.png)

Solution 2 alternative B:

As a variant of solution 2, the drop-down could provide only _failing sub-tests_ from the previous run of the full test (that VS Code knows about).

**Describe alternatives you've considered**

Solution #1748 have previously been suggested with a different solution.

**UPDATE**: Replaces wrong link to previous issue (the issue is 1748, not 597)

ramya-rao-a:
Thanks for reporting @smyrman 

#597 was related to generating of tests, where as I believe what you are asking for is running of the sub tests, did I get that right?

ramya-rao-a:
Can you provide a small sample file with subtests that I can use to check the current behavior?

smyrman:
Sorry, when navigating back and forth I managed to isnert the wrong sub-test related issue link. 
[...]

Issue #1748, opened 2018-06-21
Support ability to run individual subtests


Robinnnnn:
One of my favorite VSCode features is the ability to run a specific Go test with the click of a button via **run test**, which is auto-generated based on [certain factors](https://twitter.com/ramyanexus/status/1009659005038292992). It would be great if this button could also show up under [subtests and sub-benchmarks](https://blog.golang.org/subtests), a feature introduced in Go 1.7 for increased flexibility + modularization.

Below is a screenshot of some tests I'm working on for card registration. Notice the **run test** option appearing over `func TestCard`, but *not* over the subtests within (declared via `t.Run`):

![screen shot 2018-06-21 at 11 22 11 am](https://user-images.githubusercontent.com/12162433/41728902-f03f5e54-7545-11e8-9480-94374f4c38d6.png)

Running subtests in the above example using the command line is easy if you're trying to group them:
```
go test -run Card           // runs `TestCard` and all underlying subtests
go test -run Card/valid     // runs all subtests under `TestCard` that match `valid` (first and fourth ones)
```

Where the CLI gets cumbersome is when I want to run a _specific_ subtest and have to type out the entire name:
```
go test -run Card/Card__invalid_creation__invalid_token
```

This is a first-world problem but it would be amazing to just do this through the click of a button.

I guess the challenge is that subtests are declared _within_ a function declaration and never at the outer level. It is also common to use `t.Run` inside a loop, which would make the purpose of a button more ambiguous:

```
func TestGroupedParallel(t *testing.T) {
    for _, tc := range tests {
        tc := tc // capture range variable

        /////// what would a button do if it appeared here?
        t.Run(tc.Name, func(t *testing.T) {
            t.Parallel()
            ...
        })
    }
}
```

It definitely wouldn't be the end of the world if this is not worth implementing but I just wanted to put it on your radar! Thank you in advance.

ramya-rao-a:
> I guess the challenge is that subtests are declared within a function declaration and never at the outer level. 

Yes, thats indeed the challenge.
The Go plugin gets the outline of the file (what you see when you do a Ctrl/Cmd+Shift+O), get all functions whose name starts with either "Test", "Benchmark" or "Example" and then provides the code lens to run the tests.

In case of sub tests, we need to parse the inner contents of the function which the Go plugin isn't capable of doing as it doesn't really understand Go code. For all other rich language features, we use the Go tools like godef, gocode etc.

[...]

Issue #2144, opened 2018-11-25
File tests include too many tests


JPMoresmau:
Steps to Reproduce:

1. Create two go test files, one with two test functions like TestA and TestB, the other with one test function TestAA
2. Choose "run file tests" on first file
3. TestAA runs too
4. This is because the regexp generated is `^TestA|TestB$`, it should be `^(TestA|TestB)$`. 

Code is in src/testUtils.ts:
`util.format('^%s$', testFunctions.join('|'))`

Should be `util.format('^(%s)$', testFunctions.join('|'))`

I can try to submit a PR if the change is confirmed, thanks.

ramya-rao-a:
Nice catch!
Yes that change looks right. Please go ahead and submit the PR.

ramya-rao-a:
The fix for this is now out in the latest update to the Go extension (0.8.0)
Thanks @JPMoresmau!

PR #3199, opened 2020-04-21
Add support for running an individual subtest

seancaffery:
Being able to run subtests is something that has been brought up a [couple](https://github.com/microsoft/vscode-go/issues/2850) of [times](https://github.com/microsoft/vscode-go/issues/1748) and something that I would find useful.

I've enabled running subtests through a command only - adding a CodeLens would be nice but as mentioned in the other issues, is tricky because there isn't currently a way to list subtests. While this doesn't go as far as allowing retries of a failed test, it does allow targeting a specific subtest which would get us some access to subtest functionality.

**Approach**
Similar to `Test function at cursor`, this approach allows running a test by placing the cursor inside the test function or on the `t.Run` call. It will search back from the current cursor position within the current outer test function for a line that contains `t.Run`. If a subtest is found, an appropriate call to `runTestAtCursor` is constructed and the test is executed by existing `go test` functionality.

I guess my main question for this PR is whether or not checking for a line with specific text (i.e. `t.Run()`) is a reasonable thing to do to allow this feature.

If you're happy to include this or some variation of it, I'm happy to fix up any style issues and add any tests that are required.

This is a simple example that I used for testing the change
```go
package blah

import (
        "fmt"
        "testing"
)

func TestThingss(t *testing.T) {
        t.Run("heres a test", func(t *testing.T) {
                fmt.Println("here's a test")
        })

        t.Run("some other test", func(t *testing.T) {
                fmt.Println("some other test")
                t.FailNow()
        })
}
```

### DIFF
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -124,6 +124,10 @@ export function subTestAtCursor(goConfig: vscode.WorkspaceConfiguration, cmd: Te
 		vscode.window.showInformationMessage('No tests found. Current file is not a test file.');
 		return;
 	}
+	if (cmd !== 'test') {
+		vscode.window.showInformationMessage('Only the "test" is command supported for subtests');
+		return;
+	}
 
 	editor.document.save().then(async () => {
 		try {
@@ -142,7 +146,7 @@ export function subTestAtCursor(goConfig: vscode.WorkspaceConfiguration, cmd: Te
 			}
 
 			const testFunction = currentTestFunctions[0];
-			const runRegex = /t.Run\("(.*)"/;
+			const runRegex = /t.Run\("([^"]+)",/;
 			let lineText: string;
 			let match: RegExpMatchArray | null;
 			for (let i = editor.selection.start.line; i >= testFunction.range.start.line; i--) {
@@ -154,17 +158,13 @@ export function subTestAtCursor(goConfig: vscode.WorkspaceConfiguration, cmd: Te
 			}
 
 			if (!match) {
-				vscode.window.showInformationMessage('No sub test function found at cursor.');
+				vscode.window.showInformationMessage('No subtest function with a simple subtest name found at cursor');
 				return;
 			}
 
 			const subTestName = testFunctionName + '/' + match[1];
 
-			if (cmd === 'test') {
-				await runTestAtCursor(editor, subTestName, testFunctions, goConfig, cmd, args);
-			} else {
-				throw new Error('Unsupported command.');
-			}
+			await runTestAtCursor(editor, subTestName, testFunctions, goConfig, cmd, args);
 		} catch (err) {
 			console.error(err);
 		}
--- a/src/goTest.ts
+++ b/src/goTest.ts
@@ -20,6 +20,7 @@ import {
 	TestConfig
 } from './testUtils';
 import { getTempFilePath } from './util';
+import { Range } from 'vscode';
 
 // lastTestConfig holds a reference to the last executed TestConfig which allows
 // the last test to be easily re-executed.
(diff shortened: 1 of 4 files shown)


src/testUtils.ts
```diff
@@ -455,7 +455,11 @@ function targetArgs(testconfig: TestConfig): Array<string> {
 			// in running all the test methods, but one of them should call testify's `suite.Run(...)`
 			// which will result in the correct thing to happen
 			if (testFunctions.length > 0) {
-				params = params.concat(['-run', util.format('^(%s)$', testFunctions.join('|'))]);
```
seancaffery, line 458: I had to change this because enclosing the full subtest name causes the match to fail and the test to not execute. The [issue](https://github.com/microsoft/vscode-go/issues/2144) that introduced the brackets mentions that this is an issue for running multiple tests. With an explicit test count, this should maintain the current functionality.

hyangah:
@seancaffery Can you please include a screenshot or video recording that demonstrates this feature? Subtests are often used along with the table-driven tests (e.g. https://cs.opensource.google/go/tools/+/master:txtar/archive_test.go;l=45) - I wonder how this feature will work when the subtest name is something dynamically formatted or set with a variable.

seancaffery:
@hyangah thanks for taking a look at this. Here's a gif using my example file.

subtest

> I wonder how this feature will work when the subtest name is something dynamically formatted or set with a variable

A message will be displayed that says that there is no subtest found. e.g.

tabletest

🤔 that could be a bit misleading / confusing for people. I could add a more specific message that says dynamically named sub tests / table tests are not supported.
Do you have any thoughts on what to do here?

hyangah:
@stamblerre Any thought? For long term, is there any `gopls`-based trick to analyze code more accurately than relying on the regexp?

stamblerre:
@hyangah: `gopls` will probably be able to do this without a regex (using type information), but it's not planned for v1.0, so I'd say this could be a good first step until `gopls` handles this case.

src/goTest.ts
```diff
@@ -107,6 +107,70 @@ async function runTestAtCursor(
...
+				: currentTestFunctions
+					.map((el) => el.name)[0];
+
+			if (!testFunctionName) {
+				vscode.window.showInformationMessage('No test function found at cursor.');
+				return;
+			}
+
+			const testFunction = currentTestFunctions[0];
+			const runRegex = /t.Run\("(.*)"/;
```
hyangah, line 145: Maybe /t.Run\("([^"]+)"/ ?

src/goTest.ts
```diff
@@ -107,6 +107,70 @@ async function runTestAtCursor(
...
+			for (let i = editor.selection.start.line; i >= testFunction.range.start.line; i--) {
+				lineText = editor.document.lineAt(i).text;
+				match = lineText.match(runRegex);
+				if (match) {
+					break;
+				}
+			}
+
+			if (!match) {
+				vscode.window.showInformationMessage('No sub test function found at cursor.');
```
hyangah, line 157: How about "No sub test function with a simple subtest name found at cursor"?

src/goTest.ts
```diff
@@ -107,6 +107,70 @@ async function runTestAtCursor(
...
+				return;
+			}
+
+			const subTestName = testFunctionName + '/' + match[1];
+
+			if (cmd === 'test') {
+				await runTestAtCursor(editor, subTestName, testFunctions, goConfig, cmd, args);
+			} else {
+				throw new Error('Unsupported command.');
```
hyangah, line 166: move cmd === 'test' check to the beginning of this function.

hyangah:
Thanks for addressing the comment @seancaffery. Can you please sign the CLA?

seancaffery:
Thanks for the review @hyangah.

I am working through the CLA process with my employer. Is there a way us to sign a corporate CLA rather than an individual one? Something like [Shopify do](https://cla.shopify.com/corporate-cla)?

hyangah:
@ramya-rao-a can you please help with the corp CLA?

ramya-rao-a:
Hey @seancaffery,  

Looks like this is your first PR contribution to this project, Thanks & Welcome!
My apologies for the delay in responding.

We will have to close this PR as we are not accepting any PRs at the moment unless it is related to the move or critical bug fixes. Please consider creating this PR in the new home i.e https://github.com/golang/vscode-go.

seancaffery:
Thanks @ramya-rao-a. I've submitted this to the new repo https://github.com/golang/vscode-go/pull/87

Issue #2805, opened 2019-10-02
gogetdoc test for signature help failing consistently for Go 1.12


ramya-rao-a:
The test case `Test SignatureHelp Provider using gogetdoc` is consistently failing with the below for Go version 1.12. The test runs fine for previous 2 versions of Go.

```
AssertionError [ERR_ASSERTION]: No signature for gogetdocTestData/test.go:[object Object]
```

This could be a problem with `gogetdoc` itself, therefore needs investigation and a potential issue logged in the gogetdoc repo.

An improvement for the test itself could be a better error message than [object Object] :)

PRs are most welcome
Code pointers:
- [Set up local dev env](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code)
- Find the test `Test Definition Provider using gogetdoc` in the extension.test.ts file
- Update the `test('Test Definition Provider using gogetdoc'` to `test.only('Test Definition Provider using gogetdoc` so that you can focus only on that single test
- Use the `Launch extension tests` to start debugging the tests

ramya-rao-a:
This failure is not consistent anymore, but happens intermittently
One example: https://travis-ci.org/microsoft/vscode-go/jobs/600424998

ramya-rao-a:
This is no longer an issue

PR #3217, opened 2020-05-02
Prompt transform value while adding tags

### LINKED ISSUE #2546: promptForTags also prompt for transform
Hi, I was hoping for a way for Go: Add Tags to prompt for the transform parameter. I often want to switch between camelCase and snake_case even within the same project.

I was thinking that if go.addTags.promptForTags is true, it would also prompt for the transform parameter.

I can't find the current version of the plugin. "0.10.2 - 30th April, 2019" is at the top of the changelog.

thanks,

bulentkazanci:
In the current version, Go: Add Tags prompts to get user input for only tags and options if `go.addTags.promptForTags` is true. With the following changes, it will also prompt for transform value too.

Closes #2546

NxOF5PNwXB

### DIFF
--- a/src/goModifytags.ts
+++ b/src/goModifytags.ts
@@ -128,7 +128,14 @@ function getTagsAndOptions(config: GoTagsConfig, commandArgs: GoTagsConfig): The
 					prompt: 'Enter comma separated options'
 				})
 				.then((inputOptions) => {
-					return [inputTags, inputOptions, transformValue];
+					return vscode.window
+						.showInputBox({
+							value: transformValue,
+							prompt: 'Enter transform value'
+						})
+						.then((transformOption) => {
+							return [inputTags, inputOptions, transformOption];
+						});
 				});
 		});
 }


bulentkazanci:
Hi @ramya-rao-a, when I make a commit, my PR fails every time because of the following error. To overcome this problem, I made couple of empty commits but it still fails. As it specified [here](https://github.com/microsoft/vscode-go/issues/2805), this failure doesn't happen all the time. Is there any workaround to overcome it?

**Error:**
```
AssertionError [ERR_ASSERTION]: No signature for gogetdocTestData/test.go:24:8
```

ramya-rao-a:
Hey @bulentkazanci 

Sorry about that, we have some flaky tests that we are working on improving.

As mentioned in https://github.com/microsoft/vscode-go/pull/3218#issuecomment-631266388, please consider creating a PR to https://github.com/golang/vscode-go as project is moving to a new home

PR #3218, opened 2020-05-02
Raise error if GOOS/GOARC is not compatible with host machine in debug

### LINKED ISSUE #2696: When debugging with "go.toolsEnvVars" that are incompatible with host machine, the failure message is opaque
Steps to Reproduce:
1. Open a Go project that has launch configurations
2. Open settings and edit `"go.toolsEnvVars"` and add `"GOOS": "windows"` (or `"GOOS": "linux"` if your OS is Windows)
3. Attempt to launch a debug session

The error:

```
could not launch process: stub exited while waiting for connection: exit status 0
Process exiting with code: 1
```

The issue is that it is not at all obvious why this is failing. Ideally, `vscode-go` would know that the debugger failed because of GOOS/GOARCH. But I don't know how it would do that, without `dlv` providing a more meaningful error. Failing that, a warning would be nice. "By the way, we see you've changed GOOS/GOARCH such that the debugger will likely fail."

bulentkazanci:
As it is specified on the issue of #2696, while debugging an application, if `GOOS` or `GOARCH` value under `go.toolsEnvVars` is specified a value different than the host machine value, a nonexplanatory error is raised. The changes in this PR compares both environment variables with the ones on the host machine, if there is an incompatibility, it raises a meaningful error.

Closes #2696 

**A show case on mac:**
goos-check

### DIFF
--- a/src/debugAdapter/goDebug.ts
+++ b/src/debugAdapter/goDebug.ts
@@ -419,6 +419,17 @@ class Delve {
 					// Not applicable to exec mode in which case `program` need not point to source code under GOPATH
 					env['GOPATH'] = getInferredGopath(dirname) || env['GOPATH'];
 				}
+				if (
+					(env['GOOS'] && env['GOOS'] !== process.platform) ||
+					(env['GOARCH'] && env['GOARCH'] !== process.arch)
+				) {
+					logError(
+						`The debugger failed to launch because GOOS/GOARCH in go.toolsEnvVars is incompatible with this host`
+					);
+					return reject(
+						'The debugger failed to launch because GOOS/GOARCH in go.toolsEnvVars is incompatible with this host'
+					);
+				}
 				this.dlvEnv = env;
 				log(`Using GOPATH: ${env['GOPATH']}`);
 


src/debugAdapter/goDebug.ts
```diff
@@ -419,6 +419,17 @@ class Delve {
 					// Not applicable to exec mode in which case `program` need not point to source code under GOPATH
 					env['GOPATH'] = getInferredGopath(dirname) || env['GOPATH'];
 				}
+				if (
+					(env['GOOS'] && env['GOOS'] !== process.platform) ||
+					(env['GOARCH'] && env['GOARCH'] !== process.arch)
```
hyangah, line 424: It's not always true that Go's GOARCH and GOOS match process.arch/platform. 
For example, on my mac GOARCH=amd64 while process.arch=x64. So, need translation. 

BTW, delve doesn't return a meaningful error message in this case? I think we should file a bug upstream.

bulentkazanci: What do you suggest doing here like keeping a list of compatible GOOS' and GOARCHs to make the comparison?

ramya-rao-a:
Hey @bulentkazanci 

Looks like this is your first PR contribution to this project, Thanks & Welcome!

We will have to close this PR as we are not accepting any PRs at the moment unless it is related to the move or critical bug fixes. Please consider creating this PR in the new home i.e https://github.com/golang/vscode-go.

Also, on a side note, see https://github.com/microsoft/vscode-go/issues/2696#issuecomment-631265774. I believe we are yet to reach an agreement on the best course of action for this issue.

PR #3224, opened 2020-05-06
Allow for 'export' in environment variable file.

smugcloud:
Hi Folks,

I was using the `envFile` parameter to process an environment variable file, and discovered that if lines were prefixed with the `export` command, the regex was not picking them up.  This change would process files containing an `export` command at the beginning of a line.

### DIFF
--- a/src/goPath.ts
+++ b/src/goPath.ts
@@ -144,7 +144,14 @@ export function parseEnvFile(envFilePath: string): { [key: string]: string } {
 	try {
 		const buffer = stripBOM(fs.readFileSync(envFilePath, 'utf8'));
 		buffer.split('\n').forEach((line) => {
-			const r = line.match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/);
+			// Allow envFile to contain an export command, but remove it for processing
+			let l;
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
.env files are meant to follow the ENVNAME=ENVVALUE format and I don't believe it is common to have `export` or `set` commands in them

@stamblerre, @hyangah Thoughts?

smugcloud:
Thanks @ramya-rao-a.  In our situation, the `export` was originally added to quickly enable `source`'ing the file.  While I agree the `export` is a Bash builtin, I was hoping this change would be acceptable, as it was intended to offer the same experience, while also accounting for an `export` prefix.

stamblerre:
I don't really have any opinions on this, but my inclination is not to add extra cases or special handling. If there is an existing parser that can be used, I'd be in support of that.

ramya-rao-a:
I agree that it is a simple enough fix, but it also sets precedence for more support in the future, like supporting Windows cmd or powershell. Therefore, I don't believe we should taking in this change. Like @stamblerre said, using existing parsers that follow expected standard would be preferable than baking in special cases of our own.

Closing this PR, but thanks for your efforts here.

PR #3232, opened 2020-05-11
test/integration: fix Build Tags checking tests

hyangah:
The test invoked 'check' concurrently with different sets of tags.
Underneath, the check function invokes the goBuild, which cancels
outstanding go build commands before starting another go build command.
That means, the three go build processes race and try to cancel
others. The premature cancelation results in empty diagnostics and 
thus, flakiness when tests expect non-empty results.

Serialize the tests.

While we are here, also fixes the use of assert.equal which expects
the actual value as the first parameter and the expected value as
the second parameter.


hyangah:
Merged to golang/vscode-go with https://go-review.googlesource.com/c/vscode-go/+/233517

Issue #3243, opened 2020-05-16
// .*DO NOT EDIT.?$


butuzov:
Please answer these questions before submitting your issue. Thanks!

### What version of Go, VS Code & VS Code Go extension are you using?
- Run `go version` to get version of Go
	- `go version go1.14.2 darwin/amd64`
- Run `code -v` or `code-insiders -v` to get version of VS Code or VS Code Insiders
	- `1.45.1`
	- `5763d909d5f12fe19f215cbfdd29a91c0fa9208a`
	- `x64`

### Describe the bug
`vscode-go` only notice generated files with `// Code generated ... DO NOT EDIT` pattern at the first line.  There are actually two problems with this.
1. ~ `// Code generated ... DO NOT EDIT` described by `go help generate`  and in [https://golang.org/s/generatedcode](https://golang.org/s/generatedcode) is more *desirable* state than, actual fact. See Examples (`1`, `2`)
2. Quoting Rob Pike `The text may appear anywhere in the file.`, not just on the first line.

#### Examples
1.  some examples from `golang/go` 
```
find . -name "*.go" -exec cat "{}" \;  | grep -E "^// .*?DO NOT EDIT" | uniq -c | sort -nr 
...
155 // THIS FILE IS GENERATED BY THE COMMAND AT THE TOP; DO NOT EDIT
 18 // Created by cgo -godefs - DO NOT EDIT
  5 // DO NOT EDIT
  1 // GENERATED BY make_perl_groups.pl; DO NOT EDIT.
```
2. Search for `go` files on a GitHub gives more examples of the same behavior

### Propose 
What community really follows is a `DO NOT EDIT` phrase in a comment. Can we also have this a trigger to show `This file seems to be generated. DO NOT EDIT.` error message?

hyangah:
Wish https://github.com/golang/go/issues/13560 made any progress in this problem domain.

@stamblerre what does `gopls` do to detect generated code?

butuzov:
@hyangah  does `vscode-go` going to require `gopls`  in the next releases?

hyangah:
@butuzov `gopls` is still in beta. So we will keep the option to use this extension without `gopls` open for a while.  But we hope to consider `gopls` and `dlv dap` first if the features can be provided through LSP and DAP.

butuzov:
@hyangah got it. Anyway, my PR with fixes ready for review.

stamblerre:
[...]

PR #3244, opened 2020-05-16
Generated Code message related fixes

butuzov:
This commit changes default functionality of vscode-go extention (see #3243 ):

1) It allows to check other (then) first line of the file for target msg,
   but only the lines with comments and empty lines at begining of the
   file.
2) It uses lookup table to cache results of previous checks
3) It uses non-greedy regexp for checking targeted lines.


hyangah:
Closing this in favor of https://github.com/golang/vscode-go/pull/69

Thanks for the PR!

PR #3264, opened 2020-06-10
Update readme about the move

ramya-rao-a:


README.md
```diff
@@ -50,220 +18,15 @@ Moving to the [golang GitHub organization](https://github.com/golang) brings us
...
-
-## Frequently Asked Questions (FAQ)
-
-Please see our wiki on [Frequently Asked Questions](https://github.com/microsoft/vscode-go/blob/master/docs/Go-with-VS-Code-FAQ-and-Troubleshooting.md) to get answers to your questions or get started with troubleshooting.
-
 ## Contributing
 
-This project welcomes contributions and suggestions. Please go through our [Contributing Guide](https://github.com/Microsoft/vscode-go/blob/master/CONTRIBUTING.md)
-to learn how you can contribute. It also includes details on the Contributor License Agreement.
+This project welcomes contributions and suggestions. Please go through the [Contributing Guide](https://github.com/golang/vscode-go/blob/master/CONTRIBUTING.md) in the new repository.
```
ramya-rao-a, line 29: @hyangah, @stamblerre Are you planning to keep the Contributing guide in the new repo or does Google projects follow a different format?

stamblerre: Wow, I did all that documentation work and didn't even notice we had that page in there still. I will delete it, we have a totally different process, documented here: https://github.com/golang/vscode-go/blob/master/docs/contributing.md.

stamblerre: I think we could probably just delete the Contributing and Code of Conduct sections, as they aren't really necessary anymore and the source of truth is in the new repository.

ramya-rao-a: I'd prefer to keep the two sections. For people who just take a glance at this, I still want to communicate that we welcome contributions.

Pushed an update to update the link. Please take another look

README.md
```diff
@@ -1,46 +1,14 @@
...
-		- [Remote Debugging](#remote-debugging)
-- [Install or update all dependencies](#install-or-update-all-dependencies)
-- [Building and Debugging the Extension](#building-and-debugging-the-extension)
-- [Tools this extension depends on](#tools-this-extension-depends-on)
-- [Frequently Asked Questions (FAQ)](#frequently-asked-questions-faq)
-- [Contributing](#contributing)
-- [Code of Conduct](#code-of-conduct)
-- [License](#license)
 
 ## We are moving! 🏡 🚚 🏡
```
stamblerre, line 6: "We have moved" instead

README.md
stamblerre: My preference is to keep this README as simple as possible, so that it's clear to readers that they are looking in the wrong place. I worry that welcoming contributions while directing people to a different repository might send a mixed message.

In any case, archiving the repository will make a pretty clear statement, so either way is fine with me.

README.md
ramya-rao-a: But that will break the link

ramya-rao-a: Which is why the first sentence under it says "As of June 2020, our new home is .."

stamblerre: Oh right of course, sorry. Though I think the link will still take people to the README so maybe it's not so bad?

ramya-rao-a: Fine

